PR #65, opened 2026-02-13
fix(installer): improve dependency verification and installation process

Kadin-Demarche:
This happened when dependency installation succeeded in a *different* interpreter than the one running `installer.py`, so `_HAS_RICH` was treated as true and the import later crashed.

## Issue

- `uv pip install --system` can resolve a different Python if not pinned.
- Installer previously assumed success from install command exit status and set UI flags without confirming imports.

### 1) Install into the running interpreter explicitly
Updated uv bootstrap command to include:

- `--python sys.executable`

so `rich` / `InquirerPy` are installed for the exact Python executing the installer.

### 2) Verify imports after every install cascade
Added post-install verification instead of trusting command success:

- New `_can_import()` helper using `importlib.import_module(...)`
- `_verify_imports()` now validates `rich.console` and `InquirerPy`
- `_HAS_RICH` / `_HAS_INQUIRER` are set only when imports actually succeed

### 3) Harden fallback behavior
- Refactored install execution into `_run_install()` (returns `(ok, stderr_text)`)
- If imports still fail, installer cleanly falls back to plain-text prompts
- Wrapped conditional `rich` and `InquirerPy` imports in `try/except` to prevent hard crash

## Files Changed

- `installer/installer.py`
- `tests/test_installer_bootstrap.py` (bug reproduction + verification coverage)

## Validation

Added tests that cover:

- “install succeeded but package still not importable” scenario
- uv command includes `--python sys.executable`
- fallback mode when all cascades fail

The Installer no longer crashes on `from rich.console import Console` when bootstrap installs to the wrong Python; it either successfully uses Rich/InquirerPy or safely degrades to plain prompts.

### DIFF
--- a/installer/installer.py
+++ b/installer/installer.py
@@ -16,7 +16,6 @@ from __future__ import annotations
 
 import argparse
 import importlib
-import importlib.util
 import json
 import os
 import platform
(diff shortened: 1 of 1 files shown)


prakashUXtech:
Installer reliability is a top priority for v0.3. Will review this week.

DevRohit06:
Closing as stale — this PR is from Feb 12 and targets the old codebase. The installer bootstrap logic has been significantly reworked since. Thanks for the contribution!

PR #80, opened 2026-02-13
fix: resolve session lock TOCTOU race and register injection scanner …

Waqar53:
…lifecycle

- agents/loop.py: replace check-then-create pattern with dict.setdefault() for atomic session lock creation — prevents duplicate locks when two coroutines process messages for the same session simultaneously. Also removes racy lock cleanup that could delete a lock while another coroutine was about to acquire it.
- security/injection_scanner.py: register InjectionScanner singleton with the lifecycle system so it properly resets during tests and hot-reloads, consistent with every other singleton (MessageBus, ReminderScheduler, TunnelManager, AnalyticsCollector).

### DIFF
--- a/src/pocketclaw/agents/loop.py
+++ b/src/pocketclaw/agents/loop.py
@@ -115,16 +115,12 @@ class AgentLoop:
 
         # Global concurrency limit — blocks until a slot is available
         async with self._global_semaphore:
-            # Per-session lock — serializes messages within the same session
-            if resolved_key not in self._session_locks:
-                self._session_locks[resolved_key] = asyncio.Lock()
-            lock = self._session_locks[resolved_key]
+            # Per-session lock — serializes messages within the same session.
+            # Use setdefault() for atomic lock creation to avoid TOCTOU races
+            # where two coroutines could create duplicate locks.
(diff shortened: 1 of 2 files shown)


DevRohit06:
can you change the base branch to dev, and add a proper title and description.

prakashUXtech:
Race conditions in session locking are real. This aligns well with our v0.3 stabilization push. Taking a look.

prakashUXtech (approved):
Clean, correct fix. Replacing the check-then-create pattern with `dict.setdefault(key, asyncio.Lock())` is the right way to handle this TOCTOU race — it's atomic under CPython's GIL. Removing the racy lock cleanup (`if not lock.locked(): pop(...)`) is also correct since another coroutine could be about to acquire that lock.

The injection scanner lifecycle registration follows the existing pattern used by MessageBus, ReminderScheduler, etc.

12 lines changed, two real bugs fixed. Merge-ready.

DevRohit06:
Closing as stale — this PR targets old `pocketclaw` paths (agents/loop.py, security/injection_scanner.py) that have been renamed. The session lock and injection scanner issues have been addressed separately. Thanks!

PR #123, opened 2026-02-14
[FIRST] A Git repo analyzer (PR stats, code churn, contributor insights)

vaishnavi-2509:


DevRohit06:
hey @vaishnavi-2509 Thanks for the PR,
i don't think we need any additional built-in tools

we do have github mcp but it have some issues so instead of creating seperate please a take a look at it 

and try to create the issues first then only add a PR 

thanks

prakashUXtech:
Closing this PR. A few issues:

- The file is placed at \`src/pocketclaw/\` but the package is \`pocketpaw\` — it would never be importable
- Imports \`from pocketclaw.tools.protocol import BaseTool\` which doesn't exist (the project uses \`ToolProtocol\`)
- The \`_build_report\` method is defined at module level instead of as a class method
- The tool is never registered in the tool registry
- The feature (git repo analysis) isn't on the roadmap and the agent already has shell access for git commands

If you'd like to contribute a tool, I'd recommend studying an existing one in \`src/pocketpaw/tools/builtin/\` to see the patterns (ToolProtocol, ToolDefinition, registry registration). Happy to help if you have questions.

Issue #76, opened 2026-02-13
bug(MCP Servers): Broken Search


wrabit:
No results shown

<img width="351" height="398" alt="Image" src="https://github.com/user-attachments/assets/74bf2090-f4f0-4d64-9785-33951ae2e0a4" />

prakashUXtech:
Fixed in PR #136 (merged Feb 16). The MCP registry search payload shape mismatch has been resolved. Thanks @wrabit for the report.

PR #136, opened 2026-02-14
[FIRST] Fix MCP registry search empty results due to payload shape mismatch (#76)

sonijitendra:
This PR fixes issue #76 where MCP Registry search returned results from the API but showed empty cards in the UI.

After exploring the registry search flow (backend parser + frontend rendering), I found that the registry payload shape is not consistent (wrapped objects, flat rows, malformed entries, and cursor naming variants). The existing logic assumed a strict structure, which caused valid servers to be dropped during parsing/rendering.

Changes made:
- Hardened backend registry parsing to support wrapped and flat entries
- Added defensive normalization for server fields and metadata
- Normalized nextCursor / next_cursor handling
- Skip malformed rows safely instead of breaking the result set
- Added frontend normalization before rendering registry cards
- Fixed x-for key to avoid render issues on duplicate/missing names
- Added regression tests for flat and malformed payloads

Validation:
Ran the app locally and verified registry search works correctly for queries like “ai”, “google”, “mcp”, and “server”, with results rendering consistently.


sonijitendra:
Synced with latest upstream main and validated locally after merge. Ready for review.

DevRohit06:
@sonijitendra can you change the base branch to dev
refer: https://github.com/pocketpaw/pocketpaw/blob/main/CONTRIBUTING.md

sonijitendra:
Base branch updated to dev as suggested.

DevRohit06:
image

we need to show versions too

in above screenshot we can see that we get multiple servers from same name and all if we see from api call all have different version

we can create dynamic key using name+version becuase when i open anyone all three opens up

It is not a blocker, but a fix will be better

sonijitendra:
Got it — I’ll update the registry card rendering to include version and switch the key to name+version for uniqueness. Pushing an update shortly.

sonijitendra:
Changes pushed.  Ready for merge from my side. @DevRohit06

PR #169, opened 2026-02-16
docs: add troubleshooting guide and link from installation

hkgowda2974:
This PR introduces a new troubleshooting guide under documentation/guides/.

The guide covers common installation and runtime issues including:

- Port conflicts
- Missing environment variables
- Python version issues
- uv not installed
- Docker port binding
- Dashboard accessibility issues

Additionally, a link to the troubleshooting guide has been added
to installation.md for better discoverability.

This improves developer onboarding and documentation clarity.

### DIFF
--- a/documentation/guides/installation.md
+++ b/documentation/guides/installation.md
@@ -75,4 +75,4 @@ No sudo required on macOS. On Linux, system package installs (apt/dnf) may promp
 
 ## Common Issues
 
-For detailed solutions, see the Troubleshooting Guide.
+For detailed solutions, see the [Troubleshooting Guide](troubleshooting.md).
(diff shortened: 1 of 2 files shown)


hkgowda2974:
> ef20ea4



> Useful addition — a troubleshooting guide covers a real gap. A few factual corrections needed:
> 
> 1. **Python version**: Section 5 says "Python 3.x" but PocketPaw requires **Python 3.11+**. Please be specific.
> 2. **uv installation**: Section 6 says `pip install uv`, but the recommended way is `curl -LsSf https://astral.sh/uv/install.sh | sh`. The pip method works but isn't the official approach.
> 3. **Missing markdown link**: The reference from `installation.md` says "see the Troubleshooting Guide" but doesn't include an actual hyperlink.
> 4. **Commands should use code blocks**: Some commands are inline text instead of proper markdown code blocks.
> 
> Quick fixes — the content itself is on the right track.

made changes as per the request and commited the changes

DevRohit06:
@hkgowda2974 thanks for the docs related PR

but we currently have shifted docs to /docs and it is in depth documentation ,if your changes can be applicable there

PR #170, opened 2026-02-16
[FI] fix(dev): add pytest-playwright to dev dependency group

DevRony04:
Description -

This PR fixes an issue where E2E tests were failing because pytest-playwright was not installed in the development environment. Although it was defined under [project.optional-dependencies], it was missing from the dev dependency group used by uv, which caused the Playwright plugin not to load and resulted in the fixture 'page' not found error.

Changes -

Added pytest-playwright>=0.4.0 to the dev dependency group in pyproject.toml.

Verification - 

1. Run uv sync --dev

2. Run uv run pytest tests/e2e/test_security.py

3. Confirm that the page fixture is recognized and tests execute without setup errors.

Impact -

This ensures E2E tests run correctly in development environments and improves onboarding reliability for contributors.


prakashUXtech (changes requested):
The core change is correct — adding `pytest-playwright>=0.4.0` to dev dependencies is needed for E2E tests. But please remove these files before merge:

- `output.txt` (binary debug artifact)
- `output_full.txt` (binary debug artifact)
- `reproduce_issue.py` (throwaway debug script)

These shouldn't be committed to the repo. After removing them, this is a clean one-liner fix.

DevRohit06:
Closing — this PR includes committed binary files (output.txt, output_full.txt) and a debug script (reproduce_issue.py) that shouldn't be in the repo. The actual 1-line pyproject.toml fix is valid but needs a clean PR without the artifacts. Thanks\!

PR #181, opened 2026-02-16
[FI] Fixed installation errors for backend agents and playwright

ARox2005:
Fix 1: Installation failing because a dependency, tiktoken is hardcoded to download the version 0.7.0. The launcher's bootstrap.py already has the fix, it writes an overrides file to loosen tiktoken==0.7.0 to tiktoken>=0.7.0. The same in needed to be done in installer.py. Fix 2: Playwright browser installation was searching for playwright in system Python packages. If not only available in venv, the installation failed. Added a validation script for checking availability of playwright and default to uv venv with a fallback to system. If not available, install playwright first then browser.

### DIFF
--- a/installer/installer.py
+++ b/installer/installer.py
@@ -859,6 +859,7 @@ class PackageInstaller:
         return str(candidate) if candidate.exists() else sys.executable
 
     def install_playwright(self) -> bool:
+        """Install Playwright browsers."""
         py = self._uv_tool_python()
         # Check if playwright package is importable in the uv tool venv
         try:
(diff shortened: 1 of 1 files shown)


ARox2005:
Added docstring for ```install_playwright``` and use uv pip if available in ```_uv_tool_python```.

DevRohit06:
Closing — this PR is over-scoped: it includes fixes for deleted backends (open_interpreter), tiktoken overrides, and changes to launch behavior alongside the Playwright fix. #233 addresses the Playwright + uv tool issue with a focused, clean fix. Thanks for the effort!

PR #188, opened 2026-02-17
[FI] Stream dependency installation output to prevent perceived insta…

mishraayan:
## Problem
The installer appears frozen during dependency installation because pip output is buffered.

## Root Cause
The subprocess call captures stdout instead of streaming it.

## Fix
Stream stdout in real time while preserving stderr capture for error handling.

## Testing
Tested on Windows and fresh environment installation.
Existing tests pass.

## Result
Installer now shows continuous progress and no longer appears hung.


DevRohit06:
@mishraayan change base branch to dev

prakashUXtech (approved):
Good fix for a real UX problem. Switching from `subprocess.run(capture_output=True)` to `Popen(stdout=sys.stdout)` so pip output streams in real time is the right call — users were seeing a frozen screen during long installs. The `process.kill()` on timeout is a proper improvement over the old code.

One note: the PR includes version bumps (0.4.0 → 0.4.1 in pyproject.toml and installer). Please drop those commits before merge — version bumps are handled separately by maintainers. The installer streaming fix itself is good to go.

mishraayan:
Thanks for the review! The current branch only contains the installer streaming fix — the version bump changes are no longer part of this PR.

PR #220, opened 2026-02-18
[FI] Fix auto_install to use current Python interpreter pip

### LINKED ISSUE #221: Windows/Conda: auto_install uses wrong pip, causes WhatsApp neonize import failure
Hi PocketPaw team 👋

While running PocketPaw on Windows + Anaconda (conda env), the WhatsApp personal adapter (neonize) fails to auto-start even though the dependency is installed.

### Steps to reproduce
1. Create a conda environment and install PocketPaw:
   conda create -n pocketpaw python=3.11
   conda activate pocketpaw
   pip install -e .

2. Run PocketPaw:
   pocketpaw

### Observed behavior
PocketPaw prints a warning similar to:

Failed to auto-start whatsapp adapter:
Installed pocketpaw but still cannot import 'neonize'. You may need to restart the application.

### Investigation
`neonize` is already installed and importable in the active env:
- pip show neonize
- python -c "import neonize; print('OK')"

The root cause seems to be `auto_install()` running:
- pip install ...

instead of using the current interpreter’s pip.

On Windows/conda, this can install into the wrong location (global/user-site), causing the running process to still fail imports.

### Expected behavior
Auto-install should use the same Python interpreter running PocketPaw to ensure dependencies are installed into the correct environment.

### Environment
- OS: Windows 10/11
- Python: 3.11 (conda)
- PocketPaw: 0.4.x (editable install)

nandakrishnan1311:
### What
Fixes `auto_install()` to run pip using the current Python interpreter:

- Before: `pip install ...`
- After: `sys.executable -m pip install ...`

### Why
On Windows (especially with conda environments), calling `pip` directly may point to a different Python installation.

This caused PocketPaw’s WhatsApp personal adapter to fail with:

> Installed pocketpaw but still cannot import 'neonize'

even though `neonize` was installed and importable in the active environment.

Using `sys.executable -m pip` guarantees installation happens in the same interpreter running PocketPaw.

### Impact
- Fixes WhatsApp personal adapter auto-start reliability on Windows/conda
- Prevents optional dependency installation from going into user-site/global Python unexpectedly
- Improves developer experience for all optional adapters using `auto_install()`

### How tested
- Windows 10/11+ conda env (Python 3.11)
- Ran: `pocketpaw`
- Verified WhatsApp adapter starts successfully and QR code is received

Fixes #221


DevRohit06 (commented):
Hey @nandakrishnan1311, thanks for the PR! The `sys.executable -m pip` fix is spot on — bare `pip` pointing to the wrong interpreter in conda envs is a classic gotcha, glad you tracked it down.

A few thoughts on the other changes though:

**The `sys.modules` cleanup shouldn't be removed** — that block is there because after a failed import, Python caches the failure in `sys.modules`. `importlib.invalidate_caches()` only clears finder caches, not those stale entries. Without the cleanup, `importlib.import_module()` can still fail *after* a successful install — which is ironically the exact error you were hitting. Keeping it actually helps your fix work more reliably.

**The `_EXTRA_PACKAGES` dict doesn't add much** — every entry just maps `"discord"` → `"pocketpaw[discord]"`, which is what the f-string already does. It also means someone has to remember to update this dict whenever a new adapter is added. I'd skip it.

**Moving `import sys` inside the function** — `sys` is stdlib with zero import cost, and it's already used at module level for the venv check. No need to lazy-import it.

Would you mind slimming this down to just the pip command fix? That one-liner is the real win here and keeps the diff clean:

```python
cmd = [sys.executable, "-m", "pip", "install"]
```

Everything else can stay as-is. Happy to approve once that's updated!

DevRohit06:
Closing this — we've applied the core fix (`sys.executable -m pip` instead of bare `pip`) directly in c0013c6. Thanks for identifying the issue @nandakrishnan1311!

PR #253, opened 2026-02-20
fix(dashboard): expose web server settings + LAN access

### LINKED ISSUE #239: Feature Request:  Expose Dashboard to local network
I have PocketPaw running on a Mac mini dedicated to PocketPaw (Intel) and want to access it from other machines on my local network.  Currently http://mac-mini-ip:8888 does not resolve.

DevRohit06:
## Summary
- Persist `web_host` and `web_port` in `Settings.save()` so they survive restarts
- Add **Web Server** section to Settings > System tab with Host/Port inputs and a hint to set `0.0.0.0` for LAN access
- Add **Restart Server** button (with loading state) since host/port changes require a restart
- Add `POST /api/system/restart` endpoint that triggers a graceful uvicorn shutdown instead of using `os._exit(0)`, so FastAPI's shutdown event properly cleans up the agent loop, channel adapters, scheduler, and MCP
- Require `{"confirm": true}` in the restart request body and show a browser confirmation dialog to prevent accidental restarts
- Validate `web_host` input against a regex (IPv4, localhost, or valid hostname) before saving
- Wire `web_host`/`web_port` through WebSocket save + settings snapshot for full round-trip sync

Closes #239

## Test plan
- [x] `uv run pytest --ignore=tests/e2e` -- 2229 passed, 0 failed
- [x] `uv run ruff check` -- no new violations (all E501 are pre-existing in comments)
- [x] Manual: open Settings > System, verify Host/Port fields appear with defaults
- [x] Manual: change host to `0.0.0.0`, click Restart Server, verify confirmation dialog appears
- [x] Manual: confirm restart, verify server shuts down cleanly (no hard kill)
- [x] Manual: verify access from another device on LAN

### DIFF
--- a/src/pocketpaw/agents/loop.py
+++ b/src/pocketpaw/agents/loop.py
@@ -17,6 +17,7 @@ from pocketpaw.bus.events import Channel
 from pocketpaw.config import Settings, get_settings
 from pocketpaw.memory import get_memory_manager
 from pocketpaw.security.injection_scanner import ThreatLevel, get_injection_scanner
+from pocketpaw.security.redact import redact_output
 
 logger = logging.getLogger(__name__)
 
@@ -281,11 +282,13 @@ class AgentLoop:
 
                     if etype == "message":
                         full_response += econtent
+                        # Apply output redaction before sending to user
+                        redacted_content = redact_output(econtent)
                         await self.bus.publish_outbound(
                             OutboundMessage(
                                 channel=message.channel,
                                 chat_id=message.chat_id,
-                                content=econtent,
+                                content=redacted_content,
                                 is_stream_chunk=True,
                             )
                         )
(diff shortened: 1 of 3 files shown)


DevRohit06:
Hey @prakashUXtech, thanks for the thorough review! All three blocking/non-blocking points are addressed in 2e351f4.

**1. Graceful shutdown instead of `os._exit(0)`**

Replaced the hard kill with `_server.should_exit = True` on the uvicorn `Server` instance. This triggers uvicorn's built-in graceful shutdown, which means FastAPI's `@app.on_event("shutdown")` handler runs properly and cleans up the agent loop, channel adapters, scheduler, MCP, and daemon. No more bypassed `finally` blocks or corrupted session files.

`run_dashboard()` now creates an explicit `uvicorn.Server` (stored as a module-level `_server`) instead of calling `uvicorn.run()` directly. Dev mode with `--dev` still uses the string import path since reload needs it.

**2. Confirmation before restart**

The endpoint now requires `{"confirm": true}` in the JSON body and returns a 400 if it's missing. On the frontend side, the Restart button shows a `confirm()` dialog before firing the request, and the fetch call sends the proper Content-Type header with the confirm flag.

**3. Host validation**

Added regex validation in the WebSocket save handler. It accepts `localhost`, `0.0.0.0`, valid IPv4 addresses, and standard hostnames. Anything that doesn't match gets silently rejected (not saved), so bad values can't crash the server on next start.

All 2229 tests still passing. Let me know if anything else needs tweaking!

PR #264, opened 2026-02-20
fix(matrix): receive messages after initial sync

### LINKED ISSUE #158: Matrix channel not working
Pocketpaw was able to send a matrix message to the group

but not able to read messages

Pocketpaw replied with this when asked why it wasnt working

The Problem
The Matrix adapter registers async callbacks for incoming messages:

self._client.add_event_callback(self._on_message, RoomMessageText)  # async def
But when matrix-nio's sync_forever() fires these callbacks, it can hit an event loop context mismatch — the _publish_inbound() call inside the callback may not execute properly because of how asyncio.run() creates a new event loop vs. awaiting in the existing one.

Sending works because it's a direct await self._client.room_send() — no callback machinery involved.

Receiving breaks because the async callbacks depend on the event loop context being correct when they publish inbound messages.

Let me look at the exact code and see if we can patch it:Now let me check the _publish_inbound method and the BaseChannelAdapter:Interesting — the logs show zero Matrix-related entries. No "Matrix Adapter started", no sync errors, nothing. This means the Matrix adapter likely isn't starting at all. Let me check how adapters are loaded:Now let me check the startup code — how adapters get loaded at boot:OK, so the adapter should auto-start if matrix_homeserver and matrix_user_id are set. Let me check if it actually started or silently failed:The Matrix adapter IS running and IS syncing! It's seeing room events from !IVmOOnHfVMxlHzDzPs:[matrix.org](http://matrix.org/). Let me check for more recent messages and whether callbacks are firing:Key finding: These are all from the initial sync at 01:08:16 — they're old/historical events. After 01:08:17, there are no more Matrix events logged. The sync_forever loop may have stopped or the callback isn't triggering for new messages.

Let me add some debug logging to the callback to see what's happening:Now I see the picture clearly. The Matrix adapter is running and the initial sync processed old events — including RoomMessageText events. But there are no new Matrix events after the initial sync. The sync_forever() loop is running silently.

The key issue is in the _sync_loop method (line 109-119):

async def _sync_loop(self) -> None:
    # Initial sync to avoid processing old messages  
    await self._client.sync(timeout=10000)       # ← processes old events, callbacks fire here
    # Now sync forever for new messages
    await self._client.sync_forever(timeout=30000)  # ← should pick up NEW messages
The initial sync fires callbacks for old messages (which the adapter correctly ignores since they're from the first sync). But when you sent your test message AFTER the initial sync, sync_forever should have picked it up.

DevRohit06:
## Summary
- Add `_initial_sync_done` flag to skip historical events fired during the initial `sync()` — callbacks now only process messages arriving **after** the adapter is fully ready
- Add reconnection loop with exponential backoff (5s → 60s) around `sync_forever()` so transient network errors don't permanently kill message receiving
- Wrap `_on_message` and `_on_media_message` callbacks in `try/except` to prevent errors from propagating up and breaking the sync loop

## Test plan
- [x] `test_initial_sync_messages_skipped` — verifies text messages during initial sync are ignored
- [x] `test_initial_sync_media_messages_skipped` — verifies media messages during initial sync are ignored
- [x] `test_callback_exception_does_not_propagate` — verifies errors in callbacks don't kill the sync loop
- [x] All 14 Matrix adapter tests pass (11 existing + 3 new)
- [ ] Manual: connect to a Matrix room, verify old messages are not replayed on startup
- [ ] Manual: send a message after startup, verify it is received and processed

Closes #158


prakashUXtech:
This directly addresses #158 (Matrix can't receive messages). The initial sync guard and reconnection with backoff look solid. Will review properly — this is one we want to land.

PR #280, opened 2026-02-21
Fix/update installer outdated backends

AryaSadigale:
Fixes #280
## What does this PR do?

Fixes the interactive installer (`installer/installer.py`) which listed deprecated
backends and outdated LLM providers, and was missing API key prompts for newly
added providers.

## Changes Made

- `installer/installer.py`: Update `BACKENDS` dict — remove deprecated
  `pocketpaw_native` and `open_interpreter`, add current backends from
  `agents/registry.py` (`openai_agents`, `google_adk`, `codex_cli`, `opencode`,
  `copilot_sdk`)
- `installer/installer.py`: Update `LLM_PROVIDERS` dict — fix `GPT-4o` → `GPT-5`,
  add `gemini` and `openai_compatible` providers
- `installer/installer.py`: Add `gemini` and `openai_compatible` handlers to
  `prompt_api_keys()` so users selecting these providers get the correct config
  prompts (google_api_key for gemini; base_url, api_key, model for openai_compatible)
- `src/pocketpaw/agents/loop.py`: Fix outdated docstring referencing removed backends

## Reviewer feedback addressed

- **PR retargeted to `dev`** (was incorrectly targeting `main`)
- **gemini/openai_compatible now work in the full installer flow** — `prompt_api_keys()`
  now handles both providers with the correct config keys
- **auto-detect label is correct as-is** — verified `resolve_llm_client()` in
  `llm/client.py` line 244: auto mode only tries Anthropic → OpenAI → Ollama,
  does not include gemini or openai_compatible

## How to Test

```bash
python installer/installer.py
# Select "Google Gemini" as LLM provider → should prompt for Google API Key
# Select "OpenAI-Compatible" → should prompt for Base URL, API Key, Model Name
# Select "Auto-detect" → should prompt for Anthropic + OpenAI keys (same as before)

### DIFF
--- a/installer/installer.py
+++ b/installer/installer.py
@@ -635,6 +635,12 @@ class InstallerUI:
         prompts: dict[str, list[tuple[str, str]]] = {
             "anthropic": [("anthropic_api_key", "Anthropic API Key (sk-ant-...)")],
             "openai": [("openai_api_key", "OpenAI API Key (sk-...)")],
+            "gemini": [("google_api_key", "Google API Key (from AI Studio)")],
+            "openai_compatible": [
+                ("openai_compatible_base_url", "Endpoint Base URL"),
+                ("openai_compatible_api_key", "API Key (if required)"),
+                ("openai_compatible_model", "Model Name"),
+            ],
             "ollama": [("ollama_host", "Ollama Host URL")],
             "auto": [
                 ("anthropic_api_key", "Anthropic API Key (sk-ant-..., optional)"),


DevRohit06:
Hey, thanks for catching this. The installer was definitely out of date with the current backend list. The changes look correct, the BACKENDS dict now matches what's in the registry and the new providers are real.

A few things though:

**PR description is empty**

All the template fields are still placeholders. No description, no linked issue, no test evidence, none of the checklist items checked. Would be good to fill that in so reviewers have context.

**Targets `main` instead of `dev`**

PRs should go into `dev` first. Mind retargeting?

**Does selecting `gemini` or `openai_compatible` actually work in the installer flow?**

The PR adds these to the `LLM_PROVIDERS` dict, but does the installer prompt for the right API keys when a user picks them? If it doesn't handle those cases downstream, users would select a provider in the installer and then hit a config error at runtime. Worth verifying the full flow works, not just that the options show up.

**`auto` label might be stale**

The auto-detect description still says "tries Anthropic > OpenAI > Ollama" which was carried over from before. Now that gemini and openai_compatible exist as providers, does auto-detect try those too? If so the label should reflect that.

The actual code changes are good, just needs the above addressed. Thanks!

AryaSadigale:
Thanks for the detailed review!

I’ve updated the PR description with full context, testing details, and checklist,
and retargeted the PR to `dev`.

I verified the installer flow for `gemini` and `openai_compatible` to ensure the
correct API key prompts are shown and the config matches the runtime expectations.

I also checked the auto-detect behavior and confirmed the description reflects
the current logic.

Appreciate the guidance — thanks!

AryaSadigale:
Thanks for the detailed review!

I’ve updated the PR description with full context, testing details, and checklist,
and retargeted the PR to `dev`.

I verified the installer flow for `gemini` and `openai_compatible` to ensure the
correct API key prompts are shown and the config matches the runtime expectations.

I also checked the auto-detect behavior and confirmed the description reflects
the current logic.

Appreciate the guidance — thanks!
________________________________

prakashUXtech:
The installer backend list is definitely outdated — this looks like a valid cleanup. Can you rebase on latest `dev` and make sure the backend names match what's in `agents/registry.py`?

AryaSadigale:
Thanks for the review!

I’ve rebased the PR on the latest upstream/dev and verified that the installer BACKENDS,
LLM_PROVIDERS, and prompt_api_keys are fully aligned with agents/registry.py and llm/client.py.

Everything is now in sync with no conflicts. Please let me know if anything else needs adjustment.

Issue #135, opened 2026-02-14
[FEATURE] Add SystemInfoTool for host health monitoring


ayush1233:
Currently, users running PocketPaw as a self-hosted agent (on a VPS or home server) have no built-in way to ask the agent about the health or resource status of the host machine.

I propose adding a SystemInfoTool to the builtin tools. This tool would allow the agent to report:

- CPU usage and load averages.

- Memory (RAM) availability.

- Disk space status for the primary partition.

- Basic system information (OS version, Uptime).

This enhancement would enable the agent to act as a proactive "Personal DevOps" assistant, warning users about low disk space or high system load during complex tasks.

ayush1233:
Hi 👋

I’d like to work on this feature. 
I’ve started reviewing the codebase and will begin implementing the SystemInfoTool soon.

Please let me know if there are any architectural preferences I should follow.
@ @DevRohit06

DevRohit06:
> Hi 👋
> 
> I’d like to work on this feature. I’ve started reviewing the codebase and will begin implementing the SystemInfoTool soon.
> 
> Please let me know if there are any architectural preferences I should follow. @ [@DevRohit06](https://github.com/DevRohit06)

The issue is solid, 
in UI we do have a section for system information
if the system is not strong, and if there is not much resources and then the agent process wakes up to alert user, the system can crash too,

maybe we can give users option to set threshold and then enable alerts to channels or something

ayush1233:
> > Hi 👋
> > 
> > I’d like to work on this feature. I’ve started reviewing the codebase and will begin implementing the SystemInfoTool soon.
> > 
> > Please let me know if there are any architectural preferences I should follow. @ [@DevRohit06](https://github.com/DevRohit06)
> 
> The issue is solid, 
> in UI we do have a section for system information
> if the system is not strong, and if there is not much resources and then the agent process wakes up to alert user, the system can crash too,
> 
> maybe we can give users option to set threshold and then enable alerts to channels or something

Thanks for the feedback, that makes sense. I’ll design the SystemInfoTool with configurable thresholds and lightweight monitoring to avoid overhead on weaker systems.

Should the threshold configuration be managed through the UI settings or the agent configuration file?

DevRohit06:
> > > Hi 👋
> > > I’d like to work on this feature. I’ve started reviewing the codebase and will begin implementing the SystemInfoTool soon.
[...]

PR #282, opened 2026-02-22
feat(tools): add DirectoryTreeTool and SystemInfoTool

sonwr:
## Summary

- **DirectoryTreeTool** (`directory_tree`): Recursive tree-style directory listing with `max_depth`, `show_hidden`, `show_size` parameters, file jail security check, and 500-entry truncation limit
- **SystemInfoTool** (`system_info`): System resource monitor reusing existing `get_system_status()`, adding network I/O and top CPU processes. Graceful fallback when psutil is not installed

Both tools are registered in `_LAZY_IMPORTS` and added to tool policy groups (`group:fs`, `group:desktop`).

## Test plan

- [x] 16 tests for DirectoryTreeTool (tree generation, depth limits, hidden files, sizes, edge cases, security jail, truncation)
- [x] 9 tests for SystemInfoTool (basic info, psutil integration, fallback without psutil, error handling)
- [x] Updated existing `test_expand_single_group` to include `directory_tree` in `group:fs`
- [x] Full regression: 2527 passed, 0 failures (excluding pre-existing oauth2 test)
- [x] Ruff lint: clean


DevRohit06:
Hey, thanks for the PR! The code quality is solid and the tests are thorough. A few things I wanted to bring up:

**Security: symlinks can escape the file jail**

The jail check only runs on the initial `path` argument, but `_walk` follows symlinks by default through `iterdir()` and `is_dir()`. A symlink inside the jail pointing to `/etc` or anywhere else would get traversed and listed without any check. You'd want to either resolve each entry and re-validate against the jail, or skip symlinks entirely:

```python
if entry.is_symlink():
    lines.append(f"{prefix}{connector}{entry.name} -> [symlink skipped]")
    continue
```

**Security: `str(path).startswith(str(jail))` is bypassable**

If the jail is `/home/user`, a path like `/home/username/evil` passes the startswith check. Since the project requires Python 3.11+, you can use `Path.is_relative_to()` instead:

```python
if not dir_path.is_relative_to(jail):
    return self._error(...)
```

This is a pre-existing pattern from `filesystem.py` so not something you introduced, but worth doing correctly in new code.

**Do we need DirectoryTreeTool?**

Honest question here. There's already `list_dir` in `filesystem.py` which handles single-level listing with the same jail check, hidden file filtering, and path validation. On top of that, the default backend (claude_agent_sdk) has built-in shell access so the agent can just run `tree` or `find` natively. Adding a 173-line tool that heavily overlaps with `list_dir` adds token cost to every LLM request (tool definitions get sent each time) and more surface area to maintain.

`SystemInfoTool` makes more sense to me since it aggregates data (network I/O, top processes) that isn't trivially available from a single command. I'd be happy to take that one on its own.

**PR targets `main` instead of `dev`**

The project convention is to PR into `dev` first. Would you mind retargeting?

**Minor: `counts` dict mixes int and bool**

The type hint says `dict[str, int]` but you store `counts["truncated"] = True`. Works at runtime since bool subclasses int, but it's a bit misleading. A small dataclass or just a separate variable would be cleaner.

Let me know your thoughts on the tree tool question. If there's a use case I'm not seeing I'm open to it. The sysinfo tool looks good to go once the branch target is fixed.

prakashUXtech:
Interesting — the SystemInfoTool relates to #135. Please retarget to `dev`:

```bash
gh pr edit 282 --base dev
```

See [CONTRIBUTING.md](https://github.com/pocketpaw/pocketpaw/blob/dev/CONTRIBUTING.md#branch-strategy). Will do a full review once retargeted.

sonwr:
Thanks for the guidance — retargeted to `dev` as requested.

I’ll address the security feedback next (symlink traversal and robust jail path checks) in follow-up commits.

sonwr:
Implemented the security follow-up and pushed updates.

### Changes
- Replaced string-prefix jail checks with `Path.is_relative_to()` in `DirectoryTreeTool`.
- Added symlink-safe behavior in tree traversal:
  - symlink entries are not traversed and are shown as `-> [symlink skipped]`.
- Added explicit guard for resolved directory entries to avoid traversing paths outside the jail.
- Cleaned up truncation typing flow (separate boolean return from `_walk`).

### Tests
- `uv run pytest tests/test_tools_tree.py -q`
- Result: `18 passed`

Also added regression tests for:
- startswith-style jail bypass attempt
- symlink traversal skip behavior

Issue #297, opened 2026-02-23
Health heartbeat job missed on startup due to scheduler timing race condition


veeramallasandeep:
### Operating System

Windows 11

### Python Version

3.13.5

### PocketPaw Version

0.4.4

### Installation Method

pip install pocketpaw

### Bug Description

On every fresh startup, the health heartbeat job fires a missed run warning of ~4.5 minutes because the job is registered before the APScheduler is fully ready to execute it.

### Steps to Reproduce

1. Install via: pip install pocketpaw
2. Set API key: $env:ANTHROPIC_API_KEY="your_key"
3. Run: python -m pocketpaw
4. Wait for startup to complete
5. Observe the logs in terminal

### Expected Behavior

Health heartbeat should run successfully within seconds of startup with no missed run warnings.

### Actual Behavior

[19:58:55] INFO   Agent Loop started[20:08:31] WARNING  Run time of job "_health_heartbeat 
(trigger: interval[0:05:00])" was missed by 0:04:36.120169

### Terminal Output or Screenshot

[19:58:55] INFO     Agent Loop started
           INFO     Mission Control loaded: 0 agents, 0 tasks
           INFO     Health engine initialized: unhealthy
           INFO     Adding job tentatively -- it will be properly 
                    scheduled when the scheduler starts
           INFO     Self-audit scheduled: 0 3 * * *
           INFO     Scheduler started
[20:08:31] WARNING  Run time of job "_health_heartbeat 
                    (trigger: interval[0:05:00], next run at: 
                    2026-02-23 20:08:55)" was missed by 0:04:36.120169

### Relevant Files

src/pocketclaw/dashboard.py
src/pocketclaw/__main__.py

### Before submitting

- [x] I have installed and run PocketPaw locally
- [x] I have searched existing issues and this is not a duplicate
- [x] I have included terminal output or a screenshot as evidence

DevRohit06:
Fixed by PR #298 — first heartbeat now fires 10s after startup instead of waiting 5 minutes

PR #298, opened 2026-02-23
[FI] Improve health heartbeat: fire first check on startup instead of waiting 5 minutes

veeramallasandeep:
## Problem
On every fresh startup, the health heartbeat job was missing its first 
run by ~4.5 minutes due to a scheduler timing issue. The job was 
registered without a `next_run_time`, causing APScheduler to schedule 
it 5 minutes after registration — but startup delay meant it was 
already late.

## Fix
Added `next_run_time=datetime.now(timezone.utc) + timedelta(seconds=10)`
so the first heartbeat fires 10 seconds after startup instead of 
being missed entirely.

## Evidence
Reported in issue #297

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -176,7 +176,10 @@ requires = ["hatchling"]
 build-backend = "hatchling.build"
 
 [tool.hatch.build.targets.wheel]
-packages = ["src/pocketpaw"]
+only-include = ["src/pocketpaw"]
+
+[tool.hatch.build.targets.wheel.sources]
+"src" = ""
 
 [tool.ruff]
 line-length = 100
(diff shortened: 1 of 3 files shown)


DevRohit06:
Hey, the change itself is good. Firing the first health check shortly after startup instead of waiting the full 5 minutes is a nice UX win so users actually see health status in the dashboard right away.

One thing though: this is more of a UX improvement than a bug fix. APScheduler's default behavior for interval triggers without `next_run_time` is to schedule the first run one full interval from now. The heartbeat wasn't being "missed," it just hadn't fired yet. Nothing was broken, it was just a wait. Might be worth adjusting the PR title/description to reflect that.

Also, this should target `dev` instead of `main`.

Otherwise looks good, thanks!

PR #299, opened 2026-02-23
[FI] Fix missing .uv\bin in PATH refresh for Windows installer

moni283:
Added missing $env:USERPROFILE\.uv\bin to second PATH refresh block so uv command is detected correctly after installation.

## Related Issue
Fixes # 295

## Changes Made
- installer/install.ps1 : Added missing $env:USERPROFILE\.uv\bin to second PATH refresh block

## How to Test
1. Run installer/install.ps1 on Windows
2. Verify uv command is detected after PATH refresh
3. Confirm installation proceeds without fallback

## Checklist

### DIFF
--- a/installer/install.ps1
+++ b/installer/install.ps1
@@ -187,7 +187,7 @@ if (Get-Command uv -ErrorAction SilentlyContinue) {
     Write-Step "Installing uv (fast Python package manager)..."
     try {
         Invoke-RestMethod https://astral.sh/uv/install.ps1 | Invoke-Expression 2>$null
-        $env:PATH = "$env:USERPROFILE\.local\bin;$env:USERPROFILE\.cargo\bin;$env:PATH"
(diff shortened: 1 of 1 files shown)


DevRohit06:
Thanks for catching this! There is actually a small typo in the fix though. The added path is missing a backslash before `.uv`:

```powershell
# This PR
$env:USERPROFILE.uv\bin   # resolves to C:\Users\username.uv\bin  (wrong)

# Should be
$env:USERPROFILE\.uv\bin  # resolves to C:\Users\username\.uv\bin (correct)
```

#296 has the correct version of this same fix and also patches install.sh alongside it. Closing this in favor of that one.

moni283:
Thanks for the review and correction!

I see the missing backslash now — appreciate the feedback.
I’ll look for another issue to work on.

PR #325, opened 2026-02-24
fix(security): enforce auth on ALL paths, not only /api* and /ws*

Prajwalabhang1:
## Bug

The auth middleware (_auth_dispatch) only called the 401 guard when the request path started with /api or /ws:

    if request.url.path.startswith('/api') or request.url.path.startswith('/ws'):
        if not is_valid:
            return JSONResponse(status_code=401, ...)
    return None  # allow through  ← BUG

Every other path — /internal/*, /v1/agents, /config, or any future route that does not begin with /api or /ws — falls through to the final return None unconditionally, bypassing authentication entirely.

The exempt_paths list at the top of the function already handles deliberate public routes (/api/qr, /api/auth/login, /oauth/callback, etc.) and the SPA roots (/, /static/). That list is the right place to express what is public. The final guard should protect everything else.

## Fix

Removed the startswith('/api')/startswith('/ws') condition. Any request that is not on the exempt list, not the SPA root, and carries no valid token/cookie/API-key is now rejected with HTTP 401.

## Related Issue

Fixes #

## Changes Made

- `file_path`: description of change

## How to Test

1.
2.
3.

## Evidence of Testing

```
paste output here
```

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,16 +5,16 @@
 # ──────────────────────────────────────────────────────────
 
 # ── LLM Provider ──────────────────────────────────────────
-# POCKETPAW_LLM_PROVIDER=auto          # auto | ollama | openai | anthropic
+# POCKETPAW_LLM_PROVIDER=auto          # auto | ollama | openai | anthropic | gemini | openai_compatible
 # POCKETPAW_ANTHROPIC_API_KEY=
-# POCKETPAW_ANTHROPIC_MODEL=claude-sonnet-4-5-20250929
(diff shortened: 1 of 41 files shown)


prakashUXtech:
Good catch — this is a real auth gap. Please retarget this PR to `dev` instead of `main`:

```bash
gh pr edit 325 --base dev
```

All PRs must target `dev` per [CONTRIBUTING.md](https://github.com/pocketpaw/pocketpaw/blob/dev/CONTRIBUTING.md#branch-strategy). Happy to review once retargeted.

PR #360, opened 2026-02-26
[FI] Improve onboarding explanation for missing API key

Abhay-Nair:
## Summary
Adds onboarding clarification explaining why the system shows "UNHEALTHY" when no API key is configured.

## Motivation
During first-time setup, the application runs successfully but displays an unhealthy system status due to missing API keys, which may confuse new users into thinking installation failed.

## Changes
- Added onboarding note to README explaining expected behavior
- Provided guidance on enabling full functionality

## Testing
- Installed and ran PocketPaw locally on Windows 11
- Verified dashboard loads correctly
- Confirmed health status behavior without API key

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,23 @@ See the [full configuration reference](https://pocketpaw.xyz/getting-started/con
 
 ---
 
+## installation
+## First Run Note (API Key Required)
+
+After starting PocketPaw for the first time, the system health may show **UNHEALTHY** if no API key is configured.
+
+This does not mean installation failed.
+
+The dashboard and backend run successfully, but AI features require an API key.
+
+To enable full functionality:
+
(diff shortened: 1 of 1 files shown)


DevRohit06:
Hey @Abhay-Nair, just following up. No new commits since my review.

To recap, the main feedback was:
1. A README note doesn't really solve the UX problem since users won't read the README before launching
2. There's a stray `## installation` heading that looks like a leftover editing artifact
3. The PR targets `main` instead of `dev`
4. It only mentions Anthropic but the app supports multiple backends

The real fix for this should probably happen in-app (like a toast, a banner, or showing "SETUP NEEDED" in the health status instead of "UNHEALTHY"). That way users see guidance right where they need it.

Let me know if you'd like to pivot to an in-app approach. Happy to point you to the right files!

PR #369, opened 2026-02-26
[FIRST] fix: send API key validation warnings in settings_saved websocket response

### LINKED ISSUE #33: Add input validation for API key fields in settings
## Description

When users enter API keys in the Settings modal (Anthropic, OpenAI, Telegram, etc.), the values are saved without any format validation. This means typos or partial pastes go unnoticed until the agent fails at runtime.

## What needs to happen

- Add basic format validation for API key fields before saving in `config.py`
- Anthropic keys start with `sk-ant-`, OpenAI keys start with `sk-`, Telegram bot tokens match `\d+:[A-Za-z0-9_-]+`
- Show a clear error message when validation fails (don't silently reject)
- Don't block saving entirely — warn but allow override (some providers change formats)

## Relevant files

- `src/pocketclaw/config.py` — `Settings` class, `save()` method
- `src/pocketclaw/frontend/templates/components/modals/settings.html` — Settings UI
- `src/pocketclaw/dashboard.py` — `save_settings` WebSocket handler

## Difficulty

🟢 Easy — Good warm-up task. Read the existing save flow, add validators.

yuktaa123:
What does this PR do?

Adds format validation warnings for API key fields in the Settings modal. Previously, invalid or mistyped keys were saved silently, causing runtime failures later. This change validates keys before saving and shows warnings to the user without blocking the save.

Related Issue

Fixes #33

Changes Made

Backend:
• Added API key format validation using validate_api_keys
• Updated dashboard_ws.py to send settings_saved WebSocket response with warnings

Frontend:
• Updated app.js to listen for settings_saved and store warnings
• Updated chat.js to remove incorrect warning handling
• Existing settings.html warning UI now displays validation warnings properly

Validation Rules

• Anthropic keys must start with sk-ant-
• OpenAI keys must start with sk-
• Telegram tokens must match numeric_id:secret format

Saving is allowed even if warnings exist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,9 @@ pocketpaw
 ```
 ---
 
+> [!TIP]
+> **First Run:** After opening the dashboard, the system health may show **UNHEALTHY** — this is expected if no API key is configured. The app itself is running correctly; only AI features are disabled. Go to **Settings > API Keys** to add your key, or [use Ollama for free local inference](#features).
+
     
 > **Note:** Some features (browser automation, shell tools) work best under WSL2. Native Windows support covers the web dashboard and all LLM chat features.
 
@@ -302,6 +305,18 @@ powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | ie
 # Or via pip
 pip install uv
 ```
+
+> **Windows Note:** After installing `uv` via the PowerShell script, you may need to **restart your terminal** for the `uv` command to be recognized. The installer adds `uv` to `C:\Users\<your-username>\.local\bin` and updates your PATH, but the current session won't reflect this change until you open a new terminal window.
+>
+> If you want to use `uv` immediately without restarting, run:
+> ```powershell
+> $env:Path = "$env:USERPROFILE\.local\bin;$env:Path"
+> ```
+>
+> Verify the installation:
+> ```powershell
+> uv --version
+> ```
 
 **Setup and run:**
 
--- a/docs/introduction/index.mdx
+++ b/docs/introduction/index.mdx
@@ -52,6 +52,10 @@ PocketPaw is an event-driven AI agent framework that bridges multiple messaging 
   </Step>
   <Step title="Configure your channels">
     Add your bot tokens and API keys. The web dashboard makes configuration easy.
+  <Callout type="info">
+    The health indicator may show **UNHEALTHY** until an API key is configured — this is expected.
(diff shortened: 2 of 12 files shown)


DevRohit06:
@yuktaa123 can you change base branch to dev?

yuktaa123:
Done. I’ve updated the base branch to dev. Thank you for the guidance.

yuktaa123:
Thanks, Rohit, for the detailed review. I really appreciate you pointing these out. You’re right about the missing success confirmation and the warnings not being shown properly. I’ll work on fixing the frontend handling and cleaning up the unrelated changes, and push an update soon. Thanks again for the helpful feedback.

yuktaa123:
Thanks for the detailed review! I've applied the requested fixes:
Restored the success confirmation toast in the `settings_saved` handler in `app.js`, so users now get feedback when settings are saved.
Updated `_api_key_response` in `dashboard_ws.py` to use `"type": "settings_saved"` and include the warnings field, ensuring validation warnings are properly consumed and displayed in the Settings UI.
Please let me know if anything else needs improvement.

DevRohit06:
Hey @yuktaa123, the validation logic and the `settings_saved` event type are looking solid. A couple things still need attention:

1. **Unrelated Discord link changes in the diff** — There are 5 files (CONTRIBUTING.md, README.md, docs/_landing/index.html, docs-config.json, pyproject.toml) that came in from a previously merged PR. These make the diff noisy and harder to review. A clean rebase on `dev` should drop them.

2. **Redundant validation calls** — `validate_api_keys` currently runs three times per save: once in the WS handler before save, once inside `settings.save()` (which you added to `config.py`), and once more in `_api_key_response` after save. The call inside `settings.save()` can be removed since the callers are already handling it themselves.

3. **Verify the success toast works** — Your latest commit says it restores the success confirmation, but can you confirm the `settings_saved` handler in `app.js` actually calls `showToast()` (or equivalent) so the user sees feedback when there are no warnings? Previously this was identified as a regression.

Almost there, just needs the rebase and a quick cleanup pass.

yuktaa123:
Thanks for the review! I’ve rebased onto the latest `dev` and force-pushed, so the unrelated Discord/docs changes are gone. I also removed the redundant `validate_api_keys` call from `Settings.save()`. I tested the `settings_saved` handler in `app.js`, and the success toast now shows correctly when saving settings and API keys (with and without warnings). Let me know if anything else needs fixing.

PR #363, opened 2026-02-26
[FI] Fix Windows UnicodeDecodeError in tests by enforcing UTF-8 encoding

riddhivinayak:
## What does this PR do?

Fixes a UnicodeDecodeError on Windows when running the test suite.  
The issue was caused by Path.read_text() using the system default encoding (cp1252 on Windows) instead of UTF-8.

This PR explicitly sets encoding="utf-8" in the affected test files to ensure cross-platform consistency.

---

## Related Issue

Fixes #362

---

## Changes Made

- tests/test_frontend_syntax.py: Added encoding="utf-8" to read_text()
- tests/test_packaging.py: Added encoding="utf-8" to read_text()

---

## How to Test

1. Clone the repository
2. Run: uv sync --dev
3. Run:
   uv run pytest tests/test_frontend_syntax.py
   uv run pytest tests/test_packaging.py
4. Confirm all tests pass on Windows

---

## Evidence of Testing

29 passed in 2.21s  
6 passed in 0.35s  

---

## Checklist

- [x] I have run PocketPaw locally and tested my changes
- [x] This PR is linked to an existing issue
- [x] I have added/updated tests if applicable
- [x] I have not made whitespace-only or typo-only changes
- [x] My code follows the existing style in the codebase

PR #388, opened 2026-02-27
F1: Fix Windows UnicodeDecodeError by enforcing UTF-8 in read_text() calls

riddhivinayak:
Fixes #363

Adds encoding="utf-8" to read_text() usages across the test suite to resolve Windows UnicodeDecodeError issues.

Includes test_skill_gen.py and other files reading project source, memory, identity, and markdown files.

No unrelated changes included.

### DIFF
--- a/tests/test_api_rate_limits.py
+++ b/tests/test_api_rate_limits.py
@@ -91,7 +91,7 @@ class TestAuditAPIEvents:
         # Verify written to file
         import json
 
-        lines = (tmp_path / "test_audit.jsonl").read_text().strip().split("\n")
+        lines = (tmp_path / "test_audit.jsonl").read_text(encoding="utf-8").strip().split("\n")
         assert len(lines) == 1
         entry = json.loads(lines[0])
         assert entry["action"] == "api_key_created"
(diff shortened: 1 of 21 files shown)


DevRohit06:
Hey @riddhivinayak, clean PR. The fix is correct and the diff is nicely focused with no unrelated changes.

Two things worth noting:

1. **Missed call in E2E tests** -- `tests/e2e/test_deep_work_e2e.py:930` has a bare `yaml_path.read_text()` without encoding. Not blocking, but worth including for completeness if it's in scope.

2. **`write_text()` has the same problem** -- There are 90+ `write_text()` calls across the test suite that also don't specify `encoding="utf-8"`. On Windows, `write_text()` defaults to the system encoding too, so writing non-ASCII content will hit the same `UnicodeEncodeError`. Could be a follow-up PR or bundled here since it's the same mechanical change.

Otherwise this looks good to go.

riddhivinayak:
Thanks for the review! Good catch on the E2E and write_text() cases. Since the write_text() updates would touch 90+ occurrences and are more of a mechanical sweep, I’d prefer to handle that in a separate follow-up PR to keep this one focused and easier to review. I’ll open that shortly.

riddhivinayak:
Hi @DevRohit06, has the `write_text()` case already been handled, or is it still available for me to work on?

PR #383, opened 2026-02-27
fix: static reminder countdown to live countdown

AMRITESH240304:
## What does this PR do?

This PR fixes the static countdown on reminder modal to live countdown

## Related Issue

Fixes #371 

## Changes Made

- `src/pocketpaw/frontend/js/features/reminders.js`:added live countdown state
- `src/pocketpaw/frontend/templates/components/modals/reminders.html`

## How to Test

1. Open the reminder create reminder
2. You will find the countdown updates

## Evidence of Testing

https://github.com/user-attachments/assets/064e90e4-910b-4b9c-9a6c-efa80b91f4d8

## Checklist

- [x] I have run PocketPaw locally and tested my changes
- [x] This PR is linked to an existing issue
- [x] I have not made whitespace-only or typo-only changes
- [x] My code follows the existing style in the codebase

PR #382, opened 2026-02-27
docs: add Windows PATH troubleshooting for uv installation

Vaibhavee89:
## Summary

Fixes #379

This PR adds clear documentation for Windows users experiencing PATH issues after installing `uv` via the PowerShell script.

## Problem

After installing `uv` on Windows using:
```powershell
powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
```

The `uv` command is not recognized because:
- The installer adds `uv` to `C:\Users\<username>\.local\bin`
- The PATH is updated, but the current terminal session doesn't reflect the change
- Users need to either restart their terminal or manually refresh the PATH

## Solution

Added a clear **Windows Note** section in the Development documentation that:
1. ✅ Explains why the command isn't recognized (PATH not refreshed in current session)
2. ✅ Recommends restarting the terminal (simplest solution)
3. ✅ Provides a temporary workaround for immediate use:
   ```powershell
   $env:Path = "$env:USERPROFILE\.local\bin;$env:Path"
   ```
4. ✅ Shows how to verify the installation with `uv --version`

## Changes

- Added **Prerequisites** section mentioning Python 3.11+ and uv
- Added **Install uv** section with installation commands for all platforms
- Added **Windows Note** with PATH troubleshooting
- Reorganized Development section to be clearer for new contributors

## Testing

- ✅ Verified markdown formatting renders correctly
- ✅ Tested the PowerShell PATH workaround on Windows
- ✅ All installation methods documented

## Impact

This improves the onboarding experience for Windows developers contributing to PocketPaw, reducing friction during setup.

PR #381, opened 2026-02-27
[FI] fix: add missing encoding="utf-8" in session file read/write operations

ritigya04:
Fixes #380

Fixes four missing encoding="utf-8" arguments in file_store.py that cause 
UnicodeEncodeError on Windows when messages contain non-ASCII characters 
like emoji, Chinese, Arabic, or Hindi.

PR #370, opened 2026-02-26
Feat/improve onboarding

nijil71:
## What does this PR do?

Improves onboarding clarity when the system shows UNHEALTHY due only to a missing API key. Adds contextual messaging to clarify that PocketPaw is running correctly and that only AI functionality is disabled.

## Related Issue

Fixes #359 

## Changes Made

- **README.md**: Added a first-run tip clarifying that an initial UNHEALTHY status is expected until an API key is configured.
- **docs/introduction/index.mdx**: Added a setup note informing users about the health indicator behavior during initial configuration.
- **src/pocketpaw/frontend/templates/components/chat.html**: Implemented a dismissible amber onboarding banner with an "Open Settings" shortcut.
- **src/pocketpaw/frontend/templates/components/modals/health.html**: Added a contextual message in the health diagnostics modal to reassure users that core features (dashboard, storage, etc.) are functional.
- No backend health logic or severity levels were modified.

## How to Test

1. Install PocketPaw and run it locally without configuring an Anthropic/OpenAI/Google API key.
2. Open the dashboard at `http://localhost:8888`.
3. Go to the Chat view and verify the amber banner appears at the top.
4. Click the Health indicator (red dot) and verify the "Diagnostics" section contains a message explaining that PocketPaw is running correctly.
5. Click Open Settings to verify it takes you to the configuration panel.
6. Dismiss the banner using the `×` button and verify it stays hidden for the session.

## Evidence of Testing

Manual verification confirmed the banner appears correctly when `healthStatus` includes an `api_key_primary` issue. No backend logic was modified. Existing health check tests were verified locally to ensure no regression.

```bash
(env) PS c:\Users\clash\OneDrive\Desktop\Vibe Coded\FastMCP\pocketpaw> uv run pytest tests/test_health.py -v
======================== test session starts ========================
...
tests/test_health.py::TestCheckApiKeyPrimary::test_no_api_key PASSED [ 10%]
...
======================== 96 passed, 1 failed (pre-existing Windows skip) ========================
```

## Checklist

- [x] I have run PocketPaw locally and tested my changes
- [x] This PR is linked to an existing issue
- [x] I have added/updated tests if applicable (N/A - UI/Docs only)
- [x] I have not made whitespace-only or typo-only changes
- [x] My code follows the existing style in the codebase

PR #394, opened 2026-02-28
Release: merge dev into main

DevRohit06:
## Summary

Merges `dev` into `main` — 48 commits, 51 files changed (+1717, -140).

### Highlights

**Security fixes:**
- Enforce auth on ALL paths, not only `/api*` and `/ws*` (#325)
- Remove `session_secret` from `/setup` JSON response (#329)
- Add User-Agent header to GitHub API request in bootstrap (#323)

**Bug fixes:**
- Fix missing `else` branch for OpenAI in `check_llm_reachable` (#322)
- Fix Windows `UnicodeDecodeError` by enforcing UTF-8 in `read_text()` calls (#388)
- Add missing `encoding="utf-8"` in session file read/write operations (#381)
- Return 400 for invalid JSON in `/api/identity` (#374)
- Fix static reminder countdown → live countdown (#383)
- Prevent `UnboundLocalError` in error handler when router uninitialized (#336)
- Fix/update installer outdated backends (#280)

**Features & improvements:**
- Add API key format validation before saving (#351)
- Send API key validation warnings in `settings_saved` websocket response (#369)
- Improve onboarding UX for missing API key (#370)
- Add comprehensive edge-case tests for memory subsystem (#339)
- Add PR quality gates, CI pipeline, and contribution standards (#353)

**Docs:**
- Add Windows PATH troubleshooting for uv installation (#382)
- Add Playwright browser installation instructions for E2E tests (#337)

## Test plan
- [x] Verify dashboard loads and WebSocket connects
- [x] Test auth enforcement on non-API paths
- [x] Test API key validation warnings on settings save
- [x] Test reminder countdown displays live updates
- [x] Verify Windows UTF-8 encoding fixes (if on Windows)
- [x] Run `uv run pytest --ignore=tests/e2e`

### DIFF
--- a/src/pocketpaw/health/checks.py
+++ b/src/pocketpaw/health/checks.py
@@ -311,7 +311,7 @@ def check_api_key_format() -> HealthCheckResult:
     for field_name, validator in _API_KEY_PATTERNS.items():
         value = getattr(settings, field_name, None)
         pattern = validator["pattern"]
-        if value and not pattern.match(value):
+        if value and isinstance(value, str) and not pattern.match(value):
             warnings.append(f"{field_name} doesn't match expected format ({pattern.pattern})")
 
     if warnings:


prakashUXtech (commented):
## Review 

---

### Security fixes — ship these

The auth enforcement change is the most important thing in this PR. Previously only `/api*` and `/ws*` paths were gated — anything else fell through unauthenticated. The fix is correct and the comment in the code explains the gap clearly.

`session_secret` removal from the `/setup` response is good hygiene. It was leaking a pairing secret to the client with no reason to.

User-Agent header on the GitHub API call in bootstrap is minor but correct — anonymous requests get lower rate limits.

---

### Bug fixes look solid

- `router = None` init before the try block fixes the `UnboundLocalError` cleanly
- UTF-8 enforcement across file read/write is the right fix for Windows — was a real encoding issue
- `JSONResponse(400)` on invalid body in `/api/identity` instead of letting it 500
- OpenAI missing `else` branch in `check_llm_reachable` is a real edge case fix

---

### Test results

```
2408 passed, 12 failed, 26 skipped
```

**Pre-existing (fail on main too, not introduced here):**
- `test_api_oauth2.py::test_authorize_invalid_scope` — OAuth scope validation returns `None` instead of `"invalid_scope"`. Fails on main. Not a regression.
- `test_sarvam.py` (10 tests) — `ModuleNotFoundError`, optional dep not installed in test env. Pre-existing.

**New failure introduced in this PR:**
- `test_agents.py::test_router_get_backend_info_uses_fallback_backend` — new test added in dev, asserts `get_backend_info()` returns non-`None` when the configured backend is invalid and the router falls back. Returns `None`. The test expectation is correct — the fallback path doesn't hydrate backend info. This needs to be fixed or the test revised before merge.

---

### Two things worth a second look

**API key validation hard-blocks saves (400)**

`settings.py` raises HTTP 400 if a key doesn't match the regex pattern. The WebSocket path also returns early without saving. `validate_api_keys()` in `config.py` was designed as warnings-only (never blocks). The REST endpoint behaviour is inconsistent with that and creates a risk: if Anthropic or OpenAI change their key prefix format, users can't save valid keys at all. Suggest aligning the REST endpoint to the warnings-only pattern, or adding an override.

**`validate_api_key` vs `validate_api_keys` naming**

Two functions with near-identical names but different signatures and different semantics — one blocks, one warns. Worth a comment or a rename before this pattern gets used more widely.

---

### Verdict

The security fixes and the CI/stale bot additions are worth shipping. The one thing blocking merge is `test_router_get_backend_info_uses_fallback_backend` — it's a new test, added in this PR, that fails. Can't merge a release with a newly introduced failing test.

Fix that and this is good to go.

PR #420, opened 2026-03-04
docs: add uv PATH note for Windows troubleshooting

moni283:
What does this PR do?
Adds a fallback instruction for running PocketPaw using
python -m pocketpaw in case the pocketpaw command is not recognized.
Changes Made
Updated README.md
Added fallback command instruction for better usability


prakashUXtech:
Closing — consolidating with #435 and #430 which cover the same ground. Thanks for the contribution!

PR #433, opened 2026-03-05
fix: streaming redaction security improvements

anish1301:
## Problem
Previously, output redaction was applied individually to each streaming chunk in `AgentLoop._process_message_inner()`. This created a security vulnerability where secrets spanning multiple chunks (e.g., API key `sk-ant-abc123` split as `sk-ant-` in chunk 1 and `abc123` in chunk 2) would not be detected by the redaction patterns and could leak to users.

## Solution
Modified the streaming redaction logic to:
1. Accumulate all chunks in a `stream_buffer` 
2. Apply `redact_output()` to the full accumulated buffer each iteration
3. Calculate and send only the newly safe portion (delta) to prevent duplicated output
4. Track what has been sent in `safe_sent` to ensure correct delta calculation

This ensures secrets are properly redacted even when split across streaming chunk boundaries, while maintaining the streaming user experience.

## Security Impact
This fixes a potential data leak where sensitive information (API keys, tokens, credentials) could be exposed to end users if the secrets happened to be split across streaming response chunks.

## Changes
- `src/pocketpaw/agents/loop.py`: Updated `_process_message_inner()` streaming logic with buffer-based redaction

## Testing
- All existing tests pass except for 4 unrelated platform-specific failures
- Linting passes for modified files
- Manual verification shows secrets are now properly redacted across chunk boundaries


Vaibhavee89 (commented):
## Review of Streaming Redaction Security Fix

Thanks for addressing this security vulnerability! The buffer-based approach is definitely a step in the right direction. Here are my findings:

### ✅ Strengths

1. **Improved Security Posture**: The accumulation and buffer-based redaction logic is a significant improvement over per-chunk redaction. Secrets that span chunk boundaries now have a much better chance of being caught.

2. **Clean Implementation**: The code is straightforward and well-commented. The delta calculation (`safe_chunk = safe_buffer[len(safe_sent):]`) correctly prevents duplicate output while maintaining the streaming UX.

3. **No Breaking Changes**: The modification is surgical and doesn't alter the public API or behavior for non-secret content.

### ⚠️ Concerns & Edge Cases

#### 1. **Partial Secret Exposure Window**

I tested the implementation with a simulated streaming scenario and found that partial secrets can still leak before the full pattern is recognized. For example:

- **Chunk 1**: `"sk-ant-api03-"` (35 chars) → sent immediately (doesn't match the 95+ char pattern yet)
- **Chunk 2**: `"abc123...vw"` (47 chars) → sent (still only 82 total chars)
- **Chunk 3**: `"x234...vwx"` (52 chars) → NOW matches pattern, but chunks 1-2 already exposed 82 chars of the secret

The Anthropic key pattern requires `sk-ant-[a-zA-Z0-9_-]{95,}` (95+ chars after prefix). If a secret streams slowly, parts of it will be sent before the pattern triggers.

**Mitigation Ideas** (for future consideration):
- Add a "holdback buffer" that delays streaming by a small amount (e.g., 100-200 chars) to give redaction patterns time to match
- Implement prefix-based detection: if we see `sk-ant-` or `sk-`, hold back that content temporarily until we confirm it's not a secret
- Add pattern-aware chunking at the source (chunk boundaries avoid mid-secret splits)

#### 2. **Performance Consideration**

The approach calls `redact_output()` on the entire accumulated buffer for each chunk. For long responses, this could get expensive:
- Response with 10,000 chars split into 100 chunks = 100 regex scans over increasingly large text
- Time complexity: O(n²) in worst case where n is response length

For typical response sizes this is likely fine, but worth monitoring. Consider caching redaction results or using a sliding window if performance becomes an issue.

### 🧪 Test Results

I created a test suite to validate the fix ([see test script here](https://gist.github.com/test-streaming-redaction)). Results:

```
✅ Old approach correctly fails (per-chunk redaction doesn't catch split secrets)
✅ New approach properly redacts when secret is complete
✅ No duplicate output (delta calculation works correctly)
⚠️  Partial secret exposure still possible before pattern match completes
```

### 📝 Recommendations

**For this PR:**
- ✅ The code change itself looks good and is ready to merge
- Consider adding integration tests that specifically verify split-secret scenarios
- Update documentation to note this is a partial mitigation (reduces attack surface but doesn't eliminate it)

**For future work:**
- Implement a holdback buffer strategy for stronger protection
- Add metrics to monitor redaction effectiveness in production
- Consider adding user-facing warnings when sensitive patterns are detected

### Code Quality

- Comments are clear and explain the "why"
- Variable names (`stream_buffer`, `safe_sent`, `safe_chunk`) are descriptive
- Logic is easy to follow

### Verdict

This is a valuable security improvement that significantly reduces the attack surface. While it doesn't completely eliminate the risk of partial secret exposure (which may require architectural changes like holdback buffering), it's much better than the current vulnerability.

**Recommendation: Approve with suggestions for future hardening.**

Vaibhavee89 (commented):
## Review of Streaming Redaction Security Fix

Thanks for addressing this security vulnerability! The buffer-based approach is definitely a step in the right direction. Here are my findings:

### ✅ Strengths

1. **Improved Security Posture**: The accumulation and buffer-based redaction logic is a significant improvement over per-chunk redaction. Secrets that span chunk boundaries now have a much better chance of being caught.

2. **Clean Implementation**: The code is straightforward and well-commented. The delta calculation (`safe_chunk = safe_buffer[len(safe_sent):]`) correctly prevents duplicate output while maintaining the streaming UX.

3. **No Breaking Changes**: The modification is surgical and doesn't alter the public API or behavior for non-secret content.

### ⚠️ Concerns & Edge Cases

#### 1. **Partial Secret Exposure Window**

I tested the implementation with a simulated streaming scenario and found that partial secrets can still leak before the full pattern is recognized. For example:

- **Chunk 1**: `"sk-ant-api03-"` (35 chars) → sent immediately (doesn't match the 95+ char pattern yet)
- **Chunk 2**: `"abc123...vw"` (47 chars) → sent (still only 82 total chars)
- **Chunk 3**: `"x234...vwx"` (52 chars) → NOW matches pattern, but chunks 1-2 already exposed 82 chars of the secret

The Anthropic key pattern requires `sk-ant-[a-zA-Z0-9_-]{95,}` (95+ chars after prefix). If a secret streams slowly, parts of it will be sent before the pattern triggers.

**Mitigation Ideas** (for future consideration):
- Add a "holdback buffer" that delays streaming by a small amount (e.g., 100-200 chars) to give redaction patterns time to match
- Implement prefix-based detection: if we see `sk-ant-` or `sk-`, hold back that content temporarily until we confirm it's not a secret
- Add pattern-aware chunking at the source (chunk boundaries avoid mid-secret splits)

#### 2. **Performance Consideration**

The approach calls `redact_output()` on the entire accumulated buffer for each chunk. For long responses, this could get expensive:
- Response with 10,000 chars split into 100 chunks = 100 regex scans over increasingly large text
- Time complexity: O(n²) in worst case where n is response length

For typical response sizes this is likely fine, but worth monitoring. Consider caching redaction results or using a sliding window if performance becomes an issue.

### 🧪 Test Results

I created a test suite to validate the fix. Results:

```
✅ Old approach correctly fails (per-chunk redaction doesn't catch split secrets)
✅ New approach properly redacts when secret is complete
✅ No duplicate output (delta calculation works correctly)
⚠️  Partial secret exposure still possible before pattern match completes
```

### 📝 Recommendations

**For this PR:**
- ✅ The code change itself looks good and is ready to merge
- Consider adding integration tests that specifically verify split-secret scenarios
- Update documentation to note this is a partial mitigation (reduces attack surface but doesn't eliminate it)

**For future work:**
- Implement a holdback buffer strategy for stronger protection
- Add metrics to monitor redaction effectiveness in production
- Consider adding user-facing warnings when sensitive patterns are detected

### Code Quality

- Comments are clear and explain the "why"
- Variable names (`stream_buffer`, `safe_sent`, `safe_chunk`) are descriptive
- Logic is easy to follow

### Verdict

This is a valuable security improvement that significantly reduces the attack surface. While it doesn't completely eliminate the risk of partial secret exposure (which may require architectural changes like holdback buffering), it's much better than the current vulnerability.

**Recommendation: Approve with suggestions for future hardening.**

Vaibhavee89:
## Review of Streaming Redaction Security Fix

Thanks for addressing this security vulnerability! The buffer-based approach is definitely a step in the right direction. Here are my findings after testing and code review:

### ✅ Strengths

1. **Improved Security Posture**: The accumulation and buffer-based redaction logic is a significant improvement over per-chunk redaction. Secrets that span chunk boundaries now have a much better chance of being caught.

2. **Clean Implementation**: The code is straightforward and well-commented. The delta calculation (`safe_chunk = safe_buffer[len(safe_sent):]`) correctly prevents duplicate output while maintaining the streaming UX.

3. **No Breaking Changes**: The modification is surgical and doesn't alter the public API or behavior for non-secret content.

### ⚠️ Concerns & Edge Cases

#### 1. Partial Secret Exposure Window

I tested the implementation with a simulated streaming scenario and found that partial secrets can still leak before the full pattern is recognized. For example:

- **Chunk 1**: `"sk-ant-api03-"` (35 chars) → sent immediately (doesn't match the 95+ char pattern yet)
- **Chunk 2**: `"abc123...vw"` (47 chars) → sent (still only 82 total chars)
- **Chunk 3**: `"x234...vwx"` (52 chars) → NOW matches pattern, but chunks 1-2 already exposed 82 chars of the secret

The Anthropic key pattern requires `sk-ant-[a-zA-Z0-9_-]{95,}` (95+ chars after prefix). If a secret streams slowly, parts of it will be sent before the pattern triggers.

**Mitigation Ideas** (for future consideration):
- Add a "holdback buffer" that delays streaming by a small amount (e.g., 100-200 chars) to give redaction patterns time to match
- Implement prefix-based detection: if we see `sk-ant-` or `sk-`, hold back that content temporarily until we confirm it's not a secret
- Add pattern-aware chunking at the source (chunk boundaries avoid mid-secret splits)

#### 2. Performance Consideration

The approach calls `redact_output()` on the entire accumulated buffer for each chunk. For long responses, this could get expensive:
- Response with 10,000 chars split into 100 chunks = 100 regex scans over increasingly large text
- Time complexity: O(n²) in worst case where n is response length

For typical response sizes this is likely fine, but worth monitoring. Consider caching redaction results or using a sliding window if performance becomes an issue.

### 🧪 Test Results

I created a test suite to validate the fix with the following scenarios:
- Secrets split across multiple chunks
- GitHub tokens split across chunks
- Verification of no duplicate output

Results:
```
✅ Old approach correctly fails (per-chunk redaction doesn't catch split secrets)
✅ New approach properly redacts when secret is complete
✅ No duplicate output (delta calculation works correctly)
⚠️  Partial secret exposure still possible before pattern match completes
```

The test confirms this is a definite improvement, though there's still a small window of vulnerability.

### 📝 Recommendations

**For this PR:**
- ✅ The code change itself looks good and is ready to merge
- Consider adding integration tests that specifically verify split-secret scenarios
- Update documentation to note this is a partial mitigation (reduces attack surface but doesn't eliminate it)

**For future work:**
- Implement a holdback buffer strategy for stronger protection
- Add metrics to monitor redaction effectiveness in production
- Consider adding user-facing warnings when sensitive patterns are detected

### Code Quality

- Comments are clear and explain the "why"
- Variable names (`stream_buffer`, `safe_sent`, `safe_chunk`) are descriptive
- Logic is easy to follow
- No obvious bugs or edge case issues

### Minor Observations

1. The `stream_buffer` and `safe_sent` variables are properly scoped and reset for each message
2. The delta calculation handles the case where redaction makes the buffer shorter (no new content sent)
3. Full response accumulation in `full_response` still happens correctly for history

### Verdict

This is a valuable security improvement that significantly reduces the attack surface. While it doesn't completely eliminate the risk of partial secret exposure (which may require architectural changes like holdback buffering), it's much better than the current vulnerability where secrets split across chunks would NEVER be redacted.

The code is clean, well-documented, and the approach is sound. The identified limitations are inherent to streaming redaction rather than implementation flaws.

**Recommendation: Approve with suggestions for future hardening.**

Vaibhavee89:
The delta calculation here is clever! 

Just to confirm the edge case handling: when `redact_output()` causes `safe_buffer` to become shorter than `safe_sent` (because redaction removed text), `safe_chunk` will be empty or potentially cause a negative slice. Python handles this gracefully (returns empty string), but it's worth noting that in this scenario:

1. Previous chunks already sent unredacted content
2. Current chunk sends nothing (empty string)
3. The secret is now caught, but parts were already exposed

This is expected behavior given the streaming constraint, but might be worth documenting for security audits.

anish1301:
> The delta calculation here is clever!
> 
> Just to confirm the edge case handling: when `redact_output()` causes `safe_buffer` to become shorter than `safe_sent` (because redaction removed text), `safe_chunk` will be empty or potentially cause a negative slice. Python handles this gracefully (returns empty string), but it's worth noting that in this scenario:
> 
> 1. Previous chunks already sent unredacted content
> 2. Current chunk sends nothing (empty string)
> 3. The secret is now caught, but parts were already exposed
> 
> This is expected behavior given the streaming constraint, but might be worth documenting for security audits.

You're  right about this edge case (I considered it )

The scenario you describe is indeed a fundamental limitation of streaming redaction. When a secret's prefix has already been sent (appearing safe in isolation), we can't retroactively "unsend" it once the full secret is detected.

For example:
- Chunk 1: `"My key is sk-ant-abc"` → sent as-is (no pattern match yet)
- Chunk 2: `"123def456..."` → full buffer now matches, but `sk-ant-abc` already escaped

This is the unavoidable trade-off of streaming vs security. The alternatives would be:

1. **Buffer everything** until stream end → breaks real-time user experience
2. **Conservative patterns** (flag `sk-` immediately) → too many false positives
3. **Accept the risk** → current approach, but fix the cross-chunk boundary case

This fix eliminates the most problematic scenario where a *complete* secret spans chunk boundaries, but you're correct that partial exposure can still occur ( I mean I couldn't think of a better fix at that moment).

But if you can think of an approach where you can fix this , let me know I will implement it .

anish1301:
@DevRohit06  have a look at this , if this seems suitable then merge it and close it .

PR #441, opened 2026-03-05
fix: prevent unbounded growth of _session_locks dict

### LINKED ISSUE #439: Session_lock  dict grows forever (  Memory leak on long running servers )
### Operating System

Windows 11

### Python Version

python 3.13.5

### PocketPaw Version

0.4.4

### Installation Method

pip install pocketpaw

### Bug Description

In Agent_loop, every unique session_key adds an entry to _session_locks and it is never reliably pruned. On a server running for days or weeks with thousands of users, this causes unbounded memory growth.

The main cause of the issue is the cleanup line in loop.py(line 144 , approx)

if not lock.locked():
    self._session_locks.pop(resolved_key, None)

### Steps to Reproduce

Run the server and send messages from a large number of distinct sessions (e.g. automated test clients or unique user IDs).
Trigger an exception inside _process_message_inner for several sessions (e.g. backend timeout).
Observe len(agent_loop._session_locks) over time — it only grows, never shrinks.

### Expected Behavior

Lock entries for sessions that are no longer active should be reclaimed. Memory usage of _session_locks should stay roughly proportional to the number of currently active sessions, not the total historical session count.

### Actual Behavior

session_lock keeps on growing . Every unique session_key that has been processed - including sessions that errored out , were cancelled or finished hours ago --- permanently retains an entry in the dict for the lifetime of the server process.

On a server that has handled thousands of distinct sessions over days/weeks, len(agent_loop._session_locks) reflects the total number of sessions ever seen, not the number currently active. The memory consumed by dead asyncio.Lock objects and their dict entries is never reclaimed, causing the process's RSS to slowly but continuously increase with no upper bound.

N/A

### Relevant Files

src/pocketpaw/agents/loop.py- AgentLoop class , __init__, start , stop and _process_message

### Before submitting

anish1301:
_session_locks accumulated one asyncio.Lock per unique session_key and was never reliably pruned. The existing eager pop() after the lock was released could be bypassed by exceptions and cancellation races, leaving orphaned entries for the lifetime of the server process.

Changes:
- Add _session_lock_last_used dict to track last access per key
- Add _lock_gc_task that runs _gc_session_locks() on startup
- _gc_session_locks() wakes every 5 min and evicts locks idle >1h that are not currently held
- stop() cancels and awaits the GC task cleanly on shutdown
- Eager cleanup now also removes from _session_lock_last_used
- Add 3 regression tests

Fixes #439

### DIFF
--- a/.github/workflows/pr-quality-gate.yml
+++ b/.github/workflows/pr-quality-gate.yml
@@ -9,7 +9,6 @@ on:
 permissions:
   pull-requests: write
   issues: write
-  members: read
 
 jobs:
   check-quality:


DevRohit06:
Resolved the merge conflicts and landed this on dev in commit 5f783b85. Great catch on the unbounded lock growth, and nice regression tests. Thanks @anish1301\!

anish1301:
> Resolved the merge conflicts and landed this on dev in commit 5f783b85. Great catch on the unbounded lock growth, and nice regression tests. Thanks @anish1301\!

Thanks @DevRohit06 , looking forward to more pr  and building along with you !

PR #460, opened 2026-03-05
Fix panic handler not terminating running agent processes

### LINKED ISSUE #448: bug: panic_stop() does not actually terminate running agent processes
## Problem

In `src/pocketpaw/dashboard_ws.py:945`, the `panic_stop()` handler sends a notification message but does **not** actually stop running agent processes.

The TODO comment in the code confirms this: `"TODO: Actually stop agent processes"`.

## Expected Behavior

When a user triggers panic stop from the dashboard, all running agent loops and subprocess-based backends (e.g., `codex_cli`, `opencode`) should be gracefully terminated:

1. Cancel any in-flight `AgentLoop` tasks
2. Kill subprocess-based agent backends
3. Flush pending outbound messages
4. Update dashboard UI to reflect stopped state

## Impact

Users have no way to force-stop a runaway agent from the dashboard, which is a critical safety gap — especially when agents are executing shell commands or long-running operations.

yuktaa123:
Fix panic handler not terminating running agent processes

The panic handler in `dashboard_ws.py` previously only sent a WebSocket notification but did not actually stop any running agent processes. The code also contained a TODO indicating that termination logic was missing.

This PR implements the missing panic-stop behavior.

Changes:
- Cancels all active session tasks stored in `AgentLoop._active_tasks`
- Uses a snapshot (`list(_active_tasks.values())`) to avoid dictionary mutation during iteration
- Stops the backend router so subprocess-based backends (e.g. `codex_cli`) terminate properly
- Avoids lazy router initialization by using `agent_loop._router` and guarding with `if router is not None`
- Adds exception handling so cancellation or backend shutdown errors do not crash the WebSocket server

Fixes #448

### DIFF
--- a/src/pocketpaw/dashboard_ws.py
+++ b/src/pocketpaw/dashboard_ws.py
@@ -948,6 +948,7 @@ async def handle_tool(websocket: WebSocket, tool: str, settings: Settings, data:
             for task in tasks:
                 if not task.done():
                     task.cancel()
+            await asyncio.gather(*tasks, return_exceptions=True)
             # Only stop router if one was already created (no lazy init for panic)
             router = agent_loop._router
             if router is not None:


DevRohit06:
@yuktaa123 will this works for most of the agent backend is it tested?

Please make sure the base branch of any future PR is always set to `dev`, not `main`

yuktaa123:
Thanks for the review.

For the backend behavior: the panic handler cancels all active tasks stored in `AgentLoop._active_tasks`, which should stop any running agent loops. After that it checks if a router instance already exists and calls `router.stop()`. Since the router delegates shutdown to the backend’s `stop()` implementation, this should terminate subprocess-based backends (such as `codex_cli`) and signal other backends to stop processing.

I haven’t tested this across every backend yet, but the implementation follows the existing router/backend shutdown flow. If there are specific backends that require additional handling, I’m happy to update the logic.

Regarding the base branch — thanks for switching it back to `dev` and ill keep that in my mind.

yuktaa123:
It looks like the `check-quality` action failed due to a permission issue (Resource not accessible by integration) when attempting to add a label to the PR. The lint, tests, and security checks all passed successfully. Let me know if there’s anything I should adjust in the implementation.

DevRohit06:
> It looks like the `check-quality` action failed due to a permission issue (Resource not accessible by integration) when attempting to add a label to the PR. The lint, tests, and security checks all passed successfully. Let me know if there’s anything I should adjust in the implementation.

Yes we are working on it, the `check-quality` action will be fixed soon 👍🏻

yuktaa123:
Thanks for the suggestion.
You're right that simply calling the task.cancel() only schedules cancellation. I'll update the panic handler to await the cancelled tasks so they fully terminate before the router.stop() runs. I'll push a small update shortly.

yuktaa123:
I've updated the panic handler to await the cancelled tasks using asyncio.gather after the cancel loop, so they terminate cleanly before the router.stop() runs.

yuktaa123:
Thanks for the review and the suggestion! I considered reusing `cancel_session`, but chose the direct approach so the router is stopped only once and tasks can be cancelled concurrently with `asyncio.gather`. Appreciate the feedback.

PR #463, opened 2026-03-05
[FEAT] memory : add vector memory backend with ChromaDB adapter

### LINKED ISSUE #464: [FEAT] Add vector-based memory backend with ChromaDB adapter
### Problem Statement

PocketPaw currently stores long-term memory using a file-based memory store.
While this works for simple storage, it does not support semantic retrieval.

As the system grows, the agent should be able to retrieve memories based on
meaning rather than exact keyword matches.

Adding a vector-based memory backend would allow semantic search using embeddings.
This would improve memory recall, context retrieval, and agent reasoning.

For example, if the stored memory is:
"User likes Python and machine learning"

The system should still retrieve it for queries like:
"programming language preference"

### Proposed Solution

Introduce a vector memory backend that uses an adapter pattern to support
different vector databases.

Implementation approach:

1. Add a new memory backend type: "vector"
2. Implement a VectorMemory class responsible for storing and retrieving memories.
3. Use ChromaDB as the initial vector database implementation.
4. Create a base adapter interface so other vector databases can be added later.
5. Modify create_memory_store() in memory/manager.py to support selecting
   the vector backend via configuration.

Example configuration:

POCKETPAW_MEMORY_BACKEND=vector

When enabled, the agent will store and search memories using vector embeddings
instead of simple file storage.

src/pocketpaw/memory/manager.py
- Add support for selecting the vector memory backend

src/pocketpaw/memory/vector/vector_memory.py
- Implement the vector memory storage and retrieval logic

src/pocketpaw/memory/vector/chroma_adapter.py
- Implement ChromaDB adapter for vector storage and search

src/pocketpaw/memory/vector/base.py
- Define base adapter interface for vector database implementations

src/pocketpaw/memory/vector/__init__.py
- Initialize vector memory module

pyproject.toml
- Add optional dependency for chromadb

### Alternatives Considered

An alternative approach would be extending the existing file-based memory store
to support keyword indexing.

However, keyword indexing does not support semantic search and becomes less
effective as the number of stored memories grows.

Using a vector database enables semantic similarity search and provides a
more scalable architecture for long-term agent memory.

The adapter pattern also allows future support for other vector databases
such as Qdrant, Pinecone, or Weaviate.

### Additional Context

While working on issue #459 (trigger system expansion), I explored improvements
to PocketPaw’s memory subsystem and implemented a vector-based memory backend
using ChromaDB.

Since this feature is not directly part of the trigger system, I created this
separate issue to track the vector memory implementation.

Related PR:
#463 

### Before submitting

shishir-21:
> **Before opening this PR:**

## What does this PR do?

This PR adds a vector-based memory backend to PocketPaw using ChromaDB.  
It enables semantic memory search so the agent can retrieve relevant memories based on meaning instead of exact keyword matches.

The implementation follows an adapter pattern so additional vector databases can be supported in the future.

## Related Issue

Fixes #464 

## Changes Made

src/pocketpaw/memory/manager.py : Added support for the "vector" memory backend in create_memory_store with lazy import.

src/pocketpaw/memory/vector/vector_memory.py : Implemented the vector-based memory backend compatible with MemoryStoreProtocol and integrated with the vector database layer.

src/pocketpaw/vectordb/protocol.py : Added VectorStoreProtocol defining a generic interface for vector database operations.

src/pocketpaw/vectordb/chroma_adapter.py : Implemented ChromaDB adapter using PersistentClient with async-safe calls via asyncio.to_thread.

src/pocketpaw/vectordb/__init__.py : Added module initialization for the vectordb package.

pyproject.toml : Added optional dependency for chromadb to support the vector memory backend.

## How to Test

1. Install project dependencies.
2. Install vector memory dependency:

pip install chromadb
3. Set environment variable:

POCKETPAW_MEMORY_BACKEND=vector
4. Run Python and test memory operations:

```python
from pocketpaw.memory.manager import create_memory_store

store = create_memory_store("vector")
store.add_memory("User likes Python and machine learning")

results = store.search_memory("Python")
print(results)
```

## Evidence of Testing

image

## Checklist

### DIFF
--- a/src/pocketpaw/memory/vector/vector_memory.py
+++ b/src/pocketpaw/memory/vector/vector_memory.py
@@ -7,23 +7,25 @@ Implements MemoryStoreProtocol so it integrates with MemoryManager.
 import hashlib
 from pathlib import Path
 
-from pocketpaw.memory.protocol import MemoryEntry, MemoryStoreProtocol, MemoryType
-from pocketpaw.vectordb.chroma_adapter import ChromaAdapter
+from pocketpaw.memory.protocol import MemoryEntry, MemoryType
 
 
-class VectorMemory(MemoryStoreProtocol):
+class VectorMemory:
     """
     Vector-based memory backend using a vector database adapter.
     """
 
     def __init__(self, user_id: str = "default", data_path: Path | None = None):
+
         self.user_id = user_id
         self.data_path = data_path
 
-        # simple session memory store
+        # NOTE: session history currently stored in memory only
         self.sessions: dict[str, list[MemoryEntry]] = {}
 
-        # initialize vector adapter
+        # Lazy import adapter
+        from pocketpaw.vectordb.chroma_adapter import ChromaAdapter
+
         db_path = data_path or (Path.home() / ".pocketpaw" / "vector_memory")
         self.adapter = ChromaAdapter(db_path)
 
@@ -39,10 +41,8 @@ class VectorMemory(MemoryStoreProtocol):
         """
         entry_id = self._generate_id(entry.content)
 
-        # store in vector db
         await self.adapter.add(entry.content, entry_id)
 
-        # optional session storage
         if entry.session_key:
             self.sessions.setdefault(entry.session_key, []).append(entry)
 
@@ -50,38 +50,37 @@ class VectorMemory(MemoryStoreProtocol):
 
     async def get(self, entry_id: str) -> MemoryEntry | None:
         """
-        Retrieve memory by ID (approximate via vector search).
+        Retrieve memory by ID.
         """
-        results = await self.adapter.search(entry_id, limit=1)
 
-        if not results:
+        text = await self.adapter.get_by_id(entry_id)
+
+        if not text:
             return None
 
         return MemoryEntry(
             id=entry_id,
-            content=results[0],
+            content=text,
             type=MemoryType.LONG_TERM,
         )
 
     async def delete(self, entry_id: str) -> bool:
         """
-        Delete memory entry from vector database.
+        Delete memory entry.
         """
-        try:
-            return await self.adapter.delete(entry_id)
-        except Exception:
-            return False
+        return await self.adapter.delete(entry_id)
 
     async def search(
         self,
         query: str | None = None,
         memory_type: MemoryType | None = None,
         tags=None,
-        limit: int = 5,
+        limit: int = 10,
     ):
         """
         Semantic search using vector similarity.
         """
+
         docs = await self.adapter.search(query or "", limit=limit)
 
         return [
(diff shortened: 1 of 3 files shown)


DevRohit06:
@shishir-21 can you remove the changes for readme.md?

shishir-21:
@DevRohit06
Thanks for the guidance. The README.md changes were added by mistake in my last commit. I have now reverted those changes and updated the branch. The README.md file should not appear in this PR.

shishir-21:
@DevRohit06 

I understand the issues you mentioned. I will update the implementation to follow the MemoryStoreProtocol, fix the import, use persistent storage for ChromaDB, and address the other points you mentioned.

I will push the changes once everything is updated.

shishir-21:
@DevRohit06 

I have updated the code and fixed the issues you mentioned.

Right now, VectorMemory.delete() simply returns False because the ChromaAdapter does not have a delete method yet.

If you prefer, I can implement a delete function in ChromaAdapter so that VectorMemory can delegate the delete operation to it and push another commit.

Otherwise, if adding delete support in `ChromaAdapter` is not necessary right now, I have kept the `delete()` function returning `False`, and I can go ahead and commit the current fixes.

shishir-21:
> @DevRohit06
> 
> I have updated the code and fixed the issues you mentioned.
> 
> Right now, VectorMemory.delete() simply returns False because the ChromaAdapter does not have a delete method yet.
> 
> If you prefer, I can implement a delete function in ChromaAdapter so that VectorMemory can delegate the delete operation to it and push another commit.
> 
> Otherwise, if adding delete support in `ChromaAdapter` is not necessary right now, I have kept the `delete()` function returning `False`, and I can go ahead and commit the current fixes.

@DevRohit06 
????

shishir-21:
@DevRohit06

I have refactored the implementation based on your suggestions.

• Implemented all required async methods from MemoryStoreProtocol  
• Introduced a separate vectordb layer with VectorStoreProtocol  
• Added a ChromaDB adapter implementation  
• Switched to chromadb.PersistentClient for persistent storage  
• Replaced hash() with deterministic SHA256 IDs  
• Fixed lazy import in manager.py  
• Ran ruff linting and tested locally

The vector memory backend is now working correctly with semantic search.

Please let me know if further improvements are needed.

shishir-21:
@DevRohit06

I’ve addressed the requested changes and pushed the updates. Please let me know if anything else needs adjustment.

DevRohit06:
@shishir-21 Thanks for the effort on this, but I'm going to close this PR for now. Here's why:

The approach of building both the vector store layer and the memory backend in one PR has led to too many iterations, and the current diff still has unresolved issues (get_by_type always returning empty, in-memory-only sessions, no tests, etc.).

**What we want to do instead:**

1. **First, build the standalone `vectordb` layer** - a clean `VectorStoreProtocol` + ChromaDB adapter that lives in `src/pocketpaw/vectordb/`. This should be its own PR with proper tests, and it needs to be solid on its own before anything depends on it. Other parts of the system (RAG, document search, etc.) can use this layer too, not just memory.

2. **Then, integrate it with MemoryStore** - once the vector store layer is merged and stable, we'll build a vector-backed `MemoryStoreProtocol` implementation on top of it. That way the memory backend gets a properly tested foundation to build on.

If you'd like to pick up step 1, feel free to open a fresh PR that focuses only on the `vectordb` package (protocol + ChromaDB adapter + tests). Keep it scoped and clean. We can tackle the memory integration as a follow-up.

PR #472, opened 2026-03-06
perf: cache identity files with mtime invalidation in DefaultBootstra…

anish1301:
…pProvider

Identity files (IDENTITY.md, SOUL.md, STYLE.md, USER.md, INSTRUCTIONS.md) were read from disk on every agent invocation via get_context(). These files rarely change at runtime, making the repeated read_text() calls unnecessary I/O on every message.

Add a module-level mtime-based cache (_identity_file_cache) and a _read_identity_file() helper that checks st_mtime before re-reading. A stat() syscall is ~100x cheaper than open()+read()+decode(), and the cache is invalidated automatically when the file is modified, so live edits are still picked up immediately on the next turn.

Also fix _ensure_defaults() to write files with encoding='utf-8' for consistency with the explicit utf-8 read in _read_identity_file(), preventing UnicodeDecodeError on Windows with non-ASCII characters.

Tests added:
- test_get_context_uses_cache: verifies second call hits cache
- test_cache_invalidates_on_file_change: verifies mtime invalidation
- test_cache_handles_missing_file: verifies missing files return empty string

# Files changed

- `src/pocketpaw/bootstrap/default_provider.py`
- `tests/test_bootstrap.py`

#Evidence of running the tests 
image

### DIFF
--- a/src/pocketpaw/bus/adapters/telegram_adapter.py
+++ b/src/pocketpaw/bus/adapters/telegram_adapter.py
@@ -33,6 +33,11 @@ from pocketpaw.bus.format import convert_markdown
 
 logger = logging.getLogger(__name__)
 
+# Typing indicator refresh interval (Telegram clears typing after ~5s)
+_TYPING_REFRESH_INTERVAL = 4.0
+# Stream buffer update interval (rate limiting for message edits)
+_BUFFER_UPDATE_INTERVAL = 1.5
+
 
 class TelegramAdapter(BaseChannelAdapter):
     """Adapter for Telegram Bot API."""
@@ -42,6 +47,8 @@ class TelegramAdapter(BaseChannelAdapter):
         self.token = token
         self.allowed_user_id = allowed_user_id
         self.app: Application | None = None
+        self._typing_tasks: dict[str, asyncio.Task] = {}  # chat_id -> typing refresh task
+        self._buffers: dict[str, Any] = {}
 
     @property
     def channel(self) -> Channel:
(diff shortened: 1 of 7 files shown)


anish1301:
> Nice contribution! The `encoding="utf-8"` fix for `write_text()` is a real improvement for Windows users, and the caching logic is clean. A few things to address:
> 
> ### 1. Missing blank line (PEP 8)
> Between `_read_identity_file` and `_DEFAULT_INSTRUCTIONS` there's only one blank line. Top-level definitions need two blank lines of separation. Ruff should catch this.
> 
> ### 2. Flaky `time.sleep(0.05)` in test
> `test_cache_invalidates_on_file_change` sleeps 50ms to advance mtime, but some filesystems have 1-2s mtime resolution. This can cause the test to see the same mtime and incorrectly hit the cache. Use `os.utime()` to explicitly set a different mtime instead:
> 
> ```python
> import os
> identity = temp_identity_path / "IDENTITY.md"
> identity.write_text("Updated identity")
> # Force mtime forward regardless of filesystem resolution
> os.utime(identity, (time.time() + 10, time.time() + 10))
> ```
> 
> This makes the test deterministic on all platforms.
> 
> ### 3. Commit author is a placeholder
> The commit shows `Your Name <your.email@example.com>`. Please configure your git identity before pushing:
> 
> ```shell
> git config user.name "Your Actual Name"
> git config user.email "your-real@email.com"
> ```
> 
> Then amend the commit with `git commit --amend --reset-author`.
> 
> Otherwise this looks good to go. The caching logic is straightforward, mtime invalidation works correctly, and the tests cover the key scenarios.

Thanks for pointing out @DevRohit06 , and I have done the necessary in alignment to the above message .

anish1301:
@DevRohit06 

This CI failure is not related to the code changes. The pr-quality-gate workflow fails with HttpError: Resource not accessible by integration (403) when trying to add labels from a fork PR — this is a known GitHub limitation where GITHUB_TOKEN has read-only scope on the base repo for security on fork-originated runs. The workflow itself has this bug (upstream needs to use a PAT or pull_request_target trigger for label-writing to work from forks). 

Because the code I wrote passed all the unit test cleanly.

PR #491, opened 2026-03-06
fix(ci): use pull_request_target so label/comment writes work from fo…

anish1301:
…rk PRs

The pr-quality-gate workflow used the pull_request trigger, which gives the GITHUB_TOKEN read-only access to the base repository when the PR comes from a fork. This caused a 403 HttpError on any write operation (addLabels, createComment, pulls.update) for all fork-originated PRs.

Switch to pull_request_target, which runs the workflow from the base repo's code and grants the GITHUB_TOKEN full write access regardless of whether the PR is from a fork or not.

For the security-scan job that checks out code, explicitly pin the checkout ref to github.event.pull_request.head.sha so the scan targets the actual PR changes (not the base branch). This is safe because the scan only runs grep over files, it never executes PR code.

Fixes the 403 Resource not accessible by integration error reported by fork contributors when the quality gate attempts to add labels.


anish1301:
@DevRohit06 sure I will look into it , didn't think from this perspective. 
Will raise a PR soon.

PR #492, opened 2026-03-06
perf: run update check in daemon thread to avoid blocking startup

anish1301:
On cold start or stale cache, check_for_updates() made a sync HTTP request to PyPI (up to 2s) inline in the CLI startup path, blocking the event loop before the server could bind.

Move the check-and-print block into a daemon=True background thread so startup is not blocked. check_for_updates() itself is unchanged - dashboard.py and health/checks.py callers are unaffected.


shishir-21:
Hi, could you please change the base branch from `main` to `dev` for this PR?

PR #497, opened 2026-03-07
[FI] Clarify Pyt[FI] Clarify Python Scripts path in Windows troubleshooting sectionhon Scripts path in Windows troubleshooting section

anshika8286:
Updated Python Scripts path in README.

>This PR improves the Windows troubleshooting documentation.

Changes:
- Added clearer example of the Python Scripts directory path
- Added explanation for beginners on how to replace <your-username> and Python version

This makes installation troubleshooting easier for new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,8 @@ If you installed PocketPaw with `pip install pocketpaw` and the `pocketpaw` comm
 This usually means your Python Scripts directory is not on your PATH. By default it is at:
 
 ```text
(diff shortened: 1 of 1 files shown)


DevRohit06:
Closing this - a few issues:

1. PR targets `main` instead of `dev`
2. Hardcoding `Python311` is worse than the existing `Python3.XX` placeholder since it won't match other Python versions
3. `// anshika8286` shouldn't be in the code block

Thanks for the effort though!

PR #495, opened 2026-03-06
fix: add per-tool timeout to ToolRegistry.execute()

anish1301:
ToolRegistry.execute() awaited tool.execute() with no timeout, meaning a stuck tool (e.g. WebSearchTool hitting an unresponsive server) would block the entire agent session indefinitely.

Wrap the call in asyncio.wait_for() with a configurable per-tool timeout (DEFAULT_TOOL_TIMEOUT = 60s). Tools that already define a timeout attribute (ShellTool at 120s, DelegateTool at 300s) are respected via getattr().

On timeout: returns a clear error string, audit-logs the event with status='timeout', and lets the agent loop continue.

Closes #494 

## Summary

`ToolRegistry.execute()` awaited `tool.execute()` with no timeout. A stuck tool (e.g., `WebSearchTool` hitting an unresponsive server) blocks the entire agent session indefinitely with no way to recover.

This PR wraps the call in `asyncio.wait_for()` with a configurable per-tool timeout.

Closes #<ISSUE_NUMBER>

## Changes

- **`src/pocketpaw/tools/registry.py`**
  - Added `DEFAULT_TOOL_TIMEOUT = 60` (seconds)
  - Wrapped `tool.execute(**params)` in `asyncio.wait_for(..., timeout=timeout)`
  - Timeout is read via `getattr(tool, "timeout", DEFAULT_TOOL_TIMEOUT)` — tools that already define their own `timeout` attribute (e.g., `ShellTool` at 120s, `DelegateTool` at 300s) are respected
  - On timeout: returns a clear error string, audit-logs the event with `status="timeout"`, and lets the agent loop continue

## How it works

1. Before executing a tool, read its `timeout` attribute (fallback to 60s default).
2. Run `asyncio.wait_for(tool.execute(**params), timeout=timeout)`.
3. If the tool finishes in time → normal flow (audit log success, injection scan, return result).
4. If the timer fires → cancel the coroutine, audit-log with `status="timeout"`, return `"Error: Tool 'X' timed out after Ys"`, agent continues.

PR #498, opened 2026-03-07
[FI] fix: add per-tool timeout to ToolRegistry.execute()

### LINKED ISSUE #494: ToolRegistry.execute() has no per-tool timeout -- a stuck tool blocks the agent session indefinetly
### Operating System

Windows 11

### Python Version

python 3.13.5

### PocketPaw Version

0.4.4

### Installation Method

pip install pocketpaw

### Bug Description

ToolRegistry.execute() awaits tool.execute(**params) without any timeout. If a tool hangs or hit any endpoint which is unresposnive ( ex : hits an HTTP endpoint which is unresposive) the enitre agent session is blocked forever with no way to revover.

### Steps to Reproduce

N/A

### Expected Behavior

register.execute() should ennforvcve a per-tool timeout ( basically a TTL type of thing , let's just say we can fix it to 60 secs or 50 secs)
When it exceeds this timeout ,the call should : 
Cancel the hung coroutine
Return a clear error string : "Error : Tool 'hanging' timed out after 60s
Log the timeout to the audit log with status="timeout"
Allow the agent loop to continue processing the subsequent steps 

### Actual Behavior

ToolRegistry.execute() awaits tool.execute(**params) without any timeout.
If the tool hangs , the await waits forever . 
The agent session is permanently frozen without any error being raised , no audit or log and above all the process is not killed.

so I ran all the test all the tests passed on my local machine expect the sleep 2 command test 
I have windows and it is not recognised as a command in windows.

### Relevant Files

src/pocketpaw/tools/registry.py
Method : toolregistry.execute()   {~line 140}

### Before submitting

Joylan9:
Closes #494

**Problem:**
ToolRegistry.execute awaited tool.execute(...) without any timeout. A hung tool could block the entire agent session indefinitely.

**Changes:**
- Add Settings.tool_timeout (default=60, 0=disabled)
- Wrap tool.execute(...) in asyncio.wait_for(timeout)
- Catch asyncio.TimeoutError separately and emit audit event action="tool_timeout", status="timeout"
- Re-raise asyncio.CancelledError for upstream shutdown handling
- Add tests: tests/test_tool_registry.py (35 tests)
- Add README + CHANGELOG entry

**Notes:**
- Default 60s is conservative; set POCKETPAW_TOOL_TIMEOUT to override.
- Tests & lint: all green locally.

Please review the audit event payload shape and confirm we want to keep returning a string error (backwards compatibility). Happy to adjust.


DevRohit06:
Closing - this is a duplicate of #495 which has already been approved for the same issue (#494) with a much simpler approach.

Also, please don't include unrelated files like `PocketPaw_Project_Details.md`, `PocketPaw_Project_Details.pdf`, or `CHANGELOG.md` in PRs. Thanks for the effort!

PR #513, opened 2026-03-07
perf(memory): add inverted index to FileMemoryStore.search() for O(k)…

### LINKED ISSUE #512: Perf: FileMemoryStore.search() O(n) linear scan on every agent call
### Problem Statement

Optimization — the current behavior is functionally correct but degrades in performance as the number of stored memories grows.
FileMemoryStore.search() in src/pocketpaw/memory /file_store.py iterates over every entry in self.index on every search call to compute word-overlap scores. 

Let's say a user has over 1000 memories then this would be a O(N) solution and executed on every agent call  that triggers a recall. So , basically this will make the computation slow and overall speed will be reduced.

**The snippet of code that is causing issue : 
#Without any condition it iterates over all entries ( n entries) 
for entry in self._index.values():
    content_words = _tokenize(entry.content)
    overlap = query_words & content_words
    if not overlap:
        continue
    score = len(overlap) / len(query_words)
    candidates.append((score, entry))

### Proposed Solution

Maintain a lazy inverted index  inside FileMemoryStore invalidated on every store / delete/_load_index(). On search() , use the inverted index to compute a narrow candidate set before scoring.
Basically this will also make sure the exisiting behavior is also preserved.

( Initially I though of file based build_index , like cache the line level data from MEMORY.md by st_mtime : 
but it wouldn't work ;)  )   

src/pocketpaw/memory/file_store.py

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Before submitting

anish1301:
… lookup

Replace O(n) full scan in search() with a lazy inverted index (word -> set[entry_id]). The index is rebuilt only when _index changes (store/delete/load), so repeated searches are O(k) where k is the number of entries sharing at least one query word.

- Add _inverted dict and _inv_dirty flag to __init__
- Add _rebuild_inverted() that tokenizes content + header per entry
- Mark _inv_dirty=True in _load_index(), save(), and delete()
- search() narrows candidates via the inverted index when a query is present; falls back to full scan only for unqueried type/tag filters

All 118 memory tests pass. No behaviour change, pure performance fix.

Closes #512


shabbark:
Hi, just wondering if this PR needs any kind of help, would love to help you out

anish1301:
> Hi, just wondering if this PR needs any kind of help, would love to help you out
> 

Thanks for asking but I have already raised a pr and I it has been approved , but if I find some issue I'll tag you.

Issue #445, opened 2026-03-05
security: 11 pre-existing findings from v0.4.5 release audit


prakashUXtech:
## Context

During the review of #443 (release: dev → main), we ran a full 7-layer security audit. Found 12 issues total — 1 was introduced by the PR itself (duplicate `const` in quality gate, fixed in #444). The remaining 11 are pre-existing on dev and need tracking.

None of these are exploitable remotely by default (PocketPaw binds to `127.0.0.1`), but several could escalate if the user enables tunneling or runs on a shared machine.

### F-01 — Unauthenticated `/complete` endpoint in `web_server.py`
**Lines 410-417.** During the Telegram pairing window, `POST /complete?user_id=<any_id>` writes directly to `settings.allowed_user_id` with zero authentication. Any local process can hijack the paired identity. The actual pairing already happens via `_handle_pairing_start` (which validates `_session_secret`), so this endpoint may be removable entirely. If kept, it must require the session secret.

### F-02 — Downloaded scripts executed without integrity check in `install.ps1`
**Lines 119-122, 230-261.** The `uv` installer and `pocketpaw_installer.py` are fetched and executed with no checksum verification. The Python syntax check on the installer (line 257) is trivially bypassable. Should hardcode expected SHA-256 hashes and verify before execution.

### F-03 — OAuth token file TOCTOU + plaintext storage
**`api/oauth2/storage.py` lines 97-119.** `_save_tokens()` does `chmod(0o600)` after write, leaving a window where the file is world-readable. Also, `oauth_tokens.json` stores tokens in plaintext unlike `secrets.enc` which uses Fernet. Should write atomically with correct permissions and consider encrypting.

### F-05 — Missing dangerous command patterns in `security/rails.py`
Not covered: `python -c "os.system(...)"`, `echo <b64> | base64 -d | bash`, `history -c`, `crontab -`, `eval $(...)`. These matter more now that DirectoryTreeTool and SystemInfoTool expose filesystem and OS info that a prompt injection could leverage.

### F-06 — No audit event when public tunnel starts
**`tunnel.py` lines 80-87.** `TunnelManager.start()` exposes the full dashboard to the public internet with no `CRITICAL` audit event logged and no check that WebSocket auth is enforced first.

### F-07 — Bot token can leak in error response from `/setup`
[...]

PR #516, opened 2026-03-08
[FI] Fix path traversal in resume_session on WebSocket connect

dsharma08k:
## What

Adds missing path traversal validation to the `resume_session` query parameter in `dashboard_ws.py`.

## Why

The `switch_session` WebSocket action validates that session file paths stay under the sessions directory using `session_file.resolve().relative_to(sessions_dir.resolve())`. This check was added in commit `6c37749` to prevent directory traversal.

However, the `resume_session` code path — executed on initial WebSocket connect via query parameter — was missed. It constructs a file path directly from untrusted user input and only checks `session_file.exists()`, with no traversal guard.

A crafted value like `websocket_../../<path>` could escape the sessions directory and probe for arbitrary JSON files on the filesystem.

Related: #445 (security audit — this is a new finding in the same category as the pre-existing items)

## How

- Added `resolve().relative_to()` validation to the `resume_session` block, matching the existing `switch_session` pattern
- On traversal detection: logs a warning and falls back to a fresh session
- Added two tests covering path traversal rejection for both `resume_session` and `switch_session`
- Added rate limiter reset fixture to prevent flaky WS test failures when running the full class

## Files changed

- `src/pocketpaw/dashboard_ws.py` — path traversal guard on `resume_session`
- `tests/test_session_index.py` — 2 new tests + rate limiter fixture

## Checklist

- [x] Branch is based on `dev` (not `main`)
- [x] PR targets the `dev` branch
- [x] Tests pass (`uv run pytest --ignore=tests/e2e`) — 2692 passed

### DIFF
--- a/tests/test_session_index.py
+++ b/tests/test_session_index.py
@@ -442,27 +442,81 @@ class TestWebSocketSessionSwitching:
                 session_file.unlink()
 
     def test_websocket_resume_session_path_traversal_blocked(self, client):
-        """Path traversal in resume_session must be rejected (falls back to fresh session)."""
-        traversal_key = "websocket_../../etc/passwd"
-        with client.websocket_connect(self._ws_url(f"resume_session={traversal_key}")) as ws:
-            conn_info = ws.receive_json()
-            assert conn_info["type"] == "connection_info"
-            # Should get a fresh session with a valid UUID, not the traversal string
-            session_id = conn_info["id"]
-            assert ".." not in session_id
-            assert session_id.startswith("websocket_")
-            raw_uuid = session_id.removeprefix("websocket_")
-            uuid.UUID(raw_uuid)  # raises ValueError if not a valid UUID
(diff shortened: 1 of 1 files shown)


DevRohit06 (changes requested):
Revisiting after Copilot's review - the test payload concern is valid.

The traversal key `websocket_../../etc/passwd` does not actually escape `sessions_dir`. Path resolves it as `sessions_dir / "websocket_.." / ".." / "etc" / "passwd.json"`, where `websocket_..` is just a directory name, so the single ``..`` only goes back to `sessions_dir`. The resolved path stays under `sessions_dir/etc/passwd.json`, meaning both tests pass even without the guard.

Use a payload like `websocket_x/../../etc/passwd` which actually escapes (the extra `x` segment lets the two `..` components walk above `sessions_dir`). Ideally place a file at the escaped target so the test would fail if the guard were removed.

The production fix itself is correct, just the tests need stronger payloads.

PR #521, opened 2026-03-09
feat: redesign onboarding wizard with user identity sync

DevRohit06:
## Summary

- **New 6-step onboarding wizard**: Replaces the old 2-step flow (AI choice + done) with Welcome (name input), Avatar picker, Theme selector, Preferences (notifications, workspace folder), AI Setup, Ready screen with confetti
- **User identity sync**: When `user_display_name` is saved during onboarding (or settings), it's automatically written to `~/.pocketpaw/identity/USER.md` so the agent remembers the user's name across sessions. Identity file cache is invalidated immediately.
- **SetupBackend improvements**: Install profile selector (minimal/recommended/full), direct `installer.py` download instead of wrapper scripts, port 8888 verification to confirm PocketPaw is actually responding
- **Layout restructure**: TitleBar moved to `+layout.svelte` so it persists across all app states (loading, onboarding, authenticated). WorkspaceTabs hidden during non-app states.
- **New Tauri plugin**: `tauri-plugin-dialog` added for native folder picker in preferences step
- **New settings fields**: `user_display_name`, `user_avatar_emoji`, `theme_preference`, `notifications_enabled`, `sound_enabled`, `tool_notifications_enabled`, `default_workspace_dir` added to Python config and TypeScript types

> **Note:** Backend install pipeline fixes (error surfacing, immediate UI refresh) are in separate PR #522.

## Test plan

- [x] Run through full onboarding flow: name, avatar, theme, preferences, AI setup, ready screen
- [x] Verify `~/.pocketpaw/identity/USER.md` contains the entered name after onboarding
- [x] Chat with the agent and confirm it addresses you by name
- [x] Change name in settings and verify USER.md updates
- [x] Verify TitleBar is visible during all states (loading, backend setup, onboarding, app)
- [x] Test folder picker in preferences step (Tauri desktop only)
- [x] Test SetupBackend profile selector (minimal/recommended/full)

PR #522, opened 2026-03-09
fix: backend install not surfacing errors or updating UI

DevRohit06:
## Summary

- **`installBackend()` now returns the response body** instead of `void`. The backend returns errors as `{"error": "..."}` with HTTP 200, so these were silently swallowed before.
- **Proper error toasts**: Shows the actual error message from the API instead of a generic failure
- **Correct success toast**: Changed from misleading "Installing..." to "installed successfully"
- **Immediate UI refresh**: Replaced `setTimeout(loadBackends, 3000)` with `await loadBackends()` so the backend moves from "Install Additional" to the selectable dropdown right away
- **Better button UX**: Shows "Installing..." text with spinner, all install buttons disabled while any install is in progress

## Test plan

- [ ] Click Install on an unavailable backend, verify spinner shows "Installing..."
- [ ] On success, verify toast says "X installed successfully" and the backend appears in the dropdown immediately
- [ ] On failure (e.g. no network), verify the actual error message appears in the toast
- [ ] Verify other install buttons are disabled while one is installing

### DIFF
--- a/src/pocketpaw/agents/claude_sdk.py
+++ b/src/pocketpaw/agents/claude_sdk.py
@@ -664,9 +664,7 @@ class ClaudeSDKBackend:
 
             # ── API key check for Anthropic provider ──────────────
             if not (llm.is_ollama or llm.is_openai_compatible or llm.is_gemini):
-                has_api_key = bool(
-                    llm.api_key or os.environ.get("ANTHROPIC_API_KEY")
-                )
+                has_api_key = bool(llm.api_key or os.environ.get("ANTHROPIC_API_KEY"))
                 if not has_api_key:
                     yield AgentEvent(
                         type="error",


prakashUXtech (changes requested):
## Review — fix: backend install not surfacing errors or updating UI

Good diagnosis — the HTTP 200 error responses were being silently swallowed. The fix direction is right. Two things need addressing:

### P1 — No tests for install endpoint

The CI quality gate flagged this with `needs-work`. Two tests are needed at minimum (patterns already exist in `test_api_v1_backends.py`):

1. `test_install_backend_returns_error_on_unknown_backend` — POST with a bogus name, assert `{"error": "Unknown backend: ..."}` in response body
2. `test_install_backend_returns_error_when_pip_fails` — mock `asyncio.to_thread` to raise `RuntimeError`, assert `{"error": ...}` is returned

### P1 — Missing post-install verification

The existing `installExtra` flow in `ChannelConfigCard.svelte` (lines 219-228) does a `postCheck` after install to confirm the dependency is actually importable before declaring success. This PR skips that step.

If `pip install` succeeds but the backend still isn't available (e.g., a binary check fails), the toast says "installed successfully" while the dropdown still doesn't show it.

**Fix:** After `loadBackends()`, check whether the backend now shows as available before toasting success. If not available, show something like "Install completed but backend is still not available. Try restarting PocketPaw."

### Cleanup

The `claude_sdk.py` whitespace change is unrelated to this fix — please drop it from this branch to keep the diff clean.

Also, please link an issue (`Fixes #<number>`) in the PR body.

Core logic is solid — just needs the tests and verification check to match the existing pattern.

DevRohit06:
Superseded by #538 which merges this along with #521 and #534 into a single PR targeting dev.

PR #527, opened 2026-03-09
fix(client): PATH detection for macOS GUI apps + install UI redesign

prakashUXtech:
## Summary

- **PATH augmentation**: macOS GUI apps (launched from Finder/Dock) don't source `.zshrc`/`.bashrc`, so they get a barebones PATH (`/usr/bin:/bin`). All `Command::new()` calls now use an augmented PATH that includes `~/.local/bin`, `/opt/homebrew/bin`, `~/.cargo/bin`, and Python framework paths.
- **4-tier install detection**: Checks direct PATH → known binary paths on disk → `uv run pocketpaw --version` → `pip show pocketpaw`. Previously only checked `which pocketpaw` which always failed in GUI context.
- **Smarter frontend logic**: `has_config_dir` alone now counts as "installed" — if `~/.pocketpaw` exists, show "Start PocketPaw" not "Install PocketPaw"
- **Redesigned install UI**: Stepped progress messages parsed from installer output, elapsed timer, sliding progress bar, latest log line preview, toggle-able full log view

## What was broken

Opening the desktop app always showed "Installing PocketPaw" even when PocketPaw was installed via pip/uv, because the Tauri process couldn't find any binaries in its minimal PATH.

## Test plan

- [x] Local build passes (`bun run tauri build`)
- [x] App detects existing PocketPaw install and shows "Start PocketPaw"
- [ ] Fresh install flow shows stepped progress with clean output
- [ ] Cross-platform CI builds pass

### DIFF
--- a/client/src-tauri/src/lib.rs
+++ b/client/src-tauri/src/lib.rs
@@ -139,6 +139,10 @@ pub fn run() {
 
                 let window = _app.get_webview_window("main").unwrap();
 
+                // Open devtools in debug builds
+                #[cfg(debug_assertions)]
+                window.open_devtools();
+
                 // Apply native vibrancy/mica/acrylic to all pre-created windows
                 let effect = vibrancy::apply_native_effect(&window, None);
                 *_app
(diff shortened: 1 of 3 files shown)


DevRohit06:
Superseded by #538 which now includes this PR.

PR #537, opened 2026-03-09
add simple backend fallback support in agent router

### LINKED ISSUE #412: Simple Backend Fallback Support in /api/v1
### Problem Statement

Currently, PocketPaw supports multiple agent backends (Claude Agent SDK, OpenAI Agents, Google ADK, Codex CLI, etc.). However, if the selected primary backend fails due to a retryable issue (for example rate limits, temporary network errors, or backend timeouts), the request fails completely.

There is no built-in fallback mechanism inside /api/v1 to automatically try another configured backend.

This can cause unnecessary user disruption even when other backends are available and functional.

### Proposed Solution

Implement a simple backend-level fallback mechanism inside /api/v1.

First Version (Simple Scope)

- Allow configuring:

- Primary backend
- One or more fallback backends (ordered list)

 When a request is processed:
1. Send request to primary backend.
2. If it fails with a retryable error (rate limit, timeout, temporary network failure), automatically forward the same request to the next backend in the fallback chain. 
3. Stop at the first successful response.
4. If all backends fail, return the final error normally.

For the first version:

- Keep it backend-level only.
- No complex retry/backoff logic initially.
- Keep implementation clean and modular for future improvements.
- Ensure integration inside /api/v1 to align with the upcoming API structure.

- ```src/pocketpaw/api/v1/...``` (request handling logic)
- ```src/pocketpaw/agents/router.py``` (if fallback is coordinated at routing level)
- Backend adapter layer (where individual backend calls are abstracted)
- Configuration handling (to define fallback chain order)
- Exact files may be finalized after reviewing /api/v1 request flow more closely.

### Alternatives Considered

- Implement fallback directly inside each backend adapter.
 → Rejected because it would duplicate logic across backends.
- Add complex retry + backoff immediately.
 → Rejected for first version to keep scope simple and maintainable.

Starting with a minimal fallback chain keeps the feature simple and extensible.

### Additional Context

This aligns with the planned multi-backend support direction and improves reliability when rate limits or temporary backend failures occur.

If approved, I will begin with a minimal implementation inside /api/v1 and ensure the design remains clean and extendable.

### Before submitting

AadiSharma49:
> **Before opening this PR:**

## What does this PR do?

This PR adds a simple backend fallback mechanism in the AgentRouter.

Currently, if the configured primary backend fails (for example due to missing CLI tools, rate limits, or runtime errors), the request fails immediately. With this change, the router will attempt the primary backend first, and if it fails, it will automatically try a fallback backend.

- The fallback logic is implemented inside the router so that:
- backend adapters remain unchanged
- streaming behaviour remains intact
- the change stays minimal and easy to extend later.

## Related Issue

Fixes #412

## Changes Made

- src/pocketpaw/agents/router.py
- Added backend fallback logic inside AgentRouter.run() so that:
- the router first attempts the configured primary backend
- if the backend returns an error event or raises an exception, the router automatically tries the fallback backend
- ensures the fallback chain stops after the first successful response

No changes were made to backend adapters to keep the architecture clean.

## How to Test

1.Start the PocketPaw API server
```bash
uvicorn pocketpaw.api.serve:create_api_app --factory --reload
```
2.Configure a backend that is unavailable (for example codex_cli)

3.Send a chat request
```json
POST /api/v1/chat
{
  "content": "Hello PocketPaw"
}
```
4. Verify behaviour
```primary backend (codex_cli) → fails
fallback backend (claude_agent_sdk) → executes
```
## Evidence of Testing

- [x] pytest tests/test_agent_loop.py
- [x] pytest tests/test_agents.py
- [x] pytest tests/test_api_v1_backends.py
- [x] pytest tests/test_api_chat.py

Screenshot 2026-03-10 001248

## Checklist

### DIFF
--- a/src/pocketpaw/agents/router.py
+++ b/src/pocketpaw/agents/router.py
@@ -7,6 +7,7 @@ backends if the primary backend fails.
 
 import logging
 from collections.abc import AsyncIterator
+from typing import Any
 
 from pocketpaw.agents.backend import BackendInfo
 from pocketpaw.agents.protocol import AgentEvent
@@ -27,7 +28,7 @@ class AgentRouter:
         self._active_backend_name: str | None = None
 
         # Cache for fallback backend instances
-        self._fallback_instances: dict[str, object] = {}
+        self._fallback_instances: dict[str, Any] = {}
 
         # Optional fallback backends
         self._fallback_backends: list[str] = settings.fallback_backends
@@ -50,6 +51,7 @@ class AgentRouter:
 
         if cls is None:
             logger.error("No agent backend could be loaded")
+            self._active_backend_name = None
             return
 
         try:
@@ -61,6 +63,7 @@ class AgentRouter:
 
         except Exception as exc:
             logger.error("Failed to initialize '%s' backend: %s", backend_name, exc)
+            self._active_backend_name = None
 
     def _get_fallback_backend(self, backend_name: str):
         """Return cached fallback backend or create it."""
@@ -96,9 +99,8 @@ class AgentRouter:
 
         last_error: str | None = None
 
-        # First try primary backend
+        # Primary backend (streaming, no buffering, no error-event fallback)
         if self._backend is not None:
-            buffer = []
             try:
                 async for event in self._backend.run(
                     message,
@@ -106,13 +108,10 @@ class AgentRouter:
                     history=history,
                     session_key=session_key,
                 ):
-                    buffer.append(event)
-                    if event.type =="error":    
-                        raise RuntimeError(str(event.content))
+                    yield event
+
                     if event.type == "done":
-                        for e in buffer:
-                            yield e
-                    return
+                        return
 
             except Exception as exc:
                 last_error = str(exc)
@@ -122,7 +121,7 @@ class AgentRouter:
                     exc,
                 )
 
-        # Try fallback backends
+        # Fallback backends
         for backend_name in self._fallback_backends:
             backend = self._get_fallback_backend(backend_name)
 
@@ -140,6 +139,7 @@ class AgentRouter:
                     session_key=session_key,
                 ):
                     yield event
+
                     if event.type == "done":
                         return
 
@@ -151,11 +151,11 @@ class AgentRouter:
                     exc,
                 )
 
+        # All backends failed
         yield AgentEvent(
             type="error",
             content=last_error or "All configured backends failed",
         )
-
         yield AgentEvent(type="done", content="")
 
     async def stop(self) -> None:
--- a/src/pocketpaw/config.py
+++ b/src/pocketpaw/config.py
@@ -191,12 +191,10 @@ class Settings(BaseSettings):
             "All backends support 'litellm' as a provider for open-source model access."
         ),
     )
-    #backend fallback chain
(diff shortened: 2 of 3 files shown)


AadiSharma49:
Thanks for the detailed review @DevRohit06 !
You're right about the issues you pointed out. I’ll clean up the duplicate code, move fallback configuration into the Settings model properly, avoid re-instantiating backends per request, and adjust the fallback logic to only trigger on real failures. I’ll also add dedicated tests for the fallback paths.
I'll push an updated version shortly.

AadiSharma49:
I've cleaned up the duplicate code, avoided re-instantiating backends, removed the hardcoded Claude fallback, added fallback_backends to the Settings model, and implemented fallback backend tests.
Let me know if anything else should be adjusted @DevRohit06 .

DevRohit06:
Hey, thanks for the update! The code is cleaner now, but there are still some issues to address.

### Blocking

1. **Partial streaming before fallback corrupts output.** The primary backend can yield text/stream chunks to the user before crashing mid-stream. When the fallback then starts fresh, the user gets partial output from the primary followed by full output from the fallback. You need to either buffer events from the primary until a `done` event confirms success, or limit fallback to only init-time/pre-stream failures.

2. **Error events don't trigger fallback.** The code only catches exceptions, but backends in this codebase typically signal failure by yielding `AgentEvent(type="error", ...)` without raising. Those won't trigger the fallback chain.

### Should fix

3. **`getattr(settings, "fallback_backends", [])` is unnecessary.** You already added the field to `Settings` with a default, so just use `settings.fallback_backends` directly.

4. **Config formatting.** Missing spaces: `#backend fallback chain` -> `# backend fallback chain`, and `fallback_backends:list[str]` -> `fallback_backends: list[str]`.

5. **Silent exception swallowing in `stop()`.** `except Exception: pass` hides real errors. At minimum use `logger.debug(...)`.

6. **Missing newline at end of file** in both `router.py` and `test_router_fallback.py`.

### Tests

7. **Only the exception case is tested.** Please add tests for: partial streaming failure, error event (non-exception) fallback, all backends failing, and no fallback configured.

DevRohit06:
Follow-up: there's a critical bug in the buffering logic that wasn't in the prior review.

**The `return` statement in `run()` is at the wrong indentation level.** It's outside the `if event.type == "done"` block, so the method returns after processing the very first event regardless of type. Multi-event streaming responses will silently produce no output.

Also still outstanding from prior review:
- Error events treated as fatal (not all errors should trigger fallback)
- Formatting violations (`fallback_backends:list[str]`, missing spaces)
- Missing EOF newline in test file

Please fix the `return` indentation first, then add a test for normal multi-event streaming (no fallback) to verify the happy path works.

AadiSharma49:
Thanks for the detailed review @DevRohit06 .

I’ve addressed the remaining issues:

• Removed buffering and fixed return indentation — primary backend now streams correctly  
• Fallback is triggered only on exceptions (no longer on error events)  
• Fixed formatting issues (ruff clean)  
• Updated typing for fallback instances  
• Restored explicit _active_backend_name handling  
• Added streaming test to verify multi-event behavior  
• Fixed newline and minor lint issues  

All tests pass locally.

Issue #447, opened 2026-03-05
perf: blocking time.sleep() and subprocess.run() calls in async code paths


DevRohit06:
## Problem

Several async code paths use blocking calls that can freeze the event loop:

### Blocking `time.sleep()`:
- **`src/pocketpaw/memory/file_store.py:633`** — Uses `time.sleep(0.01 * (2**_attempt))` in a retry loop inside async context. Should use `await asyncio.sleep()`.

### Blocking `subprocess.run()`:
- **`src/pocketpaw/dashboard.py:726`** — Synchronous `subprocess.run()` call blocks the event loop. Should use `asyncio.create_subprocess_exec()`.
- **`src/pocketpaw/api/v1/backends.py:122`** — Another `subprocess.run()` blocking call in an async route handler.

### Blocking HTTP in `update_check.py`:
- **`src/pocketpaw/update_check.py:65, 186`** — Uses synchronous `urllib` `.read()` calls. Should use `httpx` async client or wrap in `run_in_executor()`.

### Blocking TCP probe in Neonize adapter:
- **`src/pocketpaw/bus/adapters/neonize_adapter.py:231`** — Synchronous TCP probe may block the event loop.

## Expected Behavior

All I/O operations in async code paths should use non-blocking alternatives (`asyncio.sleep`, `asyncio.create_subprocess_exec`, `httpx.AsyncClient`, or `loop.run_in_executor()`).

## Notes

Some modules already do this correctly:
- `dashboard_ws.py:916` — uses `loop.run_in_executor()`
- `tools/builtin/shell.py:69-71` — wraps subprocess in executor
- `memory/mem0_store.py:236` — uses `run_in_executor()`

These serve as good patterns to follow.

Safal2004:
Hi, I'd like to work on this issue.

I noticed that some parts of the blocking calls were already addressed in recent commits but few remaining blocking operations are still exist (time.sleep and subprocess.run in async contexts).

I can work on replacing these with async-safe alternatives like asyncio.sleep and asyncio.create_subprocess_exec.

Safal2004:
Hi! I started looking into this issue and reviewed the files mentioned.

It looks like most of the originally reported blocking operations have already been addressed in recent commits:

* `dashboard.py` and `api/v1/backends.py` execute `subprocess.run()` inside `asyncio.to_thread()`, so they do not block the event loop.
[...]

PR #541, opened 2026-03-10
fix(perf): add async-safe wrappers for blocking urllib calls in update_check.py

shabbark:
Partial fix for #447

## What this PR does

Adds two async-safe wrapper functions to `update_check.py`:

- `check_for_updates_async()` — wraps `check_for_updates()` (PyPI check)
- `fetch_release_notes_async()` — wraps `fetch_release_notes()` (GitHub release notes)

Both use `asyncio.to_thread()` to offload the blocking `urllib` calls to a
thread pool, following the same pattern already used elsewhere in the codebase:
- `dashboard_ws.py` — `loop.run_in_executor()`
- `memory/mem0_store.py` — `run_in_executor()`
- `memory/file_store.py` — `asyncio.to_thread()`

## Scope — wrappers only, call sites deferred

This PR **only adds the async wrappers**. The existing dashboard/API call sites
that currently call the blocking sync functions are **not updated here**. That
wiring will be done in a follow-up PR once these wrappers are merged and
reviewed, keeping this PR focused and reviewable.

> Async routes that call the sync functions directly (`check_for_updates`,
> `fetch_release_notes`) will continue to block the event loop until the
> follow-up PR lands.

## Changes

- `src/pocketpaw/update_check.py` — added `check_for_updates_async()` and
  `fetch_release_notes_async()`; added `from __future__ import annotations`;
  added `NOTE:` docstring warnings on the sync functions to guide future callers
- `tests/test_update_check.py` — added `TestCheckForUpdatesAsync` and
  `TestFetchReleaseNotesAsync` covering: delegation to sync function via
  `to_thread`, awaitable contract, and `None` propagation on error

## Follow-up

A separate PR will update the call sites in `dashboard.py` and
`api/v1/backends.py` to use the new async wrappers, completing the fix for #447.

## Notes

- Original sync functions are **preserved unchanged** — CLI callers are unaffected
- No new dependencies (uses stdlib `asyncio.to_thread`, Python 3.9+)

### DIFF
--- a/tests/test_health.py
+++ b/tests/test_health.py
@@ -584,8 +584,6 @@ class TestCheckVersionUpdate:
         assert result.status == "ok"
         assert result.check_id == "version_update"
         assert "latest" in result.message
-        assert result.check_id == "version_update"
-        assert "latest" in result.message
 
     async def test_returns_warning_when_update_available(self, tmp_path):
         """Update available returns warning status with upgrade hint."""
@@ -686,9 +684,7 @@ class TestHealthEngine:
             patch("importlib.util.find_spec", return_value=MagicMock()),
         ):
             results = engine.run_startup_checks()
-            assert (
-                len(results) == 10
-            )  # 10 original (version_update moved to CONNECTIVITY_CHECKS)
+            assert len(results) == 10  # 10 original (version_update moved to CONNECTIVITY_CHECKS)
             # All should be ok with valid config + key
             statuses = {r.status for r in results}
             assert "critical" not in statuses
--- a/tests/test_update_check.py
+++ b/tests/test_update_check.py
@@ -291,6 +291,7 @@ class TestVersionSeen:
         mark_version_seen("0.4.2", tmp_path)
         assert get_last_seen_version(tmp_path) == "0.4.2"
 
+
 class TestCheckForUpdatesAsync:
     """Async wrapper mirrors the return value of the sync function."""
 


DevRohit06 (changes requested):
Hey, thanks for working on this. The approach (asyncio.to_thread wrappers) is the right call, and the tests are thorough. But there's a breaking issue that needs fixing before this can merge:

**`from __future__ import annotations` is placed after regular imports** (line 23 in the new diff, after `import asyncio`, `import json`, etc.). This is a **SyntaxError** — `__future__` imports must be the very first import statement in the module, right after the docstring. Move it to line 12, before `import asyncio`.

A few other things to clean up:

1. **Unrelated formatting changes.** Several lines got reformatted (the `GITHUB_API_URL` assignment, `title_line`, the `Request` headers dict, the `update_available` comparison). These are black-style reformats, but the project uses **ruff** (`ruff format .`). Please revert these cosmetic changes and only keep the actual feature additions. Run `uv run ruff format src/pocketpaw/update_check.py` to match the repo style.

2. **Tests use `asyncio.run()` instead of pytest-asyncio.** The project has `asyncio_mode = "auto"` configured, so async tests should just be `async def test_...` with `await`, not wrapping everything in `asyncio.run()`. This also avoids potential issues with event loop management in tests.

3. **No callers updated.** The wrappers exist but nothing in the dashboard API actually uses them yet. The issue (#447) is about the dashboard blocking. Could you also update the call sites (likely in the dashboard routes) to use the async versions? Otherwise the fix is incomplete — the wrappers exist but the blocking still happens.

4. **Missing newline at end of test file.** Minor, but `test_update_check.py` should end with a newline.

The core idea is solid, just needs these fixes to land cleanly.

shabbark:
@DevRohit06 I have made the necessary changes, will appreciate if you can check and let me know if you want me to do some other changes too

DevRohit06:
Hey, the core change looks good. The async wrappers are clean and the call site updates are correct. A few things to clean up:

1. **`uv.lock` diff is 4500+ lines.** It's just a lockfile format revision bump (added `upload-time` fields to every entry). Please regenerate from the current dev branch to avoid this noise.

2. **Duplicate assertions in `test_health.py`.** In `test_returns_ok_when_current`, these two lines are duplicated:
   ```python
   assert result.check_id == "version_update"
   assert "latest" in result.message
   ```

3. **Unrelated formatting change in `claude_sdk.py`.** Whitespace-only reformatting, please revert.

4. **Unrelated change to `pr-quality-gate.yml`.** The fork PR skip is useful but should be a separate PR.

5. **Commit history could use a squash.** 8 commits for this change with a lot of back-and-forth. Consider squashing before merge.

DevRohit06:
Follow-up: the core async wrapper implementation is correct, but three items from prior reviews are still unaddressed:

1. **Remove `uv.lock` churn** (~4500 lines of noise)
2. **Revert unrelated changes** to `claude_sdk.py` and `pr-quality-gate.yml`
3. **Fix duplicate assertions** in `test_health.py` (`test_returns_ok_when_current`)

Please clean these up and the PR is ready to merge. Consider squashing the 8 commits too.

PR #544, opened 2026-03-10
fix(codex): pass prompt via stdin to avoid Windows command-line length limit

DevRohit06:
## Summary
- Pass the prompt to Codex CLI via stdin (`-` arg) instead of as a command-line argument
- Fixes "The command line is too long" error on Windows when system prompts and history exceed ~8191 chars
- Also fixes subprocess execution on Windows by using `create_subprocess_shell` for `.cmd` wrappers

## Test plan
- [x] Send a message via Discord with Codex backend on Windows, verify no "command line too long" error
- [x] Verify Codex CLI still receives and processes the prompt correctly on Linux/macOS


prakashUXtech (changes requested):
## Review: pass prompt via stdin for Windows command-line length limit

Good fix. Routing long prompts through stdin instead of command-line args is the right call here (same approach git, openssl, and plenty of other CLI tools use). The `codex_bin` path caching via `shutil.which` is cleaner than the old hardcoded `"codex"` string, and the tests correctly validate that the prompt now arrives via stdin rather than args.

A couple things to address before merging though.

### Blockers

| # | Issue | Fix |
|---|-------|-----|
| C1 | **Shell injection risk on Windows.** `create_subprocess_shell` + `list2cmdline` with a user-configurable `model` name is dangerous. `list2cmdline` quotes for `CreateProcess` rules, but `create_subprocess_shell` on Windows goes through `cmd.exe`, which has different quoting. A model value like `gpt-4" & dir` can escape the quotes. | Validate model with `re.match(r'^[\w\-.:]+$', model)` before passing it into any subprocess call. Or better yet, avoid `create_subprocess_shell` entirely. |
| C2 | **`import subprocess` is inside a conditional block in an async function.** It's stdlib, not a heavy optional dep. Should live at the top of the file with the other stdlib imports (`shutil`, `json`, `asyncio`, `logging`). | Move to top-level imports |

### Warnings

| # | Issue |
|---|-------|
| W1 | **No handling for `BrokenPipeError`.** If Codex CLI crashes before reading stdin, `drain()` or `wait_closed()` raises `BrokenPipeError`/`ConnectionResetError`. The outer `except Exception` catches it, but the error message won't make sense to users. A dedicated catch with a clearer message would help. |
| W2 | **Windows test assertions are too loose.** `"codex" in cmd_str` would match a model name like `codex-bespoke`. Consider asserting the path ends with `"codex"` or use a more specific pattern. |
| W3 | **Dead code.** The `or "codex"` fallback in `run()` is unreachable because `run()` returns early when `_cli_available` is False. Either remove the fallback or add a comment. |

### Minor suggestions

- Mock `sys.platform` in tests so the Windows code path runs on Linux CI too
- Update the file header comment with the modification date
- The `-` stdin support might be version-gated in Codex CLI. Worth documenting which version introduced it, or at least noting the assumption
- The `claude_sdk.py` formatting change (3-line `bool(...)` to 1-line) is correct but unrelated to this fix. Cleaner to keep style fixes in a separate commit

### Bottom line

Solid approach, well-targeted fix. Address C1 (model name validation) and C2 (import placement), and this is good to merge. W1 (broken pipe handling) is worth doing here too since it's a user-facing error path.

DevRohit06:
Hey @prakashUXtech, thanks for the detailed review! All items addressed in 86c1600e:

### Blockers
- **C1** - Model names are now validated with `re.match(r'^[\w\-.:]+$', model)` before being passed to any subprocess call. Malicious names like `gpt-4" & dir` are rejected with a clear error message. Tests cover both valid and invalid patterns.
- **C2** - `import subprocess` (and `import re`) moved to top-level imports alongside the other stdlib modules.

### Warnings
- **W1** - Added dedicated `BrokenPipeError`/`ConnectionResetError` handling around the stdin write. If Codex CLI crashes before reading the prompt, users now get a clear message ("Codex CLI exited before reading the prompt") with stderr output appended if available. Test included.
- **W2** - Tightened the Windows assertion from `"codex" in cmd_str` to `cmd_str.split()[0].endswith("codex")`, so it matches only the binary path.
- **W3** - Removed the `or "codex"` fallback on `codex_bin` since `run()` returns early when `_cli_available` is False (making the fallback unreachable).

### Minor
- Added a note in the module docstring about the stdin/`-` support and the Codex CLI version that introduced it (v0.1.2504)
- The unrelated `claude_sdk.py` formatting change was already in the branch from a prior commit, keeping it as-is since it's a one-line style fix

Ready for another look!

PR #579, opened 2026-03-11
[FI] Fix WhatsApp Cancel button during QR pairing (#345)

### LINKED ISSUE #345: In the channel, if I click on WhatsApp and then click on Start and Pair, the Cancel button does not work.
### Operating System

windows 11

### Python Version

Python 3.14.3

### PocketPaw Version

pocketpaw 0.4.4

### Installation Method

pip install pocketpaw

### Bug Description

In the channel, if I click on WhatsApp and then click on Start and Pair, the Cancel button does not work.

### Steps to Reproduce

1.python -m pocketpaw
2.open the dashboard
3.click on the channel 
4.whatsapp
4.start and pair click
5.cancel

### Expected Behavior

cancel worked

### Actual Behavior

I have fixed this bug.

I have fixed this bug.

### Relevant Files

pocketpaw\src\pocketpaw\frontend\js\features\channels.js
pocketpaw\src\pocketpaw\frontend\templates\components\modals\channels.html
https://github.com/sanket122461/pocketpaw

### Before submitting

manishgupta2026:
## What does this PR do?

Fixes the WhatsApp Cancel button during QR pairing. The `cancelInstall()` method was being called in channels.html but the method was also named `cancelInstall()` in channels.js — this naming ambiguity caused the Cancel button to not work correctly. Renamed the install prompt cancel handler to `cancelInstallPrompt()` in both the JS method definition and the HTML @click bindings.

## Related Issue

Closes #345

## Checklist

- [x] PR targets `dev` branch
- [x] Linked to an issue
- [x] Read CONTRIBUTING.md

### DIFF
--- a/src/pocketpaw/frontend/templates/components/modals/channels.html
+++ b/src/pocketpaw/frontend/templates/components/modals/channels.html
@@ -582,7 +582,7 @@
         x-show="installPrompt"
         x-transition.opacity
         class="absolute inset-0 z-10 flex items-center justify-center bg-black/60 backdrop-blur-sm rounded-[20px]"
-        @click.self="cancelInstall()"
+        @click.self="cancelInstallPrompt()"
     >
         <div class="w-full max-w-[340px] mx-4 bg-[#2a2a2e] border border-[var(--glass-border)] rounded-[16px] shadow-2xl p-5 flex flex-col gap-4" @click.stop>
             <!-- Icon + message -->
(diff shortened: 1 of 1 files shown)


yuktaa123 (approved):
Thanks for the fix. The rename resolves the method collision, and the HTML bindings are updated correctly. Looks good to me.

DevRohit06:
Hey, thanks for the PR! I took a closer look and I have some concerns about whether this actually fixes the issue.

`cancelInstall()` in `channels.js` and `cancelInstall()` in `mcp.js` live in completely separate Alpine `x-data` scopes. They don't conflict with each other, so there's no naming ambiguity causing a bug here. The rename from `cancelInstall` to `cancelInstallPrompt` is cosmetic and doesn't change any behavior since the method body is identical.

If the WhatsApp Cancel button during QR pairing is actually broken (#345), the root cause is likely somewhere else. Could you share steps to reproduce the original bug? That would help us figure out the real fix.

manishgupta2026:
Thanks for the detailed review @DevRohit06 — you're right, and I appreciate the correction.

I misidentified the root cause. Since `cancelInstall()` in `channels.js` and `mcp.js` live in separate Alpine.js `x-data` scopes, there's no naming conflict, and the rename is indeed cosmetic with no behavioral change.

I'll look more closely at the actual WhatsApp Cancel button wiring in the HTML and trace the real cause of the bug in #345 before pushing an updated fix. I'll update this PR or open a new one once I've confirmed the actual root cause.

PR #589, opened 2026-03-12
[FI] fix(usage_tracker): include cached_input in total_tokens and fix…

### LINKED ISSUE #590: bug(usage_tracker): total_tokens excludes cached_input_tokens and get_summary() caps at 10,000 records
### Operating System

Windows 11 / Linux (cross-platform Python bug)

### Python Version

Python 3.11+

### PocketPaw Version

0.4.9

### Installation Method

pip install pocketpaw

### Bug Description

Two silent bugs found in `src/pocketpaw/usage_tracker.py`:

**Bug 1:** `total_tokens` in `record()` is computed as `input_tokens + output_tokens`, completely dropping `cached_input_tokens` from the count. Cached tokens are real tokens processed by the model (just billed at a lower rate), so this causes `total_tokens` to be systematically understated whenever prompt caching is used.

**Bug 2:** `get_summary()` internally calls `get_records(limit=10_000)`, meaning any installation with more than 10,000 lifetime usage records will silently produce wrong (understated) aggregation totals in the summary.

### Steps to Reproduce

**Bug 1:**
1. Call `tracker.record()` with `cached_input_tokens=200`, `input_tokens=100`, `output_tokens=50`
2. Inspect the returned `UsageRecord.total_tokens`
3. Observe it returns `150` instead of the correct `350`

**Bug 2:**
1. Write more than 10,000 records to the usage JSONL file
2. Call `tracker.get_summary()`
3. Observe `request_count` is capped at 10,000 instead of the actual count

### Expected Behavior

**Bug 1:** `total_tokens` should equal `input_tokens + output_tokens + cached_input_tokens` (i.e., `350` in the example above).

**Bug 2:** `get_summary()` should aggregate ALL records ever written, not just the most recent 10,000.

### Actual Behavior

**Bug 1:** `total_tokens = input_tokens + output_tokens` — cached tokens are silently dropped. With `input=100, output=50, cached=200`, result is `150` instead of `350`.

**Bug 2:** `get_summary()` calls `get_records(limit=10_000)`, so installations with >10,000 records get silently truncated summaries.

```python
# Bug 1 reproduction
tracker = UsageTracker(path=tmp_path / "usage.jsonl")
rec = tracker.record(backend="anthropic", model="claude-3-5-sonnet-20241022",
    input_tokens=100, output_tokens=50, cached_input_tokens=200)
print(rec.total_tokens)  # prints 150 — should be 350

# Bug 2 reproduction
# Write 10_001 records then:
summary = tracker.get_summary()
print(summary["request_count"])  # prints 10000 — should be 10001
```

### Relevant Files

src/pocketpaw/usage_tracker.py (record() method line ~128, get_summary() method line ~178)

### Before submitting

rutujgiri2512:
## What does this PR do?

Fixes two silent bugs in `UsageTracker` that caused incorrect token counts and wrong usage summaries.

1. **`total_tokens` excluded `cached_input_tokens`**: In `record()`, total was computed as `input_tokens + output_tokens`, silently dropping cached tokens from the count even though they are real tokens processed by the model (just billed at a lower rate).

2. **`get_summary()` capped at 10,000 records**: It called `get_records(limit=10_000)` instead of reading all records, so any installation with more than 10,000 lifetime records would silently produce understated aggregation totals.

## Related Issue

Fixes #590

## Changes Made

- `src/pocketpaw/usage_tracker.py`: Fixed `record()` to include `cached_input_tokens` in `total_tokens`. Added `_iter_all_records()` helper that reads every record from disk. Updated `get_summary()` to use `_iter_all_records()` instead of the capped `get_records()`.
- `tests/test_usage_tracker.py`: Added 210-line test suite covering both bugs with edge cases (zero cached tokens, 10,001+ records, `since` filter, cost estimation).

## How to Test

1. Clone the repo and install dev deps: `uv sync --dev`
2. Run the new tests: `uv run pytest tests/test_usage_tracker.py -v`
3. All tests should pass green.

## Evidence of Testing

Tests written in `tests/test_usage_tracker.py` cover both bugs with specific assertions (e.g., `total_tokens == 350` when cached=200, `request_count == 10_001` for large datasets).

## Checklist

### DIFF
--- a/src/pocketpaw/agents/codex_cli.py
+++ b/src/pocketpaw/agents/codex_cli.py
@@ -34,6 +34,11 @@ logger = logging.getLogger(__name__)
 # Only allow safe characters in model names to prevent shell injection
 _MODEL_NAME_RE = re.compile(r"^[\w\-.:]+$")
 
+# 10 MiB buffer for subprocess stdout. Codex CLI emits NDJSON events that can
+# exceed the asyncio default of 64 KiB (e.g., large MCP tool results from
+# Playwright, code completions, etc.).
+_SUBPROCESS_BUFFER_LIMIT = 10 * 1024 * 1024
+
 
 class CodexCLIBackend:
     """Codex CLI backend — subprocess wrapper for OpenAI's terminal AI agent."""
@@ -178,6 +183,7 @@ class CodexCLIBackend:
                     stdout=asyncio.subprocess.PIPE,
                     stderr=asyncio.subprocess.PIPE,
                     env=proc_env,
+                    limit=_SUBPROCESS_BUFFER_LIMIT,
                 )
             else:
                 self._process = await asyncio.create_subprocess_exec(
@@ -187,6 +193,7 @@ class CodexCLIBackend:
                     stdout=asyncio.subprocess.PIPE,
                     stderr=asyncio.subprocess.PIPE,
                     env=proc_env,
+                    limit=_SUBPROCESS_BUFFER_LIMIT,
                 )
 
             # Feed the prompt via stdin and close to signal EOF
@@ -350,6 +357,11 @@ class CodexCLIBackend:
             self._process = None
             yield AgentEvent(type="done", content="")
 
+        except (asyncio.LimitOverrunError, asyncio.IncompleteReadError) as e:
+            logger.warning("Codex CLI session terminated: stdout buffer exceeded: %s", e)
+            self._process = None
+            yield AgentEvent(type="error", content="Codex CLI output exceeded buffer limit")
+            yield AgentEvent(type="done", content="")
         except Exception as e:
             logger.error("Codex CLI error: %s", e)
             yield AgentEvent(type="error", content=f"Codex CLI error: {e}")
--- a/tests/test_codex_cli_backend.py
+++ b/tests/test_codex_cli_backend.py
@@ -763,6 +763,96 @@ class TestCodexCLIValidation:
         assert "segfault" in errors[0].content
 
 
+class TestCodexCLIBufferLimit:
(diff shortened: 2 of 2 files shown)


DevRohit06 (commented):
Hey, nice catch on both bugs. Verified the cached_input_tokens fix is correct -- in claude_sdk.py, `input_tokens` comes from the API's non-cached count and `cached_input_tokens` is `cache_read + cache_creation`, so no double-counting. Tests are solid too.

One thing to clean up:

**`_iter_all_records` duplicates `get_records` logic.** The two methods have nearly identical parsing code. Instead of a separate method, just make the limit optional on `get_records`:

```python
def get_records(self, limit: int | None = 100) -> list[UsageRecord]:
```

Then `get_summary` calls `get_records(limit=None)`. One method, no duplication.

Other than that, this looks good.

DevRohit06 (commented):
Good fix, both bugs are real and the tests are solid. A few suggestions:

### Code duplication

`_iter_all_records()` duplicates most of the file-reading logic from `get_records()`. Consider extracting a shared private method:
```python
def _read_records(self, limit: int | None = None, reverse: bool = False) -> list[UsageRecord]:
```

### Minor

- `_iter_all_records` returns a `list`, not an iterator. Consider renaming to `_read_all_records`.
- `get_records()` uses `.strip().split("\n")` while `_iter_all_records()` uses `.splitlines()`. These behave differently with edge cases. Use `.splitlines()` consistently.
- For very large usage files, reading the entire file into memory could be expensive. A line-by-line reader would be more memory-efficient, but not a blocker for now.

Overall this is clean and merge-worthy. The suggestions above are non-blocking improvements.

PR #593, opened 2026-03-13
[FI] Avoid import-time writes during dashboard setup

### LINKED ISSUE #592: Importing dashboard or create_api_app should not write to ~/.pocketpaw
## Bug
Importing pocketpaw.dashboard or calling create_api_app() currently writes to ~/.pocketpaw during import/load.

In read-only homes, sandboxed CI, or locked-down containers this raises PermissionError before startup.

## Reproduction
1. Use a non-writable HOME.
2. Run python -c "import pocketpaw.dashboard" or python -c "from pocketpaw.api.serve import create_api_app; create_api_app()".
3. Observe a failure while trying to create ~/.pocketpaw.

## Expected
Importing modules or creating the API app should not mutate the user home directory. Runtime loop initialization and config directory creation should happen only on explicit startup or write paths.

## Notes
This appears to come from eager dashboard singleton initialization plus settings-load migration and config-dir logic running on read paths.

tohin003:
## What does this PR do?

Prevents dashboard-related read paths from writing to ~/.pocketpaw during import or app creation. This fixes PermissionError failures in read-only homes, sandboxed CI, and other restricted environments before PocketPaw startup has even begun.

## Related Issue

Fixes #592

## Changes Made

- src/pocketpaw/config.py: allow config-path helpers to read without creating ~/.pocketpaw, and make plaintext-key migration best-effort on read-only homes
- src/pocketpaw/dashboard_state.py: replace eager AgentLoop singleton construction with a lazy proxy so importing dashboard modules does not initialize memory/config state
- tests/test_import_side_effects.py: add subprocess regressions for Settings.load(), create_api_app(), and import pocketpaw.dashboard under a non-writable HOME
- tests/test_credentials.py: update config-dir monkeypatches to stay compatible with the non-creating read path

## How to Test

1. Use a non-writable HOME and run python -c "import pocketpaw.dashboard"
2. Use a non-writable HOME and run python -c "from pocketpaw.api.serve import create_api_app; create_api_app()"
3. Verify both commands succeed and do not create ~/.pocketpaw
4. Run the updated regression tests

## Evidence of Testing

```
./.venv/bin/python -m ruff check src/pocketpaw/config.py src/pocketpaw/dashboard_state.py tests/test_credentials.py tests/test_import_side_effects.py
All checks passed!

HOME=$(pwd)/.home ./.venv/bin/python -m pytest -q tests/test_import_side_effects.py tests/test_credentials.py tests/test_api_serve.py tests/test_api_v1_channels.py
75 passed, 64 warnings in 9.24s

HOME=$(pwd)/.home ./.venv/bin/python -m pytest -q tests --ignore=tests/e2e --ignore=tests/test_launcher_server.py --ignore=tests/test_neonize_adapter.py --ignore=tests/test_teams_adapter.py --ignore=tests/test_tools.py --ignore=tests/test_tools_sysinfo.py
2918 passed, 26 skipped, 68 warnings in 35.67s
```

## Checklist

- [ ] Tests pass (`uv run pytest --ignore=tests/e2e`)
- [ ] Linting passes (`uv run ruff check .`)

### DIFF
--- a/src/pocketpaw/config.py
+++ b/src/pocketpaw/config.py
@@ -124,23 +124,38 @@ def _warn_old_config() -> None:
         )
 
 
-def get_config_dir() -> Path:
-    """Get the config directory, creating if needed."""
+def get_config_dir(create: bool = True) -> Path:
+    """Get the config directory.
+
+    Args:
+        create: When True, create the directory if needed. Read-only code paths
+            should pass False so importing or loading settings does not mutate
+            the user's home directory.
+    """
     config_dir = Path.home() / ".pocketpaw"
-    config_dir.mkdir(exist_ok=True)
-    _chmod_safe(config_dir, 0o700)
+    if create:
+        config_dir.mkdir(exist_ok=True)
+        _chmod_safe(config_dir, 0o700)
     _warn_old_config()
     return config_dir
 
 
-def get_config_path() -> Path:
+def _resolve_config_dir(*, create: bool) -> Path:
+    """Call ``get_config_dir`` while tolerating older no-arg monkeypatches in tests."""
+    try:
+        return get_config_dir(create=create)
+    except TypeError:
+        return get_config_dir()
+
+
+def get_config_path(create: bool = True) -> Path:
     """Get the config file path."""
-    return get_config_dir() / "config.json"
+    return _resolve_config_dir(create=create) / "config.json"
 
 
-def get_token_path() -> Path:
+def get_token_path(create: bool = True) -> Path:
     """Get the access token file path."""
-    return get_config_dir() / "access_token"
+    return _resolve_config_dir(create=create) / "access_token"
 
 
 # Telegram bot token format: numeric id + colon + alphanumeric secret
@@ -744,7 +759,7 @@ class Settings(BaseSettings):
         # Run one-time migration from plaintext config
         _migrate_plaintext_keys()
 
-        config_path = get_config_path()
+        config_path = get_config_path(create=False)
         data: dict = {}
         if config_path.exists():
             try:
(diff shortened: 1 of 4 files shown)


ANSHSINGH050404:
please put screenshorts @yuktaa123

DevRohit06 (dismissed):
Hey, solid goal here. Making config reads non-mutating is the right call. A few things to address:

**1. `_LazyAgentLoopProxy` breaks `isinstance` and type checking**

`agent_loop` is passed to `_get_cmd_handler().set_agent_loop(agent_loop)` right after creation. If anything downstream does `isinstance(loop, AgentLoop)`, it'll fail. Type checkers also won't know what it is. A simpler approach would be to defer the `set_agent_loop` call too, or use a module-level `get_agent_loop()` function instead of a magic proxy.

**2. Broad `except Exception` in migration swallows real errors**

```python
except Exception:
    logger.debug("Skipping plaintext key migration in read-only config dir", exc_info=True)
    return
```

This catches everything, not just filesystem errors. A corrupted credential store, a bug in `store.set()`, or a missing dependency would all be silently swallowed at `DEBUG` level. Narrow this to `OSError` or `PermissionError`.

**3. `_resolve_config_dir` exists only to tolerate broken monkeypatches**

This is test infrastructure leaking into production code. If tests monkeypatch `get_config_dir`, they should match the real signature. The wrapper adds indirection for no production benefit. Just update the test monkeypatches (which you're already doing) and drop this wrapper.

**4. Proxy `__setattr__` eagerly initializes on any attribute set**

If any code does `agent_loop.some_attr = value` before the loop is actually needed, it triggers full `AgentLoop()` construction, defeating the lazy intent.

---

The config.py changes are clean, the `_MIGRATION_DONE_PATH` global removal is nice. The lazy proxy is the weak spot. Would suggest either narrowing the proxy to only delay construction until `start()`/`run()` is called, or switching to a simple factory function pattern.

DevRohit06 (changes requested):
Previous review concerns still apply. Please address: narrow the except Exception to OSError/PermissionError, remove the _resolve_config_dir compatibility shim (update test monkeypatches instead), and consider proxy limitations (isinstance, __setattr__ eager init). See earlier review for full details.

DevRohit06:
Ignore the earlier approve - that was posted by mistake. The changes requested review is the current status. Please address the outstanding feedback (narrow except Exception, remove _resolve_config_dir shim) before this can be merged.

PR #614, opened 2026-03-14
[FI] - Fix socket resource leak in IP detection

tanmay-563:
fix(api|dashboard): Fix socket resource leak in IP detection

## Problem
Socket created in IP detection (`socket.socket()`) doesn't close if an 
exception occurs during `connect()` or `getsockname()` calls, causing 
resource leaks on long-running servers.

Affected files:
- `src/pocketpaw/api/serve.py` (line 149)
- `src/pocketpaw/dashboard.py` (line 1682)

## Impact
On production servers running multiple restarts or with network issues, 
accumulated socket leaks can exhaust system file descriptor limits 
(typically ~1024 per process), eventually causing:
- Service startup failures
- "Too many open files" errors
- Connection rejections

## Solution
Replace try/except pattern with context manager (`with` statement) to 
guarantee socket cleanup even on exception.

## Testing
✅ Socket properly closes even if connect() fails
✅ Multiple startup cycles don't accumulate open file descriptors
✅ Error handling still works (fallback to "<your-server-ip>")

Fixes: #[608]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,4 @@ client/build/
 client/src-tauri/target/
 client/src-tauri/gen/
 client/bun.lock
+pocketpaw-env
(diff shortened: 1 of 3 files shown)


prakashUXtech:
This PR includes 1.17M additions — likely committed node_modules or lock files. Please reopen with a clean diff targeting dev.

PR #617, opened 2026-03-14
fix: harden local url extraction against SSRF

anish1301:
## What does this PR do?

This PR fixes an SSRF vulnerability in the local URL extraction path 
by validating URLs before fetches and checking every redirect target 
before following it. It restricts allowed schemes to HTTP and HTTPS 
only, blocks private and reserved destinations, and disables automatic 
redirects in favor of manual checked traversal.

## Related Issue

Fixes #[INSERT_ISSUE_NUMBER_HERE]

## Changes Made

- `src/pocketpaw/tools/builtin/url_extract.py`: Added scheme validation 
  (http/https only)
- `src/pocketpaw/tools/builtin/url_extract.py`: Added DNS resolution and 
  IP classification to block private, loopback, link-local, multicast, 
  reserved, unspecified destinations
- `src/pocketpaw/tools/builtin/url_extract.py`: Disabled httpx automatic 
  redirects
- `src/pocketpaw/tools/builtin/url_extract.py`: Added manual redirect 
  handling with per-hop validation and 5-hop limit

## How to Test

1. Configure URL extraction to use local provider
2. Verify normal public URLs still work: `https://example.com`
3. Verify internal targets are blocked: `http://127.0.0.1:8000` (should fail)
4. Verify redirects to internal are blocked: any public URL → redirect to 
   `http://169.254.169.254/` (should fail)

## Evidence of Testing

Tested on Python 3.13.5, Windows 11:
- Scheme validation: ✓ http/https allowed, others blocked
- IP blocking: ✓ 127.0.0.1, 10.x.x.x, 169.254.x.x, ::1 all blocked
- Redirect validation: ✓ Each hop validated before fetch
- New test framework ready (3 existing test mocks require alignment)

## Checklist

- [x] Tests updated (3 redirects mocks aligned with new validation)
- [x] Linting passes (ruff check)
- [x] No secrets or credentials in diff
- [x] No unrelated changes bundled

### DIFF
--- a/src/pocketpaw/tools/builtin/url_extract.py
+++ b/src/pocketpaw/tools/builtin/url_extract.py
@@ -43,9 +43,12 @@ def _is_private_ip(addr: str) -> bool:
     )
 
 
-async def _validate_url(url: str) -> None:
-    """Validate a URL against SSRF attacks.
+async def _validate_url(url: str) -> tuple[str, int]:
+    """Validate a URL against SSRF attacks and return resolved IP and port.
 
+    Returns the resolved IP address and port for safe use in subsequent requests.
+    This prevents DNS TOCTOU attacks by ensuring the validated IP is used for the request.
+    
     Raises ValueError if the URL targets a private/reserved IP, uses a
     disallowed scheme, or cannot be resolved.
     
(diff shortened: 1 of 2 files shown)


anish1301:
@DevRohit06  I tried to do whatever You said so , I hope this fixes the issue. 
Btw I have found a lot of security related issues . I need to test them  but certainly I am quiet engaged with other stuff , so is there anyone with whom I can communicate to fix these issues .

anish1301:
> Good progress across 3 rounds. DNS TOCTOU, async DNS, error message leaking, and test coverage are all addressed. Two issues remain:
> 
> ### Blocker: HTTPS breaks with IP-rewritten URLs
> `_safe_get` rewrites `https://example.com/page` to `https://93.184.216.34:443/page` with a `Host: example.com` header. TLS certificate validation will fail because the cert is issued for `example.com`, not `93.184.216.34`. The TLS handshake checks the hostname against the certificate's CN/SAN, and an IP address won't match.
> 
> This breaks every HTTPS URL in local extraction mode.
> 
> To fix this properly, you need to keep the hostname in the URL for TLS/SNI but pin the connection to the resolved IP at the transport level. Options:
> 
> * Use a custom `httpx.AsyncBaseTransport` that overrides DNS resolution
> * Or use httpx's `transport` with a `socket_options` / connection pool that pins the resolved address
> 
> ### Bug: Indentation error in `_extract_local`
> ```python
> async def _extract_local(self, urls: list[str]) -> str:
>     try:
> -            import html2text
> +         import html2text
> ```
> 
> The `import html2text` went from 12 spaces (correct) to 9 spaces. This will cause an `IndentationError` at runtime.
> 
> ### Minor: test hitting real DNS
> `test_validate_url_loopback_blocked` calls `_validate_url("http://127.0.0.1")` without mocking DNS resolution. It works because `getaddrinfo` resolves IP literals directly, but it's fragile and inconsistent with the other tests that mock DNS. Consider mocking it for consistency.
> 
> The HTTPS issue is the main blocker. The IP-rewrite approach fundamentally conflicts with TLS. A transport-level solution would fix both the TOCTOU and TLS problems cleanly.

Damnn @DevRohit06  didn't think this through in so much depth , no wonder why you have 100k + lines committed !!
Will work on this one and raise a pr soon !

PR #663, opened 2026-03-17
fix(scheduler): prevent corrupt reminders overwrite on startup

### LINKED ISSUE #659: bug: Silent exception swallowing in `load_reminders()` silently destroys all saved reminders on corrupt JSON
### Operating System

All platforms (Windows, macOS, Linux)

### Python Version

Python 3.11+

### PocketPaw Version

0.4.9

### Installation Method

pip install pocketpaw

### Bug Description

In `src/pocketpaw/scheduler.py`, the `load_reminders()` function silently swallows all JSON parse errors with a bare `except Exception: pass`:

```python
def load_reminders() -> list[dict]:
    path = get_reminders_path()
    if path.exists():
        try:
            data = json.loads(path.read_text())
            return data.get("reminders", [])
        except Exception:
            pass  # silently swallowed — no log, no warning
    return []
```

When `~/.pocketpaw/reminders.json` becomes corrupted (crash mid-write, disk full, encoding error, or bad manual edit), this function silently returns an empty list. The scheduler then starts with zero reminders and immediately calls `save_reminders([])`, **permanently overwriting and destroying all saved reminders** with no log entry, no warning, and no recovery path. This is a silent data-loss bug.

### Steps to Reproduce

1. Create one or more reminders via the dashboard or messaging channel
2. Verify they appear in ~/.pocketpaw/reminders.json
3. Corrupt the file: echo "{bad json" > ~/.pocketpaw/reminders.json
4. Restart PocketPaw: uv run pocketpaw
5. Observe: no error, no warning, no log — all reminders are silently gone
6. Check ~/.pocketpaw/reminders.json — it has been overwritten with {"reminders": [], "updated_at": "..."}

### Expected Behavior

`load_reminders()` should log a warning when JSON parsing fails so operators know the file is corrupted. It should also avoid immediately overwriting the corrupted file. At minimum:

```python
except Exception:
    logger.warning("Failed to load reminders from %s — file may be corrupted", path, exc_info=True)
```

Ideally, the corrupted file should be renamed to `reminders.json.bak` before starting fresh, to allow manual recovery.

### Actual Behavior

No log output. No warning. PocketPaw starts normally with zero reminders, and `~/.pocketpaw/reminders.json` is silently overwritten with `{"reminders": [], "updated_at": "..."}`. All previously saved reminders are permanently lost with no indication anything went wrong.

$ echo '{"bad json":' > ~/.pocketpaw/reminders.json
$ uv run pocketpaw
# ... normal startup output, no errors, no warnings ...
# All reminders silently gone. File reset to {"reminders": []}

### Relevant Files

src/pocketpaw/scheduler.py — load_reminders() (~L44-52), save_reminders() (~L57-60)

### Before submitting

Shivasai132678:
## Summary
Closes #659.

This fixes silent reminder data loss when `~/.pocketpaw/reminders.json` is corrupted.

### What changed
- Added `RemindersCorruptError` for explicit corruption signaling.
- Added `_quarantine_corrupt_reminders()` to move unreadable/invalid reminder payloads to:
  - `reminders.json.corrupt-<timestamp>`
- Updated `load_reminders()` to:
  - log parse/payload failures
  - quarantine the corrupt file
  - raise `RemindersCorruptError` instead of silently returning `[]`
- Updated `ReminderScheduler.start()` to fail safely:
  - continue startup with empty in-memory reminders
  - **skip startup write** when load failed, preventing overwrite of recoverable corrupt data

## Tests
Added regression tests in `tests/test_scheduler_cron.py`:
- `test_load_reminders_quarantines_corrupt_json`
- `test_start_does_not_overwrite_when_load_is_corrupt`

## Validation
- `uv run ruff check .`
- `uv run pytest --ignore=tests/e2e -q`
  - `3279 passed, 19 skipped`

### DIFF
--- a/src/pocketpaw/scheduler.py
+++ b/src/pocketpaw/scheduler.py
@@ -10,6 +10,7 @@ import uuid
 from collections.abc import Callable
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
+from typing import NoReturn
 
 from apscheduler.schedulers.asyncio import AsyncIOScheduler
 from apscheduler.triggers.cron import CronTrigger
@@ -55,7 +56,9 @@ def _quarantine_corrupt_reminders(path: Path) -> Path | None:
         return None
 
 
-def _signal_corrupt_reminders(path: Path, reason: str, *, cause: Exception | None = None) -> None:
+def _signal_corrupt_reminders(
+    path: Path, reason: str, *, cause: Exception | None = None
+) -> NoReturn:
     """Quarantine a corrupt reminders file and raise a typed corruption error."""
     backup = _quarantine_corrupt_reminders(path)
     if backup:
@@ -92,7 +95,12 @@ def _validate_reminder_entry_schema(reminder: object, index: int) -> str | None:
 
     reminder_type = reminder.get("type", "one-shot")
     if reminder_type not in ("one-shot", "recurring"):
-        return f"Reminder entry at index {index} has invalid 'type': {reminder_type!r}"
+        # Unknown types may be valid in future versions — warn and treat as one-shot.
+        logger.warning(
+            "Reminder entry at index %d has unrecognised 'type': %r — treating as one-shot",
+            index,
+            reminder_type,
+        )
 
     if reminder_type == "recurring":
         schedule = reminder.get("schedule")
@@ -121,11 +129,27 @@ def load_reminders() -> list[dict]:
         if not isinstance(reminders, list):
             _signal_corrupt_reminders(path, "Invalid reminders payload (expected list)")
 
+        valid_reminders: list[dict] = []
         for index, reminder in enumerate(reminders):
             validation_error = _validate_reminder_entry_schema(reminder, index)
             if validation_error:
-                _signal_corrupt_reminders(path, validation_error)
-        return reminders
+                logger.warning(
+                    "Skipping malformed reminder entry at index %d: %s", index, validation_error
+                )
+            else:
+                valid_reminders.append(reminder)
+
+        skipped = len(reminders) - len(valid_reminders)
+        if skipped:
+            logger.warning(
+                "Loaded %d valid reminder(s) from %s; skipped %d malformed entr%s",
+                len(valid_reminders),
+                path,
+                skipped,
+                "y" if skipped == 1 else "ies",
+            )
+
+        return valid_reminders
     return []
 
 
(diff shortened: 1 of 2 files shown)


Shivasai132678:
Addressed Copilot feedback in latest commit `3224f59`:

- Made corrupt-file quarantine filenames collision-resistant by adding microseconds + random suffix.
- Hardened decode/parsing path:
  - `read_text(encoding="utf-8")`
  - treat `UnicodeDecodeError` like parse corruption (quarantine + `RemindersCorruptError`).
- Treat non-object JSON roots as corrupt payloads (quarantine + raise), instead of silently loading `[]`.
- Added reminder-entry schema validation during load (type/required fields/trigger format), with quarantine + raise on invalid entries.

Added regression tests for:
- non-dict root payload
- non-UTF8 bytes
- invalid reminder entry schema
- unique quarantine file creation across repeated corrupt loads

Validation rerun:
- `uv run ruff check src/pocketpaw/scheduler.py tests/test_scheduler_cron.py`
- `uv run pytest -q tests/test_scheduler_cron.py`
- `uv run pytest --ignore=tests/e2e -q` (`3283 passed, 19 skipped`)

Shivasai132678:
Thanks for the detailed review @DevRohit06! All points have been addressed in commit `3b994e27`. Here's a summary of every change made:

### Must fix ✅

1. **All-or-nothing quarantine → per-entry skip/log** — `load_reminders()` now iterates through each entry individually. Invalid entries are logged as warnings and skipped; all valid entries are kept. The file is only quarantined when the JSON itself is unparseable or the root is not a `{"reminders": [...]}` dict.

2. **`_signal_corrupt_reminders` → `NoReturn`** — Added `from typing import NoReturn` and annotated the function `-> NoReturn`. Control flow after calls to it is now unambiguous.

3. **`encoding="utf-8"` on `save_reminders()`** — `path.write_text(...)` now passes `encoding="utf-8"` explicitly, matching the read side.

### Should fix ✅

4. **Unknown `type` values warn instead of quarantine** — Per-entry validation now uses `logger.warning(...)` for unrecognised reminder types rather than treating them as corrupt. Forward-compatible with new types added in future releases.

5. **Test gaps filled** — Added 4 new tests in `TestReminderFileCorruption`:
   - `test_partial_corruption_keeps_valid_entries` — mix of valid + invalid entries; only invalid ones dropped
   - `test_load_reminders_oserror_propagates` — `OSError` on read propagates as-is (not swallowed)
   - `test_quarantine_failure_raises` — if `path.replace()` itself fails, `RemindersCorruptError` is re-raised with original context
   - `test_recurring_reminder_missing_schedule_skipped` — recurring entry without `schedule` field is skipped/warned, not quarantined

PR #695, opened 2026-03-20
fix: backup encrypted secrets when machine identity changes

anish1301:
### Description
This addresses a critical data loss bug where renaming my computer caused `secrets.enc` to be completely overwritten with an empty dictionary. 

Instead of silently erasing the secrets when `InvalidToken` is caught (caused by a mismatched `machine_identity`), this fix renames the broken file to `secrets.enc.bak` and logs a verbose warning. It maintains the original security intent (stolen files still cannot be decrypted) while preventing accidental data loss for legitimate users.


DevRohit06 (changes requested):
Backup overwritten on repeat failures, no permissions set, too-broad except clause. See comment.

PR #702, opened 2026-03-21
[FI] fix: correct invalid tool trust levels, fetch.py crash, and PII scanner gaps

Dhruv18052003-web:
# PR Description

## Title
[FI] fix: correct invalid tool trust levels, fetch.py crash, and PII scanner gaps

### Problem

Three independent issues found during codebase analysis:

**1. Invalid `trust_level="elevated"` on `RunPythonTool` and `InstallPackageTool`**

Both tools return `trust_level = "elevated"`, which is not a recognized trust level. The tool system only recognizes `standard`, `high`, and `critical`. In `tools/registry.py`, the trust-to-audit-severity mapping falls through to the `else` branch for unrecognized levels:

```python
# tools/registry.py lines 109-115
if t_level == "critical":
    severity = AuditSeverity.CRITICAL
elif t_level == "high":
    severity = AuditSeverity.WARNING
else:
    severity = AuditSeverity.INFO  # ← "elevated" falls here
```

This means arbitrary Python code execution (`run_python`) and package installation (`install_package`) are audited at `INFO` severity — the same as safe read-only tools.

**2. `get_directory_keyboard()` crashes when `python-telegram-bot` is not installed**

`tools/fetch.py` conditionally imports `InlineKeyboardMarkup`, setting it to `None` when unavailable. But `get_directory_keyboard()` calls `InlineKeyboardMarkup(buttons)` unconditionally, causing `TypeError: 'NoneType' object is not callable` in non-Telegram deployments.

**3. PII scanner misses common patterns**

The PII scanner (`security/pii.py`) only detects dashed SSNs (`123-45-6789`). It misses space-separated SSNs (`123 45 6789`), contextual bare 9-digit SSNs, passport numbers, and IBANs.

### Solution

**1. Trust levels**: Changed `RunPythonTool.trust_level` to `"critical"` (arbitrary code execution, same as `ShellTool`) and `InstallPackageTool.trust_level` to `"high"` (has Guardian + regex validation guardrails). Updated corresponding test assertions.

**2. Fetch crash**: Added early return guard `if InlineKeyboardMarkup is None: return None` at the top of `get_directory_keyboard()`.

**3. PII patterns**: Added `PASSPORT` and `BANK_ACCOUNT` to `PIIType` enum. Added 4 new regex patterns:
- SSN space-separated: `\b\d{3}\s\d{2}\s\d{4}\b`
- SSN contextual bare 9-digit (requires keyword like "SSN" or "social security")
- Passport number (contextual, requires "passport" keyword)
- IBAN (2-letter country code + 2 check digits + alphanumeric body)

### Files Changed

| File | Change |
|---|---|
| `src/pocketpaw/tools/builtin/python_exec.py` | `trust_level`: `"elevated"` → `"critical"` |
| `src/pocketpaw/tools/builtin/pip_install.py` | `trust_level`: `"elevated"` → `"high"` |
| `src/pocketpaw/tools/fetch.py` | Guard `get_directory_keyboard()` for missing telegram |
| `src/pocketpaw/security/pii.py` | Add 2 enum values + 4 regex patterns |
| `tests/test_run_python.py` | Update trust_level assertion |
| `tests/test_install_package.py` | Update trust_level assertion |

### Testing

```
3,211 passed, 39 failed (all pre-existing missing-dep issues), 46 skipped
```

- All 53 directly affected tests pass
- `test_tools.py::TestFetchTool::test_handle_path_directory` — was failing, now passes
- Zero regressions introduced (39 failures are all pre-existing: missing playwright/sarvamai/psutil/aiohttp)
- Ruff lint: all checks passed
- Ruff format: all files already formatted

### Checklist

- [x] Code follows project conventions
- [x] Tests updated for changed behavior
- [x] No new dependencies introduced
- [x] Backward compatible
- [x] Ruff lint passes
- [x] Ruff format passes

### DIFF
--- a/src/pocketpaw/security/pii.py
+++ b/src/pocketpaw/security/pii.py
@@ -23,6 +23,8 @@ class PIIType(StrEnum):
     CREDIT_CARD = "credit_card"
     IP_ADDRESS = "ip_address"
     DATE_OF_BIRTH = "date_of_birth"
+    PASSPORT = "passport"
+    BANK_ACCOUNT = "bank_account"
 
 
 class PIIAction(StrEnum):
@@ -70,6 +72,10 @@ class PIIScanResult:
 _PII_PATTERNS: list[tuple[str, PIIType, int]] = [
     # SSN: 123-45-6789 (dashed format only — bare 9-digit has too many false positives)
     (r"\b\d{3}-\d{2}-\d{4}\b", PIIType.SSN, 0),
+    # SSN: 123 45 6789 (space-separated)
+    (r"\b\d{3}\s\d{2}\s\d{4}\b", PIIType.SSN, 0),
+    # SSN: contextual bare 9-digit (requires keyword nearby)
+    (r"(?:ssn|social security)\s*(?:number|num|no|#)?[\s:]*\b\d{9}\b", PIIType.SSN, re.IGNORECASE),
     # Email addresses
     (r"\b[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}\b", PIIType.EMAIL, 0),
     # US phone: (555) 123-4567, 555-123-4567, 555.123.4567, +1 555-123-4567
@@ -96,6 +102,14 @@ _PII_PATTERNS: list[tuple[str, PIIType, int]] = [
         PIIType.DATE_OF_BIRTH,
         re.IGNORECASE,
     ),
+    # Passport number (contextual — requires "passport" keyword)
+    (
+        r"(?:passport)\s*(?:number|num|no|#)?[\s:]*\b[A-Z0-9]{6,9}\b",
+        PIIType.PASSPORT,
+        re.IGNORECASE,
+    ),
+    # IBAN (2-letter country code + 2 check digits + up to 30 alphanumeric)
+    (r"\b[A-Z]{2}\d{2}[A-Z0-9]{4,30}\b", PIIType.BANK_ACCOUNT, 0),
 ]
 
 _COMPILED_PII: list[tuple[re.Pattern, PIIType]] = [
--- a/src/pocketpaw/tools/builtin/pip_install.py
+++ b/src/pocketpaw/tools/builtin/pip_install.py
@@ -37,7 +37,7 @@ class InstallPackageTool(BaseTool):
 
     @property
     def trust_level(self) -> str:
-        return "elevated"
+        return "high"
 
     @property
     def parameters(self) -> dict[str, Any]:
--- a/src/pocketpaw/tools/builtin/python_exec.py
+++ b/src/pocketpaw/tools/builtin/python_exec.py
@@ -28,7 +28,7 @@ class RunPythonTool(BaseTool):
 
     @property
     def trust_level(self) -> str:
-        return "elevated"
+        return "critical"
 
     @property
     def parameters(self) -> dict[str, Any]:
--- a/src/pocketpaw/tools/fetch.py
+++ b/src/pocketpaw/tools/fetch.py
@@ -21,6 +21,9 @@ def is_safe_path(path: Path, jail: Path) -> bool:
 
 def get_directory_keyboard(path: Path, jail: Path | None = None) -> InlineKeyboardMarkup:
     """Generate inline keyboard for directory contents."""
+    if InlineKeyboardMarkup is None:
+        return None
+
     if jail is None:
         jail = Path.home()
 
--- a/tests/test_install_package.py
+++ b/tests/test_install_package.py
@@ -257,7 +257,7 @@ def test_install_package_definition():
     defn = tool.definition
 
     assert defn.name == "install_package"
-    assert defn.trust_level == "elevated"
+    assert defn.trust_level == "high"
 
     props = defn.parameters["properties"]
     assert "package" in props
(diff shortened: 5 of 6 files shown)


DevRohit06:
## Code Review — First Review

Commit reviewed: [`12b4167`](https://github.com/pocketpaw/pocketpaw/commit/12b416752402f84973f37b241548e6d62d35ab85)

The trust-level and fetch.py fixes are solid. The PII additions have a significant false-positive problem with the IBAN pattern and a type annotation regression that need addressing before merge.

---

### [MUST FIX] IBAN regex produces severe false positives — Score: 95

**File:** `src/pocketpaw/security/pii.py`

```python
# IBAN (2-letter country code + 2 check digits + up to 30 alphanumeric)
(r"\b[A-Z]{2}\d{2}[A-Z0-9]{4,30}\b", PIIType.BANK_ACCOUNT, 0),
```

The pattern `[A-Z]{2}\d{2}[A-Z0-9]{4,30}` matches anything that starts with two uppercase letters, two digits, and four or more alphanumeric characters. In practice this fires on a huge range of real-world strings that are emphatically not IBANs:

- Error codes: `EU01BADREQUEST`, `DE01BUILD123`
- Document IDs / build refs: `US20231231REPORT`, `GH01MAIN0001`
- API keys: any key that starts with two uppercase letters followed by two digits
- UUID fragments, version strings, base tokens

Confirmed with local testing — every one of these produces a false positive match:

```
IP192168001001  ->  MATCH
US20231231REPORT  ->  MATCH
AB12XYZFOO123  ->  MATCH
EU01BADREQUEST  ->  MATCH
DE01BUILD123  ->  MATCH
```

IBANs have a verifiable check-digit algorithm (ISO 7064 MOD-97-10). Regex alone cannot distinguish a real IBAN from a code that happens to look like one. The existing SSN comment in the file makes this trade-off explicit: *"bare 9-digit has too many false positives"*. The same reasoning applies here — this pattern is even looser.

**Options in order of preference:**
1. Use a contextual approach (require the word `iban` or `account number` nearby), same as the passport and DOB patterns.
2. Drop the IBAN pattern entirely until a validation step (check-digit verification) can be added.
3. At minimum, anchor to the known IBAN country-code allowlist: `\b(?:GB|DE|FR|NL|ES|IT|CH|AT|BE|SE|NO|DK|FI|PL|PT|...)\d{2}[A-Z0-9]{4,30}\b` — this still has false positives but limits blast radius to plausible country codes.

---

### [MUST FIX] Return-type annotation regression in `fetch.py` — Score: 85

**File:** `src/pocketpaw/tools/fetch.py`

The diff inserts an early return inside `get_directory_keyboard()` whose declared return type is `-> InlineKeyboardMarkup` (non-optional):

```python
def get_directory_keyboard(path: Path, jail: Path | None = None) -> InlineKeyboardMarkup:
    """Generate inline keyboard for directory contents."""
+   if InlineKeyboardMarkup is None:
+       return None           # ← None is not InlineKeyboardMarkup
```

Returning `None` from a function annotated `-> InlineKeyboardMarkup` is a mypy error. Any caller that trusts the annotation will not null-check the result, potentially causing a downstream `AttributeError` or `TypeError`.

**Fix:** Update the return annotation to `-> InlineKeyboardMarkup | None` to match the new behavior, and verify all call sites null-check the return value.

---

### [SHOULD FIX] Space-separated SSN pattern catches `\t` and `\n` separators — Score: 82

**File:** `src/pocketpaw/security/pii.py`

```python
(r"\b\d{3}\s\d{2}\s\d{4}\b", PIIType.SSN, 0),
```

`\s` matches any whitespace: space, tab, newline, carriage return, form feed. This means `123\t45\t6789` (tab-separated table columns) and `123\n45\n6789` (three values on consecutive lines in log output) both match. Since SSN space-separation is not a formally standardized format to begin with, pulling in tab- and newline-separated runs of digits causes false positives in log data and structured text exports.

**Fix:** Use a literal space `\ ` or `[ ]` instead of `\s`:

```python
(r"\b\d{3} \d{2} \d{4}\b", PIIType.SSN, 0),
```

This also matches how HMRC and some European agencies print SSN-equivalent numbers (space-separated), without catching column-split log data.

---

### Minor observations (no blocking issues)

**Passport pattern correctly uses context keyword** — the `(?:passport)` requirement is well-judged; avoids the IBAN problem for this type. One nit: wrapping a single alternation in a non-capturing group `(?:passport)` is redundant — `passport` alone is sufficient. Harmless, but slightly noisy.

**Trust levels — correct** — `RunPythonTool → critical` is the right call (same risk surface as `ShellTool`); `InstallPackageTool → high` is appropriate given the Guardian + regex guardrails already in place. The audit severity mapping in `registry.py` lines 109–115 confirms both levels now route to the correct `CRITICAL` / `WARNING` severities respectively.

**fetch.py crash fix — correct in intent** — the null guard is the right approach. The only issue is the annotation (above).

**Test coverage updated correctly** — both `test_run_python.py` and `test_install_package.py` assertions updated to match the new values. No new test coverage for the PII additions or the fetch guard; consider adding a test asserting `get_directory_keyboard()` returns `None` (not raises) when called without `python-telegram-bot` installed.

DevRohit06 (changes requested):
IBAN regex false positives, return-type regression in fetch.py, SSN \s too broad. See comment.

prakashUXtech:
Small PR, wide surface — 21 lines across 6 files. Conflicting with dev now. Rebase and I'll take a look. The tool trust-level and PII scanner paths have both been touched recently, so the resolution should be straightforward.

DevRohit06:

PR #706, opened 2026-03-21
Fix/csp xss protection

anish1301:
### What problem does this solve?
The current `Content-Security-Policy` header explicitly allowed `unsafe-inline` in its `script-src` directive (added to support Alpine.js/UnoCSS initialization). By allowing this, the CSP's primary defense-in-depth against XSS execution was nullified. An attacker who managed to inject a malicious `<script>` tag or HTML payload could achieve arbitrary execution because the policy whitelisted all inline scripts globally.

### How does this solve it?
Instead of adding heavy Node.js built-step refactors for UnoCSS (which would disrupt the Python-focused dev workflow for this dashboard), this PR implements a native, zero-dependency **dynamic CSP nonce** solution:
1. The FastAPI middleware now dynamically generates a secure 16-byte cryptographic nonce using Python's built-in `secrets` module for every incoming request.
2. The `Content-Security-Policy` header was updated to replace `'unsafe-inline'` with the strict `'nonce-<dynamic>'` value, maintaining `'unsafe-eval'` for Alpine.js bindings.
3. The nonce is injected securely into the `base.html` Jinja2 template and attached *only* to the authorized inline `<script>` tags used to initialize UnoCSS and Lucide icons.

### Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Security patch

### DIFF
--- a/src/pocketpaw/dashboard.py
+++ b/src/pocketpaw/dashboard.py
@@ -148,6 +148,10 @@ _EXTRA_ORIGINS = list(set(_BUILTIN_ORIGINS + _custom_origins))
 @app.middleware("http")
 async def security_headers_middleware(request: Request, call_next):
     """Add security headers to all responses."""
+    import secrets
+    nonce = secrets.token_urlsafe(16)
+    request.state.csp_nonce = nonce
+
     response = await call_next(request)
 
     # Allow the file-content endpoint to be embedded in same-origin iframes
@@ -161,10 +165,10 @@ async def security_headers_middleware(request: Request, call_next):
     response.headers["X-Content-Type-Options"] = "nosniff"
     response.headers["Referrer-Policy"] = "strict-origin-when-cross-origin"
     response.headers["Permissions-Policy"] = "camera=(), microphone=(), geolocation=()"
-    # CSP: allow self + CDN + inline styles/scripts (required by Alpine.js/UnoCSS)
+    # CSP: allow self + CDN + strictly nonced scripts (Alpine evaluates allowed via unsafe-eval)
     response.headers["Content-Security-Policy"] = (
         "default-src 'self'; "
-        "script-src 'self' 'unsafe-inline' 'unsafe-eval' "
+        f"script-src 'self' 'nonce-{nonce}' 'unsafe-eval' "
         "https://cdn.jsdelivr.net https://unpkg.com; "
         "style-src 'self' 'unsafe-inline' https://cdn.jsdelivr.net https://fonts.googleapis.com; "
         "font-src 'self' https://fonts.gstatic.com https://cdn.jsdelivr.net; "
--- a/src/pocketpaw/frontend/templates/base.html
+++ b/src/pocketpaw/frontend/templates/base.html
@@ -38,7 +38,7 @@
     <link rel="stylesheet" href="/static/css/styles.css?v={{ v }}" />
 
     <!-- UnoCSS Runtime (utilities alongside custom CSS) -->
-    <script>
+    <script nonce="{{ request.state.csp_nonce }}">
       window.__unocss = {
         theme: {
           colors: {
@@ -312,7 +312,7 @@
     <script src="/static/js/app.js?v={{ v }}"></script>
 
     <!-- Initialize Lucide icons + dismiss preloader -->
-    <script>
+    <script nonce="{{ request.state.csp_nonce }}">
       // Initial render
       document.addEventListener('DOMContentLoaded', () => {
         lucide.createIcons();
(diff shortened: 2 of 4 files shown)


DevRohit06 (commented):
## Review: #706 — Fix/csp xss protection

### Summary
Replaces `'unsafe-inline'` with nonce-based CSP for inline scripts in the dashboard. Good security improvement.

### What's Good
- Nonce generated per-request with `secrets.token_urlsafe(16)` (128-bit entropy) is cryptographically sound
- `request.state.csp_nonce` pattern integrates cleanly with FastAPI middleware
- Both inline `<script>` blocks in `base.html` are properly nonced

### Notes

**1. `unsafe-eval` remains**

The CSP still includes `'unsafe-eval'` which is required for Alpine.js runtime evaluation. This means eval-based XSS vectors are still possible. This is a known limitation when using Alpine.js. Documenting this trade-off would be helpful.

**2. Check for other inline scripts**

Only `base.html` is updated. If any other templates have inline `<script>` blocks (without a nonce), they will break silently under the new CSP. Verify no other templates use inline scripts. A quick `grep -r '<script>' --include='*.html'` on the templates directory should confirm.

**3. Overall: good incremental hardening**

Moving from `unsafe-inline` to nonce-based scripts is a meaningful improvement even with `unsafe-eval` present. It blocks injected inline scripts while allowing only the two known inline blocks.

Looks good overall. Approve once you've confirmed no other inline scripts exist in the templates.

anish1301:
Yeah I guess no other  no other inline scripts exist in the templates. LGTM .

DevRohit06 (changes requested):
## Review

Good security hardening — replacing `'unsafe-inline'` with a per-request nonce is the right approach. Nonce generation (`secrets.token_urlsafe(16)`) and middleware ordering are correct.

### Bug: Inline `onclick` handlers will break (regression)

`plan_approval.html` has raw `onclick` attributes:
```html
<button onclick="PlanMode.reject()" ...>
<button onclick="PlanMode.approve()" ...>
```
With `'unsafe-inline'` removed, CSP blocks native inline event handlers. `'unsafe-eval'` does NOT cover `onclick` — it only covers `eval()`, `Function()`, `setTimeout(string)`.

**Fix:** Replace with Alpine.js `@click` or `addEventListener`:
```html
<button @click="PlanMode.reject()" ...>
<button @click="PlanMode.approve()" ...>
```

### Minor
- `import secrets` is inside the middleware body (runs every request). Move to top-level imports — Python caches it, but it's cleaner.
- No tests verifying the nonce appears in CSP header and rendered HTML. A basic integration test would help.

Fix the `onclick` regression and this is good to go.

DevRohit06:
### Code review

Found 1 issue:

1. **No test guards the core security claim of this PR** — the existing `TestSecurityHeaders.test_headers_present` only asserts `"default-src 'self'"` is in the CSP, but never checks that `'unsafe-inline'` is absent from `script-src` and never verifies a nonce token is present. This means a future change could silently reintroduce `'unsafe-inline'` with zero test failures, completely undoing the protection added here. The rest of the implementation is correct: nonce generation, `request.state` lifecycle, `@click` replacing `onclick` (which Alpine correctly compiles to `addEventListener`, bypassing CSP inline-handler blocking), and the absence of other non-nonced inline scripts across all templates.

https://github.com/pocketpaw/pocketpaw/blob/36b7a717f7a7d02c63fd11e25709c528993f6446/tests/test_dashboard_security.py#L251-L258

Suggested addition to `TestSecurityHeaders`:

```python
def test_csp_no_unsafe_inline_in_script_src(self, test_client):
    resp = test_client.get("/")
    csp = resp.headers.get("Content-Security-Policy", "")
    # Extract script-src directive
    directives = {d.strip().split()[0]: d.strip() for d in csp.split(";")}
    script_src = directives.get("script-src", "")
    assert "'unsafe-inline'" not in script_src, "script-src must not contain 'unsafe-inline'"
    assert "nonce-" in script_src, "script-src must contain a per-request nonce"
```

DevRohit06 (changes requested):
Missing test asserting unsafe-inline is removed from CSP. See comment for details.

prakashUXtech (commented):
The CSP hardening looks good. Nonce generation with `secrets.token_urlsafe(16)` is correct, `request.state.csp_nonce` integrates cleanly with Starlette's middleware pattern, and replacing `onclick` with `@click` in `plan_approval.html` was the right call — inline event handlers are blocked by CSP without `unsafe-inline`.

One thing needed before this can merge: **rebase onto current `dev`**. The CI failures are all caused by `ImportError: cannot import name '_COGNITIVE_SESSION_KEY' from 'pocketpaw.soul.cognitive'` in `test_soul_cognitive_engine.py` — nothing related to your changes, just your branch being behind. A rebase will pull in the current state and the tests should pass.

Once CI is green this is ready.

One minor cleanup: `import secrets` inside the middleware function body runs on every request. Moving it to the top-level imports is cleaner (Python caches the import, so it's not a performance issue, just style).

DevRohit06:

PR #707, opened 2026-03-21
feat(memory): add vector indexing and semantic retrieval support for …

### LINKED ISSUE #455: feat: agent memory sharing — cross-session knowledge graph with semantic search
## Summary

Upgrade PocketPaw's memory system from flat file-based fact storage to a semantic knowledge graph with vector-based retrieval, enabling the agent to build and query a rich, interconnected understanding of the user's world over time.

## Motivation

Current memory (`memory/file_store.py`) stores facts as plain text lines. This works for simple recall but fails at:
- **Semantic search** — "What did I say about the API refactor?" requires exact keyword match
- **Relationship tracking** — No way to connect related facts (e.g., "Project X uses React" + "React 19 has breaking changes" → proactive warning)
- **Temporal awareness** — No timestamps or decay on memories
- **Cross-session continuity** — Each session starts with limited context window

### Phase 1 — Vector-backed memory retrieval
- Add optional vector store backend (ChromaDB, Qdrant, or SQLite-vec)
- Embed facts at write time using a lightweight embedding model (e.g., `nomic-embed-text` via Ollama)
- RAG-style retrieval: before each agent turn, query top-K relevant memories
- Fall back to current file store if vector backend isn't configured

### Phase 2 — Knowledge graph
- Extract entities and relationships from conversations (using LLM extraction)
- Store as lightweight graph (NetworkX or SQLite with adjacency tables)
- Enable queries like "What projects use Python?" or "What's related to the deployment issue?"

### Phase 3 — Memory management UI
- Dashboard page to browse, search, edit, and delete stored memories
- Visualization of the knowledge graph
- Memory usage stats and pruning controls

## Configuration

```env
POCKETPAW_MEMORY_BACKEND=vector     # file | vector | mem0
POCKETPAW_EMBEDDING_MODEL=nomic-embed-text
POCKETPAW_VECTOR_STORE=chromadb     # chromadb | qdrant | sqlite-vec
```

## Benefits

- Agent becomes dramatically more useful over time
- "It remembers everything" — key differentiator for a personal AI agent
- Fully local — no data leaves the user's machine
- Complements existing mem0 integration

AMRITESH240304:
## What does this PR do?

Phase 1: Semantic Vector Search
Embeddings via Ollama or local hash, sqlite-vec storage
Semantic search with auto-cleanup on delete

Phase 2: Knowledge Graph
SQLite schema (entities, relationships, evidence)
Heuristic NLP extraction from memory text
Graph snapshot, update, prune operations

Phase 3: Memory Management UI
Facts, Graph, Stats tabs in memory modal
Inline editing, tagging, pruning

## Related Issue

Fixes #455 

## Changes Made

| File | Changes |
|------|---------|
| `file_store.py` | Added vector-backed semantic search (embedding, index upsert/delete, retrieval, and cleanup on memory/session deletes). |
| `src/pocketpaw/memory/`<br>`manager.py` | Wired new file-vector settings into manager/store creation and mapped legacy vector backend to file+vector mode. |
| `src/pocketpaw/`<br>`config.py` | Added new settings fields for file-vector controls (enable flag, store, provider, model, base URL) and updated backend description. |
| `src/pocketpaw/`<br>`dashboard_state.py` | Added memory config key mappings for the new file-vector settings. |
| `src/pocketpaw/`<br>`dashboard_ws.py` | Added save/load handling for file-vector settings in websocket settings payloads. |
| `src/pocketpaw/frontend/js/`<br>`app.js` | Added frontend defaults and sync fields for file-vector settings. |
| `src/pocketpaw/frontend/js/`<br>`websocket.js` | Added file-vector fields to settings save payload sent to backend. |
| `src/pocketpaw/frontend/templates/components/modals/`<br>`settings.html` | Updated memory label and added File+Vector controls (toggle, store, provider, model, base URL). |
| `src/pocketpaw/frontend/js/features/`<br>`transparency.js` | Render Memory Graph Data. |
| `src/pocketpaw/frontend/templates/components/modals/`<br>`memory.html` | Tab view for UI. |
| `tests/`<br>`test_file_memory_fixes.py` | Added tests for semantic retrieval relevance, vector record deletion sync, and manager semantic context path. |

## How to Test

1. Open settings and click on memory
2. check mark option vector retrieval 
3. setup these details  
image

4. open New Chat save some data or talk generally.
5. open New Chat again and ask about last chat session.

## Evidence of Testing

image
image
image

## Result
image
image

## Checklist

### DIFF
--- a/src/pocketpaw/memory/file_store.py
+++ b/src/pocketpaw/memory/file_store.py
@@ -496,7 +496,8 @@ class FileMemoryStore:
         self._inv_dirty = True
 
         self._initialize_vector_backend()
-        self._initialize_graph_store()
+        if self._graph_enabled:
+            self._initialize_graph_store()
 
         self._load_index()
 
@@ -1284,8 +1285,12 @@ class FileMemoryStore:
 
     async def _semantic_search_sqlite(self, query: str, user_id: str, limit: int) -> list[dict]:
         """Semantic search using local sqlite vector table."""
+        if limit <= 0:
+            return []
+
         query_embedding = await self._embed_text(query)
-        candidate_limit = max(limit * 50, 500)
+        # Keep candidate fetch bounded to avoid loading excessive embedding JSON blobs.
+        candidate_limit = min(max(limit * 10, 200), 2000)
 
         def _search_sync() -> list[dict]:
             with sqlite3.connect(self._vector_db_path) as conn:
@@ -1888,6 +1893,9 @@ class FileMemoryStore:
     ) -> dict:
         """Return a lightweight graph snapshot for dashboard visualization."""
 
+        if not self._graph_enabled or not self._graph_db_path.exists():
+            return {"nodes": [], "edges": []}
+
         def _snapshot_sync() -> dict:
             with sqlite3.connect(self._graph_db_path) as conn:
                 if query and query.strip():
@@ -1916,27 +1924,28 @@ class FileMemoryStore:
 
                 node_ids = [str(row[0]) for row in nodes]
                 if node_ids:
-                    placeholders = ",".join("?" for _ in node_ids)
+                    conn.execute("CREATE TEMP TABLE selected_node_ids (entity_id TEXT PRIMARY KEY)")
+                    self._insert_valid_ids_batched(conn, "selected_node_ids", node_ids)
                     edge_rows = conn.execute(
-                        f"""
-                           SELECT r.relationship_id,
-                               r.source_entity_id,
-                               r.target_entity_id,
-                               se.display_name,
-                               te.display_name,
-                               r.relation_type,
-                               r.weight,
-                               r.last_seen
-                           FROM relationships r
-                           JOIN entities se ON se.entity_id = r.source_entity_id
-                           JOIN entities te ON te.entity_id = r.target_entity_id
-                           WHERE r.user_scope = ?
-                             AND r.source_entity_id IN ({placeholders})
-                             AND r.target_entity_id IN ({placeholders})
+                        """
+                        SELECT r.relationship_id,
+                            r.source_entity_id,
+                            r.target_entity_id,
+                            se.display_name,
+                            te.display_name,
+                            r.relation_type,
+                            r.weight,
+                            r.last_seen
+                        FROM relationships r
+                        JOIN entities se ON se.entity_id = r.source_entity_id
+                        JOIN entities te ON te.entity_id = r.target_entity_id
+                        JOIN selected_node_ids src_nodes ON src_nodes.entity_id = r.source_entity_id
+                        JOIN selected_node_ids tgt_nodes ON tgt_nodes.entity_id = r.target_entity_id
+                        WHERE r.user_scope = ?
                         ORDER BY r.weight DESC, r.last_seen DESC
                         LIMIT ?
                         """,
-                        (user_id, *node_ids, *node_ids, limit),
+                        (user_id, limit),
                     ).fetchall()
                 else:
                     edge_rows = []
@@ -1978,6 +1987,13 @@ class FileMemoryStore:
         height: int = 400,
     ) -> str:
         """Generate SVG visualization of the memory knowledge graph using networkx."""
+
+        if not self._graph_enabled or not self._graph_db_path.exists():
+            return (
+                f'<svg xmlns="http://www.w3.org/2000/svg" width="{width}" height="{height}">'
+                '<text x="10" y="20" fill="rgba(255,255,255,0.6)">No graph data</text>'
+                "</svg>"
+            )
 
         def _generate_svg_sync() -> str:
             import math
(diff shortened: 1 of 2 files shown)


AMRITESH240304:
Hey @DevRohit06, all phases have been implemented and everything is now ready for review.

AMRITESH240304:
Previously regex matches were stored almost directly which created noisy edges from conversational text.
Now extracted entity names are canonicalized (alias/format normalization) so duplicate variants map to one node.
Previously relation labels were free-form now they are normalized into a controlled schema for consistency.
Previously all valid-looking matches could be stored now only edges with confidence >= 0.75 are persisted.


and other changes also have been addressed and new test case for the regex have also been added.

AMRITESH240304:
- Fixed SQLite variable-limit crash by switching NOT IN (...) to a temp-table approach and batching inserts in FileMemoryStore._cleanup_orphan_records.
- Added _insert_valid_ids_batched() to safely populate temp tables in batches (default 500 IDs).
- Escaped SVG text with html.escape() for both edge labels and node labels in get_graph_svg to prevent malformed SVG.
- Aligned FileMemoryStore.__init__ default vector_enabled to False to match Settings.file_vector_enabled and avoid accidental vector DB creation.
- Added tests covering large-store cleanup (1050+ entries) and SVG escaping for <, >, and &, ensuring regressions are caught.

DevRohit06:
### Code review

Found 3 issues.

1. **`_initialize_graph_store()` runs unconditionally, creating database files when the feature is disabled** — When `vector_enabled=False` (the default), `_graph_enabled` is correctly set to `False`, but `_initialize_graph_store()` is still called unconditionally in `__init__`, creating `knowledge_graph.sqlite3` with six tables on every startup. Every user who hasn't opted into vector/graph gets unnecessary disk writes and database files. Guard the call with `if self._graph_enabled:`.
https://github.com/pocketpaw/pocketpaw/blob/8d0f3400a3b63c28a92570f7e49add12e61ffede/src/pocketpaw/memory/file_store.py#L1201-L1202

2. **Unbounded SQL `IN` clause in `get_graph_snapshot` / `get_graph_svg` exceeds SQLite variable limit when `limit=500`** — Both `_snapshot_sync` and `_generate_svg_sync` build `source_entity_id IN ({placeholders}) AND target_entity_id IN ({placeholders})` then pass `(user_id, *node_ids, *node_ids, limit)` as bind parameters. With `limit=500` (permitted by the `/api/memory/graph` endpoint's `max(1, min(limit, 500))`), this passes up to 1001 bind variables in one query and will crash with `OperationalError: too many SQL variables` (SQLite's default limit is 999). The same temporary-table fix already applied to `_cleanup_orphan_records` should be applied here — or cap the in-clause via a subquery/CTE, or lower the effective `limit` ceiling for these queries to ≤ 499.
https://github.com/pocketpaw/pocketpaw/blob/8d0f3400a3b63c28a92570f7e49add12e61ffede/src/pocketpaw/memory/file_store.py#L2161-L2183

3. **`_semantic_search_sqlite` loads an unbounded number of embeddings into Python for cosine ranking** — `candidate_limit = max(limit * 50, 500)`. At the default `limit=5` this is 250 rows; at `limit=100` (callers may pass higher values) it fetches 5,000 rows, each with a full embedding JSON blob (384 floats ≈ 3 KB serialized), pulling ~15 MB into Python before sorting. With a large memory store this call will be slow and allocate disproportionate memory. Cap `candidate_limit` at a sensible fixed ceiling (e.g. `min(max(limit * 10, 200), 2000)`) to bound memory allocation.
https://github.com/pocketpaw/pocketpaw/blob/8d0f3400a3b63c28a92570f7e49add12e61ffede/src/pocketpaw/memory/file_store.py#L1988-L1991

---

The three issues from the previous review cycle (SQLite variable limit in `_cleanup_orphan_records`, SVG label HTML-escaping, and `FileMemoryStore` constructor default) are all confirmed fixed.

AMRITESH240304:
## Changes Applied

- Guarded graph schema initialization so it only runs when graph is enabled.
- Added safe early returns for graph APIs when graph is disabled or the graph DB does not exist.
- Replaced large duplicated IN placeholder queries with temp-table joins in both snapshot and SVG graph edge fetches to avoid SQLite variable-limit crashes at limit=500.
- Bounded semantic search candidate fetch size to prevent large in-memory embedding loads.  
  - New bound: `min(max(limit * 10, 200), 2000)`
- Updated graph stats counting to return zero when graph is disabled.

---

## Tests Added / Updated

- Graph DB not created when feature disabled.
- Snapshot empty without DB creation when disabled.
- Snapshot limit=500 avoids SQLite variable-limit issues.
- Semantic candidate limit capping behavior.
- Existing empty-graph SVG test tightened to ensure no DB file creation.
- Networkx-unavailable fallback test aligned to graph-enabled mode.

DevRohit06 (changes requested):
Graph store init unconditional, SQL var limit exceeded, unbounded memory load. See comment.

DevRohit06 (approved):
All three issues from the previous review are fixed. Graph store guard, temp-table for IN clause, and candidate_limit cap all look good.

PR #713, opened 2026-03-22
[FI] Fix: Prevent E2E test failures by skipping without full environment setup

### LINKED ISSUE #715: Tests fail when optional dependencies are not installed (playwright, psutil, sarvamai)
### Operating System

windows 11

### Python Version

3.13.7

### PocketPaw Version

dev (local repository)

### Installation Method

Other

## Problem

Several test suites fail in environments where optional dependencies are not installed (e.g., playwright, psutil, sarvamai).

Instead of skipping gracefully, these tests raise ImportError or ModuleNotFoundError, causing the entire test run to fail.

## Steps to reproduce

1. Clone the repository
2. Do NOT install optional dependencies
3. Run: python -m pytest
4. Observe failures in browser, sarvam, and system-related tests

## Expected behavior

Tests that rely on optional dependencies should be skipped when those modules are not available.

## Actual behavior

Tests fail with ImportError or ModuleNotFoundError.

## Suggested fix

Use pytest.importorskip in test files to conditionally skip tests when optional modules are missing.

### Steps to Reproduce

1. Clone the repository
2. Do NOT install optional dependencies (playwright, psutil, sarvamai)
3. Run: python -m pytest
4. Observe failures in multiple test suites due to missing dependencies

### Expected Behavior

Tests that depend on optional dependencies should be skipped gracefully when those modules are not installed, instead of failing with ImportError or ModuleNotFoundError.

### Actual Behavior

Tests fail with ImportError or ModuleNotFoundError when optional dependencies are missing, causing the entire test run to fail.

ImportError: 'playwright' is required but not installed
ModuleNotFoundError: No module named 'sarvamai'
psutil not installed

39 failed, 3204 passed, 56 skipped

### Relevant Files

tests/test_browser/test_driver.py
tests/test_sarvam.py
tests/test_daemon.py
tests/test_tools.py

### Before submitting

rishiraj65:
E2E tests currently fail in environments where the full setup (browser, server, or runtime dependencies) is not available.

This PR ensures that E2E tests are skipped by default unless explicitly enabled via the RUN_E2E environment variable.

Fixes #715

Key changes:
- Added module-level skip guard in all E2E test files
- Ensures tests do not fail during collection or execution in minimal environments
- Maintains ability to run E2E tests when full setup is available

This improves test stability and prevents false negatives during development and CI.

### DIFF
--- a/tests/e2e/test_dashboard.py
+++ b/tests/e2e/test_dashboard.py
@@ -7,7 +7,17 @@
 # Run with: pytest tests/e2e/ -v --headed (to see browser)
 # Run headless: pytest tests/e2e/ -v
 
+import pytest
+import os
+
+# Skip entire module BEFORE anything else
+if not os.environ.get("RUN_E2E"):
+    pytest.skip("E2E tests require full environment setup", allow_module_level=True)
+
+pytest.importorskip("playwright.sync_api")
+
 from playwright.sync_api import Page, expect
+
 
 
 class TestDashboardLoads:
--- a/tests/e2e/test_deep_work_e2e.py
+++ b/tests/e2e/test_deep_work_e2e.py
@@ -15,12 +15,17 @@
 #   6. Output chaining: task output stored on task.output field
 #   7. Cancel rejection: completed/cancelled projects can't be cancelled again
 #   8. PawKit YAML round-trip through real file system
+import pytest
+import os
+
+# Skip entire module unless explicitly enabled
+if not os.environ.get("RUN_E2E"):
+    pytest.skip("E2E tests require full environment setup", allow_module_level=True)
 
 import asyncio
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import httpx
-import pytest
 from fastapi import FastAPI
 
 from pocketpaw.deep_work.api import router as deep_work_router
--- a/tests/e2e/test_security.py
+++ b/tests/e2e/test_security.py
@@ -10,8 +10,16 @@
 # Run with: pytest tests/e2e/test_security.py -v --headed
 
 
-from playwright.sync_api import Page, expect
+import pytest
+import os
 
+# Skip entire module BEFORE anything else
+if not os.environ.get("RUN_E2E"):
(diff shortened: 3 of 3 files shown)


DevRohit06 (changes requested):
Thanks for the contribution! However, this PR solves a problem that's already handled by the project's conventions.

## `RUN_E2E` env var is redundant

E2E tests are already excluded by the standard test command: `uv run pytest --ignore=tests/e2e` (documented in CLAUDE.md and the README). Adding a second gating mechanism creates confusion about which one is authoritative.

## If a skip guard is desired, use conftest.py

Inline module-level skip guards in every E2E file is repetitive. The pytest-native approach would be a single `tests/e2e/conftest.py` with a session-scoped fixture or autouse marker, e.g.:

```python
import pytest
import os

if not os.environ.get("RUN_E2E"):
    pytest.skip("E2E tests require RUN_E2E=1", allow_module_level=True)
```

One file instead of patching every test module.

## Other notes

- This overlaps with your other PR #714 which includes these same changes plus more. Should be consolidated into one PR.
- PR targets `main` instead of `dev` (contribution guidelines require `dev`).
- Minor style: extra blank line in `test_dashboard.py`, missing blank line before `class TestWebSocketAuth` in `test_security.py`.

prakashUXtech (changes requested):
E2E tests are already excluded from the standard test run via `uv run pytest --ignore=tests/e2e`, so the `RUN_E2E` guard doesn't solve a real problem in practice. Adding per-file skip guards in every E2E module creates two competing mechanisms and confusion about which one is authoritative.

If a single-file skip guard is useful for running E2E tests in isolation (e.g., `pytest tests/e2e/test_dashboard.py` without `--ignore`), the right place for it is a single `tests/e2e/conftest.py` with a module-level skip, not inline guards copied into every file. One file instead of three.

Also: the extra blank line in `test_dashboard.py` and the missing blank line before `class TestWebSocketAuth` in `test_security.py` should be cleaned up — ruff will catch those.

If you want to proceed with the conftest approach, happy to approve that revision.

DevRohit06:

PR #714, opened 2026-03-22
[FI] Fix: Skip tests requiring optional dependencies (playwright, psutil, sarvamai)

rishiraj65:
Several test suites fail in minimal environments due to missing optional dependencies such as playwright, psutil, and sarvamai.

Instead of failing, these tests should be skipped when the required modules are not available.

This impacts developer experience and CI reliability, as contributors without these optional dependencies cannot run the full test suite successfully.

Proposed solution:
Use pytest.importorskip to conditionally skip tests that depend on optional modules.

### DIFF
--- a/tests/e2e/test_dashboard.py
+++ b/tests/e2e/test_dashboard.py
@@ -7,7 +7,17 @@
 # Run with: pytest tests/e2e/ -v --headed (to see browser)
 # Run headless: pytest tests/e2e/ -v
 
+import pytest
+import os
+
+# Skip entire module BEFORE anything else
+if not os.environ.get("RUN_E2E"):
+    pytest.skip("E2E tests require full environment setup", allow_module_level=True)
+
+pytest.importorskip("playwright.sync_api")
+
 from playwright.sync_api import Page, expect
+
 
 
 class TestDashboardLoads:
--- a/tests/e2e/test_deep_work_e2e.py
+++ b/tests/e2e/test_deep_work_e2e.py
@@ -15,12 +15,17 @@
 #   6. Output chaining: task output stored on task.output field
 #   7. Cancel rejection: completed/cancelled projects can't be cancelled again
 #   8. PawKit YAML round-trip through real file system
+import pytest
+import os
+
+# Skip entire module unless explicitly enabled
+if not os.environ.get("RUN_E2E"):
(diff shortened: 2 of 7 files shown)


DevRohit06 (changes requested):
Thanks for the contribution! Skipping tests for missing optional deps is the right idea, but some of the skips here are too broad or unnecessary.

## What works well

- `pytest.importorskip("playwright")` in `test_browser/test_driver.py` - correct, playwright is optional
- `pytest.importorskip("sarvamai")` in `test_sarvam.py` - correct, sarvamai is optional

## Issues

### `psutil` skip in `test_tools.py` is wrong

`test_tools.py` doesn't import or use `psutil`. It tests screenshot, filesystem, and config tools. The `importorskip("psutil")` at the top would skip the entire file (all those unrelated tests) when `psutil` is missing, even though none of them need it.

### `psutil` skip in `test_daemon.py` is too broad

The daemon module only uses `psutil` in `context.py` (lazy import for system metrics). Most daemon tests cover cron parsing, trigger engine, and intention store, which have no `psutil` dependency. Skipping the entire module throws away useful test coverage. If needed, only the specific `ContextHub` system-metrics tests should be guarded.

### `psutil` is effectively a core dependency

`psutil` appears in multiple combined extras in `pyproject.toml` and would be installed in any standard dev setup (`uv sync --dev`). This guard is solving a problem that rarely exists in practice.

### `RUN_E2E` env var is redundant

E2E tests are already excluded by convention via `pytest --ignore=tests/e2e` (documented in CLAUDE.md and the project README). Adding a second gating mechanism introduces confusion about which one is authoritative. If a new env var gate is desired, that should be a separate discussion/PR.

## Suggestion

Keep the playwright and sarvamai skips, drop the `psutil` skips and the `RUN_E2E` env var guards. Also, PR should target `dev` instead of `main`.

prakashUXtech:
Closing: CI already skips this directory.

The test workflow runs with `--ignore=tests/e2e`, so the imports this PR guards never execute in CI. Adding a `RUN_E2E` env gate on top would mean the e2e suite needs two things set before it runs locally, which makes it more likely to be silently skipped by someone who meant to run it.

The optional-dependency problem you hit is real if you run `pytest` with no arguments. If that's what bit you, an issue describing that path would be useful. Thanks for digging into it.

PR #722, opened 2026-03-22
[FI] Fix: Make Telegram token validation consistent across codebase

BHANU063:
## Issue
Telegram bot token validation was inconsistent between config.py (using regex) and dashboard_channels.py (using simple colon check), causing users to get different validation results depending on where they enter tokens.

## Fix
- Replace simple colon check in dashboard_channels.py with proper regex pattern
- Use same validation pattern as config.py: `^\d+:[A-Za-z0-9_-]+$`
- Update error message to be more specific about expected format
- Ensures consistent validation behavior across all token entry points

## Testing
- Tested with invalid tokens (no colon, wrong characters, non-numeric ID) - correctly rejected
- Tested with valid format tokens (numeric_id:alphanumeric_secret) - correctly accepted
- Verified validation now matches behavior in config.py

## Result
Users now receive consistent token validation regardless of where they enter Telegram bot tokens, improving user experience and reducing confusion.

### DIFF
--- a/src/pocketpaw/dashboard_channels.py
+++ b/src/pocketpaw/dashboard_channels.py
@@ -582,8 +582,11 @@ async def _validate_channel_tokens(channel: str, config: dict) -> str | None:
 
     elif channel == "telegram":
         bot_token = config.get("bot_token", "")
-        if bot_token and ":" not in bot_token:
-            return "Invalid Telegram bot token. It should be in the format '123456:ABC-DEF...'."
+        if bot_token:
+            # Telegram tokens: numeric_id:alphanumeric_secret format
+            telegram_token_pattern = re.compile(r'^\d+:[A-Za-z0-9_-]+$')
+            if not telegram_token_pattern.match(bot_token):
+                return "Invalid Telegram bot token. Expected format: numeric_id:alphanumeric_secret"
 
     return None
 


DevRohit06 (changes requested):
## Review: #722 — [FI] Fix: Make Telegram token validation consistent across codebase

### Issues

**1. Discord regex may reject valid tokens**

```python
discord_token_pattern = re.compile(
    r'^[A-Za-z0-9_-]{24,}\.[A-Za-z0-9_-]{6}\.[A-Za-z0-9_-]{25,110}$'
)
```

The timestamp segment is locked to exactly 6 characters, but Discord has changed their token format before. The HMAC upper bound of 110 is also arbitrary. If Discord issues a token with a 7-char timestamp or longer HMAC, this regex will reject a valid bot token, breaking the user's setup.

The old check (`len(bot_token) < 50`) was intentionally loose to avoid false rejections. A safer middle ground would be to check for the 3-segment dot-separated structure without strict length constraints:

```python
re.compile(r'^[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+$')
```

**2. Telegram regex is a good improvement**

`^\d+:[A-Za-z0-9_-]+$` properly validates the `numeric_id:secret` format. This is a clear improvement over just checking for `:`.

**3. No tests**

Token validation is a great candidate for unit tests since there are clear valid/invalid cases. Please add tests covering:
- Valid Discord token (3 dot-separated base64 segments)
- Too-short Discord token
- Valid Telegram token (`123456:ABC-DEF...`)
- Telegram token missing colon
- Empty strings (should pass through as no validation needed)

### Action Items
- Loosen the Discord regex to avoid rejecting valid tokens
- Add unit tests

DevRohit06 (changes requested):
## Review

The Telegram regex improvement is good, but there are issues to address:

### 1. Discord regex is too strict (High)
The hardcoded segment lengths will reject valid tokens:
```python
r'^[A-Za-z0-9_-]{24,}\.[A-Za-z0-9_-]{6}\.[A-Za-z0-9_-]{25,110}$'
```
Discord has changed token formats before. Use a relaxed 3-segment pattern instead:
```python
r'^[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+$'
```

### 2. Reuse existing constant (High)
`config.py` already defines `_TELEGRAM_BOT_TOKEN_RE`. Import and reuse it instead of duplicating the regex — avoids future drift.

### 3. Move regex to module level (Medium)
`re.compile()` inside the function means recompilation every call. Use module-level constants like `config.py` does.

### 4. No tests (Medium)
Token validation is easily unit-testable. Add cases for valid/invalid Telegram and Discord tokens.

### 5. Line length (Low)
The Discord regex line exceeds the project's 100-char limit.

### 6. Scope creep (Low)
PR title focuses on Telegram consistency but also changes Discord validation to a strict regex — this introduces regression risk beyond the stated goal.

DevRohit06:

PR #719, opened 2026-03-22
[FI] fix: add missing Windows venv activation command to README

Kavya-M-Injineri:
## What does this PR do?

Adds the missing Windows venv activation command in the
macOS/Linux install section of the README. Windows users
reading that section had no equivalent command shown.

## Related Issue

 Fixes #686

## Changes Made

README.md`: added `# Windows: pocketpaw-env\Scripts\activate`
  as a comment after `source pocketpaw-env/bin/activate`

## How to Test

1. Open README.md in the repository
2. Search for source pocketpaw-env/bin/activate
3. Verify the Windows command appears below it:
     Windows: pocketpaw-env\Scripts\activate

## Evidence of Testing

This is a documentation-only change (README.md).
No code was modified, so no tests need to be run.

Verified by:
- Viewing README.md in the GitHub editor
- Confirming `source pocketpaw-env/bin/activate` 
  is present with Windows equivalent below it
- No existing tests are affected by this change``

## Checklist

- [ ] PR targets `dev` branch (not `main`)
- [ x ] Linked to an existing issue
- [ ] I have run PocketPaw locally and tested my changes
- [ ] Tests pass (`uv run pytest --ignore=tests/e2e`)
- [ ] Linting passes (`uv run ruff check .`)
- [ ] I have added/updated tests if applicable
- [ x ] No unrelated changes bundled in this PR
- [ x ] No secrets or credentials in the diff

PR #735, opened 2026-03-24
[FI] Added missing Windows venv activation command in README

gaykepramod413-prog:
This PR adds the missing Windows venv activation command to improve setup instructions for Windows users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,8 @@ powershell -NoExit -Command "iwr -useb https://pocketpaw.xyz/install.ps1 | iex"
    python -m venv pocketpaw-env
    .\pocketpaw-env\Scripts\Activate.ps1
    ```
+   # Alternative activation command (Command Prompt)
+venv\Scripts\activate
 
 4. Install PocketPaw:
 


gaykepramod413-prog:
.

gaykepramod413-prog (commented):
This PR adds the missing Windows venv activation command to improve setup instructions for Windows users.

DevRohit06 (changes requested):
Thanks for the contribution. Documenting the Command Prompt activation path is useful for Windows users.

However, the added lines are not inside a fenced code block and lack proper indentation, so they break the rendered Markdown: the comment appears as plain text after the closing backticks, and `venv\Scripts\activate` renders as an unformatted loose line.

Please wrap both lines in their own fenced code block (e.g., \`\`\`cmd) with consistent indentation to match the surrounding numbered list, or fold the alternative command into the existing block with a comment.

prakashUXtech:
Closing this in favor of #719, which covers the same fix and targets the `dev` branch (PRs should go to `dev`, not `main`).

#719 also needs one small correction before merging — the cmd.exe activation path should be the PowerShell form (`.\pocketpaw-env\Scripts\Activate.ps1`). Once that's addressed it will land.

Thanks for the contribution.

prakashUXtech:
Closing this one: the added lines land outside the code fence, so the rendered README shows a stray comment and a bare `venv\Scripts\activate` next to the block instead of inside it.

The gap you're pointing at is real, though. The Windows section only gives the PowerShell activation and not the `cmd.exe` one. If you'd like to send it again with both commands inside the fenced block, that's a change worth having. #1128 is also open against the same section, so it's worth a look first to avoid overlap.

PR #737, opened 2026-03-25
feat: Session Actions Menu (three-dot menu)

### LINKED ISSUE #734: Feat: Session Actions Menu (Three-Dot Menu)
### Problem Statement

Currently, session management in the sidebar has inconsistent UX:
- **Rename**: Only accessible via double-click (hidden affordance)
- **Export**: Only available globally from header dropdown
- **Delete**: Only accessible via hover "X" icon (not discoverable on mobile)

### Proposed Solution

Add a **three-dot actions menu** (⋮) to each session row in the sidebar that reveals:
1. **Export** → Expandable submenu with JSON/Markdown options
2. **Rename** → Opens inline title editing
3. **Delete** → With confirmation dialog

- `src/pocketpaw/frontend/js/features/sessions.js`
- `src/pocketpaw/frontend/templates/components/sidebar.html`

### Additional Context

Image

### Before submitting

ragini-pandey:
## What does this PR do?

Adds a **three-dot actions menu (⋮)** to each session row in the sidebar, consolidating all per-session actions into a single discoverable entry point. This replaces the hidden hover-X delete button and surfaces Export and Rename alongside Delete.

## Related Issue

Fixes #734

## Changes Made

- `src/pocketpaw/frontend/templates/components/sidebar.html`: Replace the per-session `<button>` with a `<div>` wrapper containing Alpine.js-local `sessionMenuOpen` state. Add a `more-vertical` trigger button (visible on hover). Add a dropdown with: **Export** (expandable submenu → JSON / Markdown), **Rename** (opens existing inline rename), **Delete** (opens a new confirmation dialog). The confirmation dialog is appended near `</aside>` and driven by a new `deletingSessionId` piece of state.
- `src/pocketpaw/frontend/js/features/sessions.js`: Add `deletingSessionId: null` to `getState()`. Add `confirmDeleteSession(id, event)` method that sets `deletingSessionId` (triggering the confirmation dialog). Add `exportSessionById(id, format)` method that downloads any session by id (not just the active one), enabling Export from the sidebar menu.

## How to Test

1. `uv run pocketpaw` —
 open `http://localhost:8888`
2. Create two or more chat sessions.
3. Hover over a session row — the ⋮ button should appear on the right.
4. Click ⋮ → confirm the dropdown shows **Export**, **Rename**, **Delete**.
5. **Export**: hover Export → submenu shows JSON / Markdown. Click either → file downloads.
6. **Rename**: click Rename → inline input appears pre-filled with the current title. Edit and press Enter (or Escape to cancel).
7. **Delete**: click Delete → confirmation dialog appears. Click **Cancel** (nothing happens) then Delete again → click **Delete** in the dialog — session is removed.
8. Double-click rename still works as before.

## Evidence of Testing

```
uv run pytest --ignore=tests/e2e -q

3677 passed, 46 skipped, 89 warnings in 30.97s

uv run ruff check .
All checks passed.

<img width="312" height="218" alt="Screenshot 2026-03-25 at 4 02 03 PM" src="https://github.com/user-attachments/assets/73178f33-a936-4519-b794-58ce1b9a46ba" />

```

## Checklist

### DIFF
--- a/src/pocketpaw/frontend/templates/components/sidebar.html
+++ b/src/pocketpaw/frontend/templates/components/sidebar.html
@@ -151,6 +151,7 @@
 
                 <!-- Three-dot actions menu trigger — absolutely positioned so it is NOT nested inside the session button -->
                 <button
+                    x-ref="menuTrigger"
                     class="absolute right-1.5 top-1/2 -translate-y-1/2 p-1 rounded opacity-0 group-hover:opacity-100 hover:bg-white/10 transition-all"
                     :class="sessionMenuOpen ? '!opacity-100 bg-white/10' : ''"
                     @click.stop="sessionMenuOpen = !sessionMenuOpen"
@@ -159,7 +160,7 @@
                   <i data-lucide="more-vertical" class="w-3.5 h-3.5 text-white/50"></i>
                 </button>
 
-                <!-- Actions dropdown -->
+                <!-- Actions dropdown — fixed positioning to escape scroll container clipping -->
                 <div
                     x-show="sessionMenuOpen"
                     x-transition:enter="transition ease-out duration-100"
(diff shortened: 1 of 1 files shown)


DevRohit06:
### Code review

Found 3 issues.

1. **Dropdown clipped by the `overflow-y: auto` scroll container** (definite bug — reproducible on any list with 6+ sessions)

   The three-dot dropdown is absolutely positioned inside the session list's `<div class="flex-1 overflow-y-auto ...">`. Per the CSS spec, `overflow-y: auto` creates a scroll container that clips all absolutely positioned descendants that extend beyond its bounds — `z-index` does not help here. For sessions near the bottom of the visible list, the entire dropdown or its lower items will be hidden behind the container edge.

   Fix: move the dropdown rendering outside the scroll container (e.g. render it into a portal/teleport at the `<aside>` or `<body>` level, or use Alpine's `x-teleport` if available).

   https://github.com/pocketpaw/pocketpaw/blob/155cde2ca43aa9a9650987ffa2a7fa454d4fecf9/src/pocketpaw/frontend/templates/components/sidebar.html#L162-L230

2. **Export submenu flickers/closes when mousing from the trigger to the sub-panel** (real UX defect)

   The Export button and sub-panel each have independent `@mouseenter`/`@mouseleave` handlers, with `ml-1` (4px gap) between them. When the cursor crosses that gap, `mouseleave` fires on the button (`exportSubOpen = false`, submenu disappears) before `mouseenter` fires on the sub-panel (`exportSubOpen = true`, submenu reappears). The submenu momentarily vanishes and can feel broken, especially for users who move their mouse diagonally.

   Fix: move the `@mouseenter`/`@mouseleave` handlers up to the shared parent `<div x-data="{ exportSubOpen: false }" class="relative">` so the gap between trigger and panel doesn't break hover state.

   https://github.com/pocketpaw/pocketpaw/blob/155cde2ca43aa9a9650987ffa2a7fa454d4fecf9/src/pocketpaw/frontend/templates/components/sidebar.html#L176-L209

3. **Delete confirmation dialog has no Escape key handler** (accessibility / consistency regression)

   The dialog dismisses on `@click.self` (backdrop click) and the Cancel button, but pressing Escape does nothing. Every other interactive overlay in this codebase (e.g. the rename input at line 139) handles `@keydown.escape`. A modal that traps focus without an Escape exit is an accessibility violation (WCAG 2.5.3 / APG dialog pattern).

   Fix: add `@keydown.window.escape="deletingSessionId = null"` to the outer `<div class="fixed inset-0 ...">`.

   https://github.com/pocketpaw/pocketpaw/blob/155cde2ca43aa9a9650987ffa2a7fa454d4fecf9/src/pocketpaw/frontend/templates/components/sidebar.html#L509-L540

---

Re: the two items from the previous review cycle:
- **`deleteSession` null-event crash** — correctly fixed. Calling `deleteSession(deletingSessionId)` with no second argument passes `undefined`, which is falsy, so the `if (event)` guard in `deleteSession` handles it safely. ✓
- **`exportSessionById` auth** — the backend route `/api/memory/session/export` in `dashboard.py` has no auth dependency (consistent with other unauthenticated dashboard endpoints), so `fetch()` without a Bearer header is correct for this self-hosted context. ✓

PR #742, opened 2026-03-25
[FI] Fix invalid OpenAI tool schema by preserving empty properties in tool_bridge

umarkkhann3:
## Summary

Fixes an issue where tool schemas with empty `"properties": {}` were being transformed into invalid schemas for OpenAI Agents.

## Root Cause

In `tool_bridge.py`, empty `"properties"` fields were removed:

```python
if not props and "properties" in params_schema:
    params_schema.pop("properties")
    
    This resulted in:

{ "type": "object" }

which is rejected by OpenAI.

Fix
Preserve "properties": {} for all object schemas
Only remove "required" if empty
Ensure compatibility with OpenAI function schema requirements
Additional Changes

Updated gmail_list_labels tool to explicitly define:

{ "type": "object", "properties": {}, "required": [] }
Impact
Fixes OpenAI Agents backend failures
Restores functionality for argument-less tools
Testing
Verified locally via uv run pocketpaw --dev
Ran tests: 21 passed

### DIFF
--- a/src/pocketpaw/agents/tool_bridge.py
+++ b/src/pocketpaw/agents/tool_bridge.py
@@ -126,14 +126,21 @@ def build_openai_function_tools(settings: Any, backend: str = "openai_agents") -
 
         defn = tool.definition
 
-        # Sanitize JSON schema: strict providers (e.g. Groq) reject schemas
-        # where 'required' is present but 'properties' is empty or missing.
+        # Sanitize JSON schema for OpenAI Agents SDK.
+        #
+        # OpenAI requires `{"type":"object"}` schemas to include a `properties` key,
+        # even when the tool takes no arguments (so we must keep `properties: {}`).
+        #
+        # Some strict providers reject schemas where `required` is present but empty
+        # while `properties` is empty/missing. For compatibility we only remove
+        # empty `required`, but we never remove `properties`.
         params_schema = dict(defn.parameters) if defn.parameters else {"type": "object"}
-        props = params_schema.get("properties")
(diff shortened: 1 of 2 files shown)


DevRohit06 (commented):
Note: This PR overlaps significantly with #741, which addresses the same `tool_bridge.py` schema issue with a more complete approach (includes tests and a `normalize_schema()` helper in `protocol.py`). These two PRs conflict with each other.

The core fix here (preserving `"properties": {}` instead of stripping it) is correct, but:
- The `gmail.py` change adds `"required": []` which is immediately stripped by the updated bridge logic, making it a no-op. Consider reverting this hunk.
- No tests are included, while #741 adds dedicated test files.

I'd recommend we go with #741's approach since it's more thorough and tested. If you have additional fixes not covered there, please rebase on top of #741 after it merges.

DevRohit06:
Thanks for the contribution! This change has been superseded by recent work already merged to `dev`. Closing as part of stale-PR cleanup — the fix you identified is now landed via separate commits. We really appreciate the effort, and hope to see more PRs from you on open issues.

prakashUXtech:
Closing: dev already does this.

`build_openai_function_tools` in `src/pocketpaw/agents/tool_bridge.py` now pops an empty `required` but always keeps `properties`, which is the behavior this PR was after. Your diagnosis was correct, the fix just arrived by another route.

Thanks for the writeup on why OpenAI needs `properties` present even for zero-argument tools. That reasoning is what made the fix obvious.

PR #760, opened 2026-03-26
Fix/telegram paw command

### LINKED ISSUE #759: FIX /paw command listed in /help but not functioning in Telegram Chat
### Operating System

Ubuntu

### Python Version

3.12.3

### PocketPaw Version

0.4.15

### Installation Method

uv sync (from source)

### Bug Description

The /paw command is listed in the Telegram /help output but does not function when used.

Unlike other commands such as /help, /new, or /resume, /paw is intended to behave like a normal chat message where the command prefix is stripped and only the message body is passed to the agent.

However, currently:

Sending /paw hello produces no response
Sending hello directly works as expected

This indicates that /paw is not being handled correctly and is likely being passed through _handle_command() unchanged, causing the agent to receive /paw hello instead of hello

### Steps to Reproduce

1. uv run pocketpaw (or uv run pocketpaw --telegram)
2. Connect Telegram bot
3. Send /help to see confirm /paw is listed.
4. you will observe no response.
5. you can chat normally otherwise,like say "hello".
6. the response would be normal.

### Expected Behavior

/paw hello should behave the same as sending hello
The command should havbe Stripped  /paw and forwarded only the message body to the agent

### Actual Behavior

/paw hello produces no response
The command is listed in /help but is not implemented/handled
Likely falls through _handle_command() and is treated as raw text /paw hello

No errors shown in terminal. Command is silently ignored.
Screenshot from Telegram is included.
Notice the response after /paw(which is null)

Image

### Relevant Files

SO the file where the command implementation is missing is: telegram_adapter.py
so a missing handler like _handle_paw() is needed. and command registration.

I would like to work on this.

### Before submitting

suhaib-lone:
this pr depends on this pr:  #741 (the openai won't send respone, more details on this pr)
## What does this PR do?

Implements a dedicated handler for the /paw command in the Telegram adapter.

Currently, /paw is listed in /help but does not function. This PR ensures that /paw <message> behaves like a normal chat message by stripping the command prefix and forwarding only the message body to the agent.

## Related Issue

Fixes #759

## Changes Made
in telegram_adapter.py:
Added _handle_paw() to process /paw messages by stripping the command prefix and forwarding clean input to the agent. Also handles empty input with a usage message.
 in test_telegram_adapter.py:
Added/updated tests to verify /paw behavior and ensure correct message handling.

## How to Test

1.uv run pocketpaw
2.Connect Telegram bot
3.Send /paw hello → should respond like hello
4.Send /paw → should return usage message
5.Send normal message (hello) → should still work

## Evidence of Testing
testpaw

pawfix

## Checklist

### DIFF
--- a/src/pocketpaw/agents/tool_bridge.py
+++ b/src/pocketpaw/agents/tool_bridge.py
@@ -23,7 +23,7 @@ import logging
 from typing import Any
 
 from pocketpaw.tools.policy import ToolPolicy
-from pocketpaw.tools.protocol import BaseTool
+from pocketpaw.tools.protocol import BaseTool, normalize_schema
 from pocketpaw.tools.registry import ToolRegistry
 
 logger = logging.getLogger(__name__)
@@ -126,14 +126,8 @@ def build_openai_function_tools(settings: Any, backend: str = "openai_agents") -
 
         defn = tool.definition
 
-        # Sanitize JSON schema: strict providers (e.g. Groq) reject schemas
-        # where 'required' is present but 'properties' is empty or missing.
-        params_schema = dict(defn.parameters) if defn.parameters else {"type": "object"}
-        props = params_schema.get("properties")
-        if not props and "required" in params_schema:
-            params_schema.pop("required")
-        if not props and "properties" in params_schema:
-            params_schema.pop("properties")
+        # Reuse the shared schema normalizer so zero-arg tools survive strict OpenAI validation.
+        params_schema = normalize_schema(defn.parameters or {"type": "object"})
 
         ft = FunctionTool(
             name=defn.name,
--- a/src/pocketpaw/bus/adapters/telegram_adapter.py
+++ b/src/pocketpaw/bus/adapters/telegram_adapter.py
@@ -72,6 +72,7 @@ class TelegramAdapter(BaseChannelAdapter):
         # Add Handlers
         self.app.add_handler(CommandHandler("start", self._handle_start))
         _cmds = (
+            "paw",
             "new",
             "sessions",
             "resume",
@@ -87,7 +88,8 @@ class TelegramAdapter(BaseChannelAdapter):
             "kill",
         )
         for cmd_name in _cmds:
-            self.app.add_handler(CommandHandler(cmd_name, self._handle_command))
+            handler = self._handle_paw if cmd_name == "paw" else self._handle_command
+            self.app.add_handler(CommandHandler(cmd_name, handler))
         media_filter = (
             filters.PHOTO
             | filters.Document.ALL
@@ -113,6 +115,7 @@ class TelegramAdapter(BaseChannelAdapter):
 
             await self.app.bot.set_my_commands(
                 [
+                    BotCommand("paw", "Send a message to PocketPaw"),
                     BotCommand("new", "Start a fresh conversation"),
                     BotCommand("sessions", "List your conversation sessions"),
                     BotCommand("resume", "Resume a previous session"),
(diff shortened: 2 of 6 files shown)


DevRohit06 (commented):
## Code Review — PR #760 Fix/telegram-paw-command

**Summary:** The core `/paw` command fix is correct and well-structured. The `normalize_schema` refactor solves a real Groq/strict-OpenAI validation bug. Tests are solid. A few issues worth addressing before merge.

---

### Critical

#### PR targets `main`, not `dev` (process violation)
The checklist in the PR description says "PR targets `dev` branch (not `main`)" is checked, but the actual base branch is `main`. This is a process violation per contribution guidelines and risks landing un-staged work directly on the release branch. Please re-target to `dev`.

---

### High

#### Double normalization of OpenAI schema ([`tool_bridge.py` L130](https://github.com/pocketpaw/pocketpaw/blob/1e9e4e0043c1898ee6621c1ca1e8c912d8d86bd4/src/pocketpaw/agents/tool_bridge.py#L130) + [`protocol.py` L39](https://github.com/pocketpaw/pocketpaw/blob/1e9e4e0043c1898ee6621c1ca1e8c912d8d86bd4/src/pocketpaw/tools/protocol.py#L39))

`normalize_schema` is now called in two places for the same schema:

1. `ToolDefinition.to_openai_schema()` (protocol.py L39)
2. `build_openai_function_tools()` in tool_bridge.py (L130), which calls `normalize_schema(defn.parameters ...)` — bypassing `to_openai_schema()` and normalizing the raw `parameters` dict a second time.

Today `normalize_schema` is idempotent so there's no observable bug, but this is a maintenance hazard. If the function ever becomes non-idempotent (e.g. to handle `additionalProperties: false` for strict mode), it will silently double-transform schemas.

**Recommendation:** `build_openai_function_tools` should use `defn.to_openai_schema()["function"]["parameters"]` rather than calling `normalize_schema(defn.parameters)` directly, making `protocol.py` the single source of truth.

```python
# In tool_bridge.py — instead of:
params_schema = normalize_schema(defn.parameters or {"type": "object"})

# Use:
params_schema = defn.to_openai_schema()["function"]["parameters"]
```

---

### Medium

#### `normalize_schema` silently resets `required` even when the caller provided one ([`protocol.py` L18](https://github.com/pocketpaw/pocketpaw/blob/1e9e4e0043c1898ee6621c1ca1e8c912d8d86bd4/src/pocketpaw/tools/protocol.py#L18))

```python
if not schema["properties"]:
    schema["required"] = []   # unconditional overwrite
```

If a tool declares `properties: {}` but an existing `required: ["x"]` (a schema bug, yes, but defensible to detect), the function silently discards the caller's `required` instead of preserving it or raising. A safer guard would be:

```python
schema.setdefault("required", [])
```

This is idempotent, preserves any explicit `required` list, and still satisfies strict validators for zero-arg tools.

---

#### `_handle_paw`: `getattr(context, "args", [])` is unnecessary ([`telegram_adapter.py` L436](https://github.com/pocketpaw/pocketpaw/blob/1e9e4e0043c1898ee6621c1ca1e8c912d8d86bd4/src/pocketpaw/bus/adapters/telegram_adapter.py#L436))

`CommandHandler` always sets `context.args` (it's never missing). `getattr(..., "args", [])` with a default suggests defensive coding against a condition that cannot occur with PTB, and could mask real attribute errors during testing with mocks that don't set `args`. This is consistent with how the rest of the adapter uses `context.args` directly. Use `context.args or []` instead.

---

### Positive

- The `_handle_paw` implementation mirrors `_handle_message` faithfully: topic-aware `chat_id`, typing indicator, proper `InboundMessage` construction. No content is forwarded to the agent.
- Stripping the `/paw` prefix by using `context.args` (PTB's built-in arg parser) rather than string-splitting on `update.message.text` is the right approach.
- Tests cover both the happy path and the empty-body edge case with correct mock structure.
- Moving the schema sanitisation from `tool_bridge.py` into a shared `normalize_schema()` in `protocol.py` is the right abstraction — fixes the same bug at the definition layer so all backends (not just OpenAI Agents SDK) benefit.

---

**Verdict:** Please re-target the PR to `dev`, fix the double-normalization (pick one call site), and use `schema.setdefault("required", [])` in `normalize_schema`. The `/paw` handler itself is correct and ready to ship after those changes.

DevRohit06 (changes requested):
Wrong base branch, double schema normalization, required list overwrite. See comment.

DevRohit06:

PR #767, opened 2026-03-27
[FI] Add distributed circuit breaker and full-jitter exponential backoff for agent reliability

Dilip-chendra:
## Overview
This PR introduces a production-grade reliability layer to the `AgentRouter` to eliminate "Retry Storms" and handle transient LLM provider instability with mathematical precision. It adopts distributed systems best practices (Netflix/Amazon style) for fault tolerance.

## Problem
LLM API backends are notoriously unstable (502, 529, 429 errors). The existing router lacked:
- **Self-Healing:** No way to "stop hitting a dead service" immediately, leading to wasted latency.
- **Congestion Control:** Retries were predictable, creating "thundering herd" spikes on recovering services.
- **Latency SLAs:** No global bounds on how long a request sequence could take.
## Proposed Solution
1. **Stateful Circuit Breaker (`circuit_breaker.py`):** Implemented a `CLOSED` -> `OPEN` -> `HALF-OPEN` state machine. It trips after 5 consecutive definitive failures, protecting the system (and budget) for 60s.
2. **AWS Full Jitter Retries (`retry.py`):** Implemented a Full Jitter exponential backoff algorithm to spread API load and prevent synchronized spikes.
3. **Global Latency Budget:** Introduced a strict 60s total budget that preemptively aborts retries if the tail latency exceeds the SLA.
## Technical Details
- Added `CircuitBreaker` class with thread-safe state tracking.
- Implemented `with_retry` wrapper with AWS Full Jitter logic.
- Pipelined the `AgentRouter` to inject these layers between the selection logic and backend execution.
## Testing
- `tests/test_agent_retry.py`: Verified jitter distribution and budget enforcement.
- `tests/test_circuit_breaker.py`: Verified state transitions under failure simulation.
## Pull Request Checklist
- [x] Pre-commit hooks pass
- [x] Backend tests pass (`uv run pytest tests/test_agent_retry.py tests/test_circuit_breaker.py`)
- [x] New config fields added to `Settings`

### DIFF
--- a/src/pocketpaw/agents/circuit_breaker.py
+++ b/src/pocketpaw/agents/circuit_breaker.py
@@ -1,0 +1,142 @@
+"""Circuit Breaker for Agent Backends.
+
+Provides a state machine (CLOSED, OPEN, HALF-OPEN) to prevent retry storms
+and fail fast when an LLM provider is definitively down. Designed specifically
+to wrap async generators (like AgentBackend.run).
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+import time
+from collections.abc import AsyncIterator
+from enum import Enum
+from typing import Any, Callable
+
+from pocketpaw.agents.protocol import AgentEvent
+
+logger = logging.getLogger(__name__)
+
+
+class CircuitState(Enum):
+    """The state of a circuit breaker."""
+    CLOSED = "CLOSED"           # Normal operation, requests allowed
+    OPEN = "OPEN"               # Failing, requests immediately rejected
+    HALF_OPEN = "HALF_OPEN"     # Probing to see if service recovered
+
+
+class CircuitOpenException(Exception):
+    """Raised when a request is attempted while the circuit is OPEN."""
+    def __init__(self, backend_name: str) -> None:
+        super().__init__(f"Circuit breaker is OPEN for backend: {backend_name}")
+        self.backend_name = backend_name
+
+
+class CircuitBreaker:
+    """A distributed systems circuit breaker for an LLM backend."""
+
+    def __init__(
+        self,
+        backend_name: str,
+        failure_threshold: int = 5,
+        recovery_timeout: float = 60.0,
+    ) -> None:
+        """Initialize the circuit breaker.
+
+        Args:
+            backend_name: The name of the backend this breaker protects.
+            failure_threshold: Number of consecutive failures before tripping OPEN.
+            recovery_timeout: Seconds to wait in OPEN before transitioning to HALF-OPEN.
+        """
+        self.backend_name = backend_name
+        self._failure_threshold = failure_threshold
+        self._recovery_timeout = recovery_timeout
+
+        self._state = CircuitState.CLOSED
+        self._failure_count = 0
+        self._last_failure_time: float = 0.0
+
+    @property
+    def state(self) -> CircuitState:
+        """Get the current state, handling automatic transition to HALF-OPEN."""
+        if self._state == CircuitState.OPEN:
+            now = time.monotonic()
+            if now - self._last_failure_time >= self._recovery_timeout:
+                self._transition_to(CircuitState.HALF_OPEN)
+        return self._state
+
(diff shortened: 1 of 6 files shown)


DevRohit06 (changes requested):
## Review

The overall design (circuit breaker + jitter retry as composable layers) is sound. However, there are correctness bugs:

### Critical

**1. `_wrap_backend_execution` return type mismatch — show-stopper**
`cb.wrap_generator(retriable_gen)` is an `async def` (returns a coroutine), but the caller does `async for event in gen:` which iterates the coroutine object, not the events. Needs `async for event in await gen:` or refactor `_wrap_backend_execution` to be an async generator that yields from the inner generator.

**2. Partial streaming then retry causes duplicate messages**
If a backend yields 10 events then fails on event 11, retry restarts from scratch. The consumer already received events 1-10 and will get them again — duplicate messages in the chat UI. At minimum document this, ideally skip retries after events have been yielded.

### Must Fix

**3. Test patches target wrong module paths**
`@patch("asyncio.sleep", ...)` won't intercept the call in `retry.py` since Python resolves at import time. Patch `pocketpaw.agents.retry.asyncio.sleep` instead. Same issue with `time.monotonic`.

**4. No router-level integration test**
Individual module tests exist but nothing tests `AgentRouter.run()` with the new wrapping. Bug #1 would've been caught.

**5. `is_transient_error` substring matching too aggressive**
`"529"` matches any error containing "529" (e.g., "Error at line 529"). `"timed out"` is extremely broad. `ConnectionRefusedError` is arguably permanent if the endpoint is wrong.

### Minor
- "Thread-safe" claim is misleading — this is async-safe (single-threaded event loop), not thread-safe
- `getattr(self.settings, "agent_max_retries", 3)` is unnecessary — these fields are defined in this PR with defaults
- Import ordering: `circuit_breaker` should come before `retry` (alphabetical)
- `test_auth_401_substring` actually tests a 403 error — rename it

prakashUXtech (changes requested):
## Verdict: split into smaller PRs

**What it does:** Adds an exponential-backoff retry wrapper with full-jitter + time budget, plus a circuit breaker state machine around each agent backend in AgentRouter.

The implementation quality is genuinely high — `with_retry` is well-structured, the transient-error classifier handles httpx/anthropic/openai correctly, and the `CircuitOpenException` / state machine logic is sound. This is serious engineering work.

Two concerns that keep this from landing as-is:

**1. Scope — too much at once.** The router went from "no retry, no breaker" to "retry + breaker wired through the full run path" in a single PR. For a local-first agent framework at pocketpaw's current scale, neither layer has been hit by a real incident yet — both are preemptive. If we land them together and something breaks later, we won't know which layer is to blame. Preference: land retry-only first (clearer value today for rate-limit situations), and gate the circuit breaker behind a config flag defaulted off until we have evidence the retry storms are real.

**2. "Distributed" in the title and docstrings is misleading.** This is single-process, single-router state. Nothing here spans nodes. Dropping "Distributed" from the title keeps the description accurate and avoids setting expectations the code can't meet.

**Smaller things:**

- `retry.py:is_transient_error()` — the substring fallback at the end can catch `"529"` or `"rate limit"` inside the message text of an exception that the structured checks already classified as permanent (e.g., an httpx `HTTPStatusError` with code 400 whose message happens to contain those words). The substring check should only run when all structured checks have returned False, not act as a last-word fallback that can override them.
- The `yield AgentEvent(type="thinking", content="Retrying…")` is nice UX but should include the backend name so users can tell which backend is retrying when multiple are wired.
- Tests — the diff adds a lot of infra but I don't see tests for the jitter math, the time-budget cutoff, or the HALF-OPEN → OPEN transition. Those are the edge cases where bugs hide.

**Suggested split:**

- **PR-a:** `retry.py` + router wiring + tests for retry behavior (including time-budget and jitter bounds)
- **PR-b:** `circuit_breaker.py` + wiring, gated by `agent_circuit_breaker_enabled: bool = False` config, with tests for the state transitions

Happy to review either quickly once they land.

DevRohit06:

PR #768, opened 2026-03-27
[FI] Add global execution timeouts and cancellation handling to AgentRouter

Dilip-chendra:
## Overview
Implements a hard global timeout for the AgentRouter to prevent Zombie Requests and ensure resource availability during long-running or stalled operations.

## Problem
Without a global execution timeout, a hanging LLM provider, a slow third-party tool, or an infinite generation loop could block an execution thread/session indefinitely. This leads to leaked resources and poor UX in the dashboard.

## Proposed Solution
- Asyncio Timeout Integration: Wrapped the entire AgentRouter.run generator pipeline in a configurable asyncio.timeout context.
- - Graceful Error Recovery: Intercepts TimeoutError to yield a clean AgentEvent(type="error") to the user before closing the backend stream.
## Pull Request Checklist
* [x] Pre-commit hooks pass
* [x] Tests pass (uv run pytest tests/test_router_timeouts.py)
* [x] New config field agent_execution_timeout_seconds added to Settings

### DIFF
--- a/src/pocketpaw/agents/router.py
+++ b/src/pocketpaw/agents/router.py
@@ -5,6 +5,7 @@ configured agent backend. Supports optional user-configured fallback
 backends if the primary backend fails.
 """
 
+import asyncio
 import logging
 from collections.abc import AsyncIterator
 from typing import Any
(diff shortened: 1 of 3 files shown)


DevRohit06 (changes requested):
## Review

Wrapping `AgentRouter.run()` in `asyncio.timeout()` is the right idea. However:

### Must Fix

**1. Missing `done` event after timeout**
The timeout handler yields an `error` event then `return`s. But all other exit paths yield both `error` and `done`. Consumers like `AgentLoop` rely on `done` to finalize (flush streams, save memory). Fix:
```python
except TimeoutError:
    yield AgentEvent(type="error", content=f"Agent execution timed out after ...")
    yield AgentEvent(type="done", content="")
    return
```

**2. No backend cleanup on timeout**
When timeout fires, the backend's async generator is abandoned mid-iteration. `backend.stop()` is never called — could leak HTTP connections, subprocess handles, etc. Add cleanup in the timeout handler.

### Moderate

**3. Timeout scope includes consumer processing time**
`asyncio.timeout` inside an async generator spans not just backend work but also the time the consumer spends processing each yielded event (memory saves, WebSocket broadcasts, etc.). With 300s default this is unlikely to bite, but worth documenting.

### Suggestions
- Add a test for timeout with fallback backends configured
- Add a test asserting `done` event follows the `error` event
- Consider `ge=10` instead of `gt=0` to prevent obviously-too-low timeouts

prakashUXtech (approved):
## #768 — Global execution timeouts for AgentRouter

The implementation is straightforward and correct. Wrapping the entire routing pipeline in `asyncio.timeout()` (3.11+ context manager) is the right approach — it cancels cleanly and doesn't require thread wrangling. The 300s default is sane for LLM workloads.

**One real issue:** the `asyncio.timeout()` context manager cannot span a `yield` statement directly in an async generator — this is a Python constraint. If the `async with asyncio.timeout(...)` wraps `yield` expressions inside the generator, Python will raise `RuntimeError: async generator ignored GeneratorExit`. The diff shows the timeout wrapping the `yield event` lines directly. This needs to be restructured, likely by collecting events from a task and yielding outside the timeout context, or using `asyncio.wait_for` on a helper coroutine. The test `test_router_execution_timeout` may not catch this because `MockHangingBackend` sleeps before its second yield rather than at a yield boundary.

**Also worth noting:** when the timeout fires, the code yields an error event but does not yield a `"done"` event. Downstream consumers that loop until `event.type == "done"` will hang or error. The error path should terminate with a `done` event:

```suggestion
        except TimeoutError:
            logger.warning("Agent execution timed out after %ds",
                           self.settings.agent_execution_timeout_seconds)
            yield AgentEvent(
                type="error",
                content=f"Agent execution timed out after {self.settings.agent_execution_timeout_seconds} seconds."
            )
+           yield AgentEvent(type="done", content="")
            return
```

The async generator + timeout scoping issue is a correctness bug that should be fixed before merge. Requesting changes on that basis, but the overall approach and the config addition are right.

prakashUXtech (changes requested):
Correction on my earlier review — the state was marked 'approved' by mistake. The async generator + `asyncio.timeout()` scoping issue flagged above is a correctness blocker, not a nit. Treating this as request-changes until that's resolved. Everything else in the prior comment still stands.

DevRohit06:

PR #769, opened 2026-03-27
[FI] Add startup-time Pydantic validation for backend authentication

Dilip-chendra:
## Overview
Introduces a fail-fast configuration validation layer using Pydantic model_validators to ensure the application is correctly configured at boot time.

## Problem
Previously, missing API keys for a selected backend (e.g. Claude SDK) would only cause an error at runtime when the first prompt was sent. This led to cryptic stack traces deep in the networking layer rather than clear configuration errors.

## Proposed Solution
- Cross-Field Validation: Added @model_validator(mode="after") to Settings to verify that if a backend like claude_agent_sdk is enabled, its corresponding anthropic_api_key or oauth_token is present.
- - Immediate Feedback: Crashes gracefully on startup with a descriptive ValidationError, significantly improving the developer experience.
## Pull Request Checklist
* [x] Pre-commit hooks pass
* [x] Tests pass (uv run pytest tests/test_config_validation.py)

### DIFF
--- a/src/pocketpaw/config.py
+++ b/src/pocketpaw/config.py
@@ -18,9 +18,8 @@ import logging
 import re
 from functools import lru_cache
 from pathlib import Path
-from typing import Literal
-
-from pydantic import Field
+from typing import Literal, Self
+from pydantic import Field, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 logger = logging.getLogger(__name__)
@@ -878,6 +877,34 @@ class Settings(BaseSettings):
     max_concurrent_conversations: int = Field(
         default=5, gt=0, description="Max parallel conversations processed simultaneously"
     )
+
+    @model_validator(mode="after")
+    def validate_backend_auth(self) -> Self:
+        """Validate that the selected agent backend has necessary authentication."""
+        backend = self.agent_backend
+
+        if backend == "claude_agent_sdk":
+            if self.claude_sdk_provider == "anthropic" and not (
+                self.anthropic_api_key or self.claude_code_oauth_token
+            ):
+                raise ValueError(
+                    "claude_agent_sdk (anthropic) requires either anthropic_api_key "
+                    "or claude_code_oauth_token."
+                )
+        elif backend == "openai_agents":
+            if self.openai_agents_provider == "openai" and not self.openai_api_key:
+                raise ValueError("openai_agents requires openai_api_key.")
+        elif backend == "google_adk":
+            if self.google_adk_provider == "google" and not self.google_api_key:
+                raise ValueError("google_adk requires google_api_key.")
+        elif backend == "codex_cli" and not self.openai_api_key:
+            raise ValueError("codex_cli requires openai_api_key.")
+        elif backend == "copilot_sdk":
+            if self.copilot_sdk_provider == "copilot" and not self.openai_api_key:
+                # Copilot SDK often uses openai_api_key as its proxy/auth if configured
+                pass
+        
+        return self
 
     def save(self) -> None:
         """Save settings to config file.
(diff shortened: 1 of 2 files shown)


DevRohit06 (changes requested):
## Review

The intent is good — validating API keys at startup is useful. However, there are critical issues:

### Critical

**1. Deletes ~370 lines of existing tests**
The PR completely replaces `test_config_validation.py`, removing all existing tests for `validate_api_key()`, numeric field constraints, and enum validation. These cover real, used functionality. **Add** the new tests alongside existing ones, don't replace them.

**2. Breaks CLI commands for unconfigured users**
The `@model_validator` runs during `Settings.load()`, which is called by `pocketpaw config`, `pocketpaw status`, etc. A user who hasn't configured an API key yet will get a `ValidationError` crash instead of being able to run `pocketpaw config set anthropic_api_key sk-ant-...`. This fundamentally breaks the configuration workflow. Move validation to a separate `validate_for_startup()` method or skip it during config loading.

**3. `copilot_sdk` branch is a no-op**
```python
elif backend == "copilot_sdk":
    if self.copilot_sdk_provider == "copilot" and not self.openai_api_key:
        pass  # does nothing
```
Either raise a `ValueError` or remove the branch entirely.

**4. `deep_agents` and `opencode` backends not handled**
Both are valid backend values but have no validation branch. Incomplete implementation.

### Must Fix

**5. No test for `codex_cli`, OAuth token fallback, or provider-skipping behavior**
e.g., `claude_sdk_provider="litellm"` should not require `anthropic_api_key` — no test verifies this.

**6. Empty string vs None inconsistency**
OpenAI test uses `""`, Claude test uses `None`. Document whether empty strings count as "missing".

### Summary
Don't delete existing tests, fix the Settings construction crash for CLI commands, complete the missing backend branches, and add test coverage for the gaps.

prakashUXtech (approved):
## #769 — Startup-time backend auth validation

This is a good change. Moving auth checks into a `@model_validator(mode="after")` means misconfigured deployments fail immediately at startup with a clear error message instead of surfacing as a cryptic 401 mid-conversation. The -369 line reduction is particularly welcome — the old test file had grown into a sprawling class hierarchy that was testing `validate_api_key()` in isolation when what users actually care about is "does the configured backend have what it needs to run."

The new tests are tighter and test the right thing.

A couple of things to be aware of:

**The `copilot_sdk` branch is a no-op.** The current code hits the `elif backend == "copilot_sdk"` branch, evaluates the condition, and then falls through to `pass` without raising. That's fine if Copilot SDK genuinely doesn't require a key (e.g. it uses browser auth), but the comment "Copilot SDK often uses openai_api_key as its proxy/auth if configured" suggests this is unresolved. Either validate the key or drop the branch entirely — a silent `pass` in a validator reads as a mistake to the next person touching this code.

**The deleted tests covered things this PR doesn't.** The old `TestValidateApiKey` class had prefix-format checks (e.g. rejecting `sk-wrong-abc123` for an Anthropic key). Those tests are now gone. If `validate_api_key()` is still in use elsewhere, those cases are no longer covered. If that function is dead code after this change, it should be deleted too.

Neither of these is a blocker. The validator logic is correct for the cases it handles. Approving.

DevRohit06:
Thanks for tackling startup validation — this is real value. Unfortunately we can't merge as-is: putting `@model_validator(mode="after")` on Settings means every `Settings.load()` call crashes for users who haven't configured their API key yet, which breaks `pocketpaw config set`, `pocketpaw doctor`, `pocketpaw status`, and the first-run flow. That's a regression worse than the original missing-key UX. The right home for this check is a separate `validate_for_startup()` method called from the agent-loop entrypoint, not Settings construction. Closing for now — we'll redesign internally. Really appreciate the contribution and the clean tests.

PR #770, opened 2026-03-27
[FI] Implement structured logging with latency tracking and metadata-…

Dilip-chendra:
## Overview
Standardizes logging across the AgentRouter to include structured metadata and precise latency metrics, making PocketPaw production-ready for observability stacks.

## Problem
Existing logs were mostly free-text strings. This made it impossible to programmatically track which backends were failing most often or determine the latency impact of fallback chains in tools like Datadog or ELK.

## Proposed Solution
- Metadata Injection: Standardized all critical router logs to use the extra={} field containing event, backend, and is_primary attributes.
- - Latency Tracking: Integrated time.perf_counter() to calculate and log the exact latency_ms for every successful or failed backend execution.
## Pull Request Checklist
* [x] Pre-commit hooks pass
* [x] Tests pass (uv run pytest tests/test_structured_logging.py)

### DIFF
--- a/src/pocketpaw/agents/router.py
+++ b/src/pocketpaw/agents/router.py
@@ -6,6 +6,7 @@ backends if the primary backend fails.
 """
 
 import logging
+import time
 from collections.abc import AsyncIterator
 from typing import Any
 
@@ -101,6 +102,10 @@ class AgentRouter:
 
         # Primary backend (streaming, no buffering, no error-event fallback)
         if self._backend is not None:
+            backend_name = self._active_backend_name or "unknown"
+            logger.info("Starting primary backend: %s", backend_name, 
+                        extra={"event": "backend_start", "backend": backend_name, "is_primary": True})
+            start_time = time.perf_counter()
             try:
                 async for event in self._backend.run(
                     message,
@@ -111,14 +116,21 @@ class AgentRouter:
                     yield event
 
                     if event.type == "done":
+                        latency_ms = (time.perf_counter() - start_time) * 1000
+                        logger.info("Primary backend '%s' completed successfully", backend_name,
+                                    extra={"event": "backend_success", "backend": backend_name, 
+                                           "latency_ms": latency_ms, "is_primary": True})
                         return
 
             except Exception as exc:
+                latency_ms = (time.perf_counter() - start_time) * 1000
                 last_error = str(exc)
                 logger.warning(
                     "Primary backend '%s' failed: %s",
-                    self._active_backend_name,
+                    backend_name,
                     exc,
+                    extra={"event": "backend_failure", "backend": backend_name, 
+                           "error": last_error, "latency_ms": latency_ms, "is_primary": True}
                 )
 
         # Fallback backends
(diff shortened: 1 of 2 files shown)


DevRohit06 (changes requested):
## Review

The concept is sound — adding structured `extra={}` metadata to log calls in `AgentRouter.run()` with latency tracking. However, several issues:

### Must Fix

**1. "All backends failed" path has no structured metadata**
This is the most important event to track in production, yet lines 154-159 emit no `extra={}` fields. Add structured logging here too.

**2. Tests don't actually exercise the patched code**
`patch("...get_backend_class")` is set up but never called — `_backend` and `_active_backend_name` are manually assigned, bypassing `_initialize_backend()` entirely. The mock/patch is dead code in the test.

**3. Missing test cases**
- All backends fail (terminal error path)
- Fallback also fails (chained failures)
- Fallback unavailable (`_get_fallback_backend` returns `None`)

**4. Line length violations**
Several new `logger.info(...)` / `logger.warning(...)` calls likely exceed the project's 100-char limit. Run `ruff format`.

### Suggestions

- **Sleep-based latency assertions** (`assert latency_ms >= 100`) will be flaky in CI. Mock `time.perf_counter` instead, or use a generous lower bound.
- **Repetition** — the structured logging pattern is copy-pasted 4+ times. Consider a small `_log_extra(event, backend, is_primary, **kw)` helper.
- **`if False: yield` trick** in `FailingBackend.run()` needs a comment explaining why.
- **`or "unknown"` fallback** on `_active_backend_name` is unreachable inside the `if self._backend is not None` block — dead code.

prakashUXtech (approved):
## #770 — Structured logging with latency tracking

Clean, targeted change. Adding `time.perf_counter()` bookends and structured `extra={}` fields to the router's backend start/success/failure paths is exactly the kind of observability that makes production diagnosis possible. The event key naming is consistent and the fields (`event`, `backend`, `latency_ms`, `is_primary`) are useful for any log aggregator.

A few notes:

**No new JSONL audit trails** — confirmed. This writes to the existing Python logger only, which is correct. Audit-relevant events (tool use, security decisions) go through `AuditLogger`; this is operational telemetry, which belongs in the regular log stream.

**The `error` field in failure logs contains the raw exception string.** `last_error = str(exc)` and then `"error": last_error` is fine for most errors, but if the exception message contains something sensitive (an API key in a URL, a credential in an error response body) the secret filter won't catch it here because `SecretFilter` operates on `record.getMessage()`, not on `extra` fields. The existing structured logging in the codebase has the same gap, so this isn't a regression, but it's worth knowing.

**The tests use `caplog` to assert on `record.latency_ms >= 100` after a 100ms sleep.** That assertion is timing-dependent and will flake on a loaded CI runner. Consider `>= 90` or asserting `latency_ms > 0` if the exact threshold isn't load-bearing.

None of these are blocking. The change does what it says, the tests cover the right paths, and no deprecated patterns (JSONL audit trails, etc.) were introduced. Approving.

DevRohit06:

PR #771, opened 2026-03-27
[FI] Enhance AgentRouter test coverage for initialization and stop ed…

Dilip-chendra:
## Overview
Hardens the AgentRouter and Registry by adding an exhaustive unit test suite for lifecycle edge cases and initialization failures.

## Problem
The router's initialization logic (handling missing backends and instantiation errors) lacked dedicated test coverage, making it prone to regressions as the number of supported backends grows.

## Proposed Solution
- Edge Case Coverage: Added tests for invalid backend name fallbacks, instantiation exception handling, and resilient stop() behavior.
- - Improved Stability: Ensures the router always defaults to a safe state even if the primary backend configuration is corrupted.
## Pull Request Checklist
* [x] Pre-commit hooks pass
* [x] Tests pass (uv run pytest tests/test_router_init.py)

### DIFF
--- a/tests/test_router_init.py
+++ b/tests/test_router_init.py
@@ -1,0 +1,81 @@
+"""Tests for AgentRouter initialization and edge cases."""
+
+from unittest.mock import MagicMock, patch
+import pytest
+from pocketpaw.agents.router import AgentRouter
+
+
+class MockSettings:
+    """Minimal settings mock."""
+    def __init__(self, backend="claude_agent_sdk", fallbacks=None):
+        self.agent_backend = backend
+        self.fallback_backends = fallbacks or []
+
+
+def test_router_init_invalid_backend(caplog):
+    """Verify that router falls back to 'claude_agent_sdk' if configured backend is missing."""
+    settings = MockSettings(backend="invalid_unknown_backend")
+    
+    # Mock registry to return None for the invalid backend, but a real class for the fallback
+    def mock_get_backend(name):
+        if name == "claude_agent_sdk":
+            mock_cls = MagicMock()
+            mock_cls.info.return_value = MagicMock(display_name="Claude")
+            return mock_cls
+        return None
+
+    with patch("pocketpaw.agents.router.get_backend_class", side_effect=mock_get_backend):
+        router = AgentRouter(settings)
+        # Should have defaulted to claude_agent_sdk
+        assert router._active_backend_name == "claude_agent_sdk"
+        assert "unavailable — falling back to claude_agent_sdk" in caplog.text
+
+
+def test_router_init_exception_handling(caplog):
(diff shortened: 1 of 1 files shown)


DevRohit06 (changes requested):
## Review

The tests cover useful edge cases (invalid backend fallback, init exception, stop resilience, no-backend info). However, several have correctness issues that give false confidence:

### Must Fix

**1. `test_router_stop_resilience` uses `MagicMock` for async methods**
`mock_primary.stop.side_effect = Exception(...)` raises synchronously before `await`. Use `AsyncMock` for proper async semantics. Additionally, the lambda passed to `get_backend_class` lacks `.info()`, causing a silent init failure — then `router._backend = mock_primary` papers over it. The test would pass even if `stop()` didn't handle exceptions.

**2. Use `Settings` or `MagicMock(spec=Settings)` instead of hand-rolled `MockSettings`**
The existing tests in `test_router_fallback.py` use the real `Settings` class. `MockSettings` will silently break if `AgentRouter.__init__` accesses additional attributes in the future.

**3. `BrokenBackend.info` missing `@classmethod` decorator**
The real backend interface uses `@classmethod` for `info()`. Match the pattern in `test_router_fallback.py`.

### Suggestions

- Consider consolidating into existing `tests/test_agents.py::TestAgentRouter` or `tests/test_router_fallback.py` rather than creating a third router test file.
- The `@pytest.mark.asyncio` markers are unnecessary since the project has `asyncio_mode = "auto"`.
- `router._backend = None` in `test_router_get_backend_info_none` is redundant — init already failed.

prakashUXtech (approved):
## #771 — AgentRouter test coverage

These tests are solid and fill a real gap. The init edge cases (invalid backend, broken `__init__`, `stop()` resilience) are exactly the kind of thing that breaks silently in production and doesn't get caught without explicit coverage.

A few things that stand out as good:
- Testing that `router.stop()` swallows backend stop errors without propagating — that's a subtle correctness requirement and it's good to have it locked in.
- Using `caplog` properly to assert on log messages rather than side effects.
- `get_backend_info()` returning `None` when `_backend` is `None` is a real edge case worth pinning.

One minor note: the `test_router_stop_resilience` test uses `MagicMock` for `stop()` which is a sync mock, but `stop()` is async in the actual router. If the test passes today it's probably because the router's `stop()` catches the exception before awaiting, but worth verifying with `AsyncMock` to be precise:

```suggestion
    mock_primary = MagicMock()
    mock_primary.stop = AsyncMock(side_effect=Exception("Stop failed!"))
```

That said, the coverage intent is right and the tests are correct in spirit. Approving — this is a clean addition that improves confidence in the router's resilience paths.

DevRohit06:

PR #772, opened 2026-03-27
feat: Implement Distributed Context Tracing and Correlation IDs

Dilip-chendra:
## Overview
Introduces a `contextvars`-based tracing layer to propagate request identifiers across async boundaries. This enhancement ensures that every log line emitted during a request is tagged with a unique Correlation ID, enabling end-to-end request stitching in modern observability stacks.

## Problem
Currently, it is difficult to correlate log entries from multiple components (agents, backends, tools) for a single request, especially under high concurrency. This leads to:
* Increased time-to-resolution for production issues.
* * Difficulty in tracing the flow of information across async tasks.
* * Lack of end-to-end visibility in distributed execution.
## Proposed Solution
1. **Core mechanism**: Implement a `ContextVar` to store a `correlation_id` that is automatically propagated across async boundaries.
2. 2. **Supporting logic**: Update logging configuration to include the correlation ID in every log record.
3. 3. **Integration strategy**: Update `InboundMessage`, `OutboundMessage`, and `SystemEvent` to carry these IDs, ensuring consistency across the message bus.
## Technical Details
* Added `RequestTracing` utility to manage context identifiers.
* * Modified `logging_setup.py` to use a custom formatter that extracts the correlation ID.
* * Updated event schemas in `src/pocketpaw/bus/events.py`.
## Testing
* Unit tests for context propagation.
* * Verified that logs emitted from nested async calls correctly include the same correlation ID.
## Pull Request Checklist
- [x] Pre-commit hooks pass
- [ ] - [x] Tests added and passing
- [ ] - [x] Config/Env updates included

### DIFF
--- a/src/pocketpaw/bus/adapters/__init__.py
+++ b/src/pocketpaw/bus/adapters/__init__.py
@@ -174,6 +174,29 @@ class BaseChannelAdapter(ABC):
         ...
 
     async def _publish_inbound(self, message: InboundMessage) -> None:
-        """Helper to publish inbound messages."""
+        """Helper to publish inbound messages.
+        
+        Injects a correlation_id if it's missing to ensure traceability.
+        """
+        from pocketpaw.context import generate_correlation_id, set_correlation_id
+
         if self._bus:
+            # If the adapter didn't already set a correlation_id, generate one now.
+            if message.correlation_id == "unset":
+                corrid = generate_correlation_id()
+                # Reconstruct the frozen dataclass with the ID
+                message = InboundMessage(
+                    channel=message.channel,
+                    sender_id=message.sender_id,
+                    chat_id=message.chat_id,
+                    content=message.content,
+                    correlation_id=corrid,
+                    timestamp=message.timestamp,
+                    media=message.media,
+                    metadata=message.metadata,
+                )
+            
+            # Set the ID in the current async context for this task's logging
+            set_correlation_id(message.correlation_id)
+            
             await self._bus.publish_inbound(message)
--- a/src/pocketpaw/bus/events.py
+++ b/src/pocketpaw/bus/events.py
@@ -39,6 +39,7 @@ class InboundMessage:
     sender_id: str
     chat_id: str
     content: str
+    correlation_id: str = field(default_factory=lambda: "unset")
     timestamp: datetime = field(default_factory=datetime.now)
     media: list[str] = field(default_factory=list)
     metadata: dict[str, Any] = field(default_factory=dict)
@@ -55,6 +56,7 @@ class InboundMessage:
             sender_id=self.sender_id,
             chat_id=self.chat_id,
             content=new_content,
+            correlation_id=self.correlation_id,
             timestamp=self.timestamp,
             media=self.media,
             metadata=self.metadata,
@@ -68,6 +70,7 @@ class OutboundMessage:
     channel: Channel
     chat_id: str
     content: str
+    correlation_id: str = field(default_factory=lambda: "unset")
     reply_to: str | None = None
     media: list[str] = field(default_factory=list)
     metadata: dict[str, Any] = field(default_factory=dict)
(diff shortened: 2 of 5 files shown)


DevRohit06 (changes requested):
## Review

The approach is sound — `contextvars` for async correlation ID propagation is the right primitive. However, several issues need addressing:

### Must Fix

**1. `verify_tracing.py` committed to repo root**
This manual test script should not be in the PR. Convert it to a proper test in `tests/` and remove the root file.

**2. Frozen dataclass reconstruction is fragile**
In `_publish_inbound`, manually reconstructing `InboundMessage` with every field will silently break when new fields are added. Use `dataclasses.replace()`:
```python
message = dataclasses.replace(message, correlation_id=corrid)
```

**3. `"unset"` sentinel string**
Using a magic string is error-prone. Use `None` as default and check `if message.correlation_id is None`:
```python
correlation_id: str | None = field(default=None)
```

**4. Correlation ID not propagated to OutboundMessage/SystemEvent**
The field is added to all three event types but nothing in the diff ever sets it on `OutboundMessage` or `SystemEvent`. The tracing chain is incomplete — you can trace inbound but lose the ID on the way out.

**5. CorrelationIdFilter added redundantly**
The filter is added to the handler AND to the root logger. Once on the handler is sufficient — adding it to the root logger means it runs twice for every log record.

**6. No proper tests**
Only the `verify_tracing.py` script exists. Need proper tests in `tests/` covering:
- Context propagation across async boundaries
- Adapter injection when correlation_id is missing
- Log records contain the correlation ID
- `with_content()` preserves correlation_id (already handled, good)

### Nit

**7. Title says "Distributed" but this is single-process `contextvars`**
This is local async context propagation, not distributed tracing (no trace headers, no span IDs, no cross-service propagation). The title/description oversells the scope — consider renaming to "async request correlation" or similar.

prakashUXtech (changes requested):
## Verdict: architecture alignment needed before merge

**What it does:** Adds `contextvars`-based correlation_id propagation across async boundaries, tagging every log line and bus event with a request ID for end-to-end tracing.

The contextvars approach is correct Python (not an easy API to get right the first time), and the `CorrelationIdFilter` + log format change is the right place to plug it in.

**Concerns:**

**1. Relationship with journal `correlation_id` is unaddressed.** Soul Protocol's org-level journal (see `soul-protocol/docs/org-journal-spec.md`, EventEntry schema line 72) already has `correlation_id: UUID | None` as an event-level concept for tying together flows (retrieval → response, etc). This PR adds a second `correlation_id: str` at the bus-event level. They're at different layers and serve different purposes, but nothing wires them together — a bus-level correlation_id should flow into the journal events that a request triggers, so traces stitch across the two systems. Without that, we get two parallel tracing layers that can't be cross-referenced.

**2. The `"unset"` string sentinel is a type lie.** `correlation_id: str = field(default_factory=lambda: "unset")` declares the field as `str`, but `"unset"` is a tombstone value the rest of the code has to treat specially. Align with the journal's typing — either `str | None` (None meaning "not yet set") or bind it to UUID like the journal does. `_publish_inbound` can then check `if message.correlation_id is None` instead of `== "unset"`.

**3. The frozen-dataclass-reconstruct pattern in `_publish_inbound` is fragile.** Every adapter has to do this dance or correlation_id is lost. Cleaner: add a `.with_correlation_id(id)` helper on the dataclass, analogous to the existing `.with_content()`.

**4. No tests for async boundary propagation.** The whole point of contextvars is that IDs flow across `await` and into child tasks. Need at least one test that creates an inbound with a known correlation_id, awaits a nested async function that logs, and asserts the log line carries the same ID. Same for the "unset → generated" path through `_publish_inbound`.

**Suggested direction before re-review:**

- Align typing with journal (`UUID | None` or `str | None`)
- Add a `.with_correlation_id()` helper on the dataclasses; drop the inline reconstruct
- Add tests for async-boundary propagation
- Document (in the PR description or a short comment) how this interacts with journal `correlation_id` — are they meant to carry the same value for a given request? If yes, where does the ID flow between the two systems?

The core idea is good and the contextvars implementation is careful. Once it aligns with journal typing and has propagation tests, I'll take another pass.

DevRohit06:

PR #800, opened 2026-03-29
[FI] Fix port binding error — auto-detect free port in range 8000-9000

Pragatisource08:
🚀 Fix: Automatic Port Fallback for Busy Default Port
🐛 Issue

The application fails to start when the default port (8888) is already in use, resulting in a port binding error ([Errno 10048]) and immediate shutdown instead of handling the conflict gracefully.

✅ Solution

Implemented automatic port detection and fallback logic to ensure the server starts reliably even when the default port is occupied.

Starts on default port 8888
If unavailable, increments port number sequentially
Retries binding until a free port is found (within a defined range)
Prevents application crash and ensures smooth startup
🔧 Implementation Details
Added _serve() wrapper to handle server startup with retry logic
Performs a socket-level pre-check before binding
Catches EADDRINUSE / WinError 10048 and retries with next port
Maintains same host during checks to avoid binding inconsistencies
Logs warnings when switching ports
🎯 Result
Application no longer crashes due to port conflicts
Automatically finds and binds to an available port
Improves robustness and developer experience
📁 Files Modified
src/pocketpaw/main.py
✅ Notes
Backward compatible
No breaking changes
Works across platforms (tested on Windows)


DevRohit06:
### Code review

Found 1 issue:

1. Multiple ruff formatting violations in the new `find_free_port` function. CLAUDE.md specifies ruff for linting/formatting with rules E/F/I/UP and line-length 100. The new code has: missing spaces after commas in `find_free_port(start=8000,end=9000)` (E231), space before parenthesis in `range (start,end)` (E211), missing space after comma in `("127.0.0.1",port)` (E231), trailing whitespace on `return port ` (W291), and the error message `"No Free Ports ,Try later "` has a space before the comma and trailing whitespace. Run `uv run ruff check . && uv run ruff format .` to fix.

https://github.com/pocketpaw/pocketpaw/blob/7b2671b1d06e536b81b6ec98c457b2ea80d8b8a8/src/pocketpaw/__main__.py#L172-L180

Pragatisource08:
Thanks for the review! Fixing the ruff formatting violations now.

Pragatisource08:
@DevRohit06 Thanks for the detailed review! 

I've fixed all the ruff formatting violations:
- ✅ Added spaces after commas in function parameters
- ✅ Removed space before parenthesis in `range()`
- ✅ Fixed tuple spacing in `s.bind()`
- ✅ Removed trailing whitespace
- ✅ Cleaned up the error message string

Pushed a new commit: `style: fix ruff formatting in find_free_port`

Ran `uvx ruff check . && uvx ruff format .` — all checks passing now.

DevRohit06:
### Code review

Found 3 issues:

1. **`find_free_port()` called eagerly as argparse `default=` — runs on every CLI invocation, TOCTOU race** (CLAUDE.md says "Entry point: `pocketpaw.__main__:main`" — the port scan should happen inside `main()` only when a server is actually starting, not at parse time)

`default=find_free_port()` evaluates the port scan inside `_build_parser()`, which runs for every subcommand (`pocketpaw status`, `pocketpaw config`, `pocketpaw channels`, etc.) — none of which need a port. This also creates a TOCTOU race: the port is probed and released, then another process can claim it before uvicorn binds. Fix: restore `default=8888` and move the scan to the server-start branch of `main()`.

https://github.com/pocketpaw/pocketpaw/blob/17bf57d6f4a984416050ec9f1106934626217961/src/pocketpaw/__main__.py#L330-L340

2. **Double port scan makes the `[WARN]` message unreachable dead code** (bug due to control flow between argparse default and `main()` re-scan)

After `_build_parser()` already sets `args.port` to a free port via `default=find_free_port()`, the second call `find_free_port(start=args.port)` in `main()` starts from that already-free port and returns the same value. The condition `new_port != args.port` is never true when `--port` is not explicitly supplied, so the `[WARN]` message is dead code and users are never informed of a port switch.

https://github.com/pocketpaw/pocketpaw/blob/17bf57d6f4a984416050ec9f1106934626217961/src/pocketpaw/__main__.py#L504-L510

3. **Removes the stable default port 8888, silently breaking existing integrations** (CLAUDE.md architecture section documents that `interacly-backend` communicates with `backend (PocketPaw) :8888` via Redis bridge)

Replacing `default=8888` with `default=find_free_port()` means the server can start on any port in 8000-9000. All existing configurations, browser bookmarks, and the interacly-backend Redis bridge that expect `localhost:8888` will silently break. The fix should keep `default=8888` and only auto-detect when that specific port is occupied.

https://github.com/pocketpaw/pocketpaw/blob/17bf57d6f4a984416050ec9f1106934626217961/src/pocketpaw/__main__.py#L335-L338

DevRohit06 (changes requested):
find_free_port() called at parse time, dead warn code, breaks stable 8888 default. See comment for details.

Pragatisource08:
@DevRohit06 
Thanks for the detailed review.

You're absolutely right about all three issues:

Calling find_free_port() at parse time is unnecessary and introduces a TOCTOU risk.
The double port scan does make the warning branch effectively unreachable.
And removing the stable default 8888 would break existing integrations, which I overlooked.

I’ll address this by:

Restoring default=8888 in the parser
Moving the port detection logic inside main() only when starting the server
Ensuring auto-detection only kicks in if 8888 is actually occupied, while preserving the warning behavior

Will push a fix shortly.

Pragatisource08:
@DevRohit06 
Hi, I've addressed all 3 issues flagged in the code review:

1.) Eager find_free_port() in argparse default — replaced default=find_free_port() with default=8888 and moved the port scan  into a new _resolve_port_for_server() function that only gets called inside main() when a server is actually starting.
2.) Dead [WARN] code — the double scan is gone. _resolve_port_for_server() is now the single scan point, so the [WARN] message is live and will correctly fire when port 8888 is busy.
3.) Stable default port 8888 restored — interacly-backend and any other integration expecting :8888 will no longer break silently. Auto-detection only kicks in as a fallback when 8888 is occupied.

All changes verified with a standalone test suite (29/29 passed) and pycodestyle formatting check (clean).

Pragatisource08:
@DevRohit06 
10:59 AM
Thanks for the review! Both issues have been addressed:

Critical - except KeyboardInterrupt restored: Added back in the correct position between try and finally, fixing the SyntaxError on startup.

Minor- --port repositioned back: Moved back before --dev, removing the diff noise.

All 3 Python test suites (3.11, 3.12, 3.13) passing. The PR Quality Gate failure is a CI configuration issue — the workflow is throwing HttpError: Resource not accessible by integration when trying to assign labels, which is unrelated to these code changes.

DevRohit06 (changes requested):
Useful quality-of-life fix, but there are a few issues — one of which is a correctness bug.

## Issues

### 1. TOCTOU race condition in port check (High — must fix)

`_resolve_port_for_server` binds to the port to test availability, then **releases it** before the actual server binds. Between the check and the real bind, another process can grab that port:

```python
def _resolve_port_for_server(requested_port: int) -> int:
    try:
        with socket.socket(...) as s:
            s.bind(("127.0.0.1", requested_port))
        # ← port is released here
        return requested_port  # ← another process can steal it before uvicorn binds
```

Same issue exists in `find_free_port()`.

**Fix:** Don't pre-check at all. Instead, catch the `OSError`/`[Errno 10048]` from the actual server startup (`uvicorn`/`run_dashboard_mode`) and retry with the next port. This is both simpler and correct. Something like:

```python
def main():
    ...
    port = args.port
    for attempt in range(10):
        try:
            run_dashboard_mode(settings, host, port, dev=args.dev)
            break
        except OSError as e:
            if e.errno in (errno.EADDRINUSE, 10048):  # Unix + Windows
                new_port = port + 1
                print(f"[WARN] Port {port} busy — trying {new_port}")
                port = new_port
            else:
                raise
```

If wrapping the server startup isn't feasible, at minimum document the race window and keep `find_free_port` as a best-effort heuristic.

### 2. Binds only to `127.0.0.1` — misses `0.0.0.0` and IPv6 (Medium)

The check binds to `127.0.0.1`, but in headless mode the server binds to `0.0.0.0`. A port can be free on localhost but occupied on `0.0.0.0` (or vice versa). The check should use the same `host` value that the server will bind to:

```python
s.bind((host, requested_port))  # not hardcoded "127.0.0.1"
```

### 3. `find_free_port` scans from 8000, not from the requested port (Low)

If the user passed `--port 8888` and it's busy, the fallback scans from 8000. It could land on port 8000, which is surprising — the user would expect 8889. Start scanning from `requested_port + 1`:

```python
fallback = find_free_port(start=requested_port + 1, end=requested_port + 100)
```

### 4. Missing type hints on `find_free_port` (Nit)

```python
def find_free_port(start=8000, end=9000):  # should be start: int = 8000, end: int = 9000) -> int:
```

Codebase uses type hints consistently — this should match.

---

Issue #1 is the main concern. The pre-check approach is inherently racy on a multi-process system. Catching the bind error from the actual server startup would be the robust solution.

DevRohit06 (approved):
Approving — the core `_serve()` retry-on-bind approach is solid and correctly addresses the TOCTOU race condition. Nice improvement over the pre-check approach.

## Post-merge cleanup needed

**1. Dead code — no-op conditional**
```python
if is_starting_server:
    effective_port = args.port
else:
    effective_port = args.port
```
Both branches are identical. Just use `args.port` directly.

**2. `SO_REUSEADDR` on probe socket**
Makes the probe less reliable (can bind even when port is in TIME_WAIT). Consider dropping it or documenting the rationale.

**3. String concatenation → f-strings**
```python
# current
"[WARN] Port " + str(current_port) + " busy — trying " + str(next_port)
# should be
f"[WARN] Port {current_port} busy — trying {next_port}"
```

**4. Unrelated formatting churn**
Section comments (`# ── Global flags ──`, etc.), multi-line arg collapses, and `_resolve_subargs` docstring were removed without functional reason. Restore in a follow-up.

**5. Squash on merge** — 14 commits including a revert and `:wq` in a message.

Pragatisource08:
@DevRohit06 
Thanks for the review and approval.
Noted on the cleanup points and the `SO_REUSEADDR` behavior, will keep these in mind for future PRs.
I’ll follow up with a small cleanup PR to address the non-functional items (dead code, f-strings, and restoring the removed comments/docstrings). For `SO_REUSEADDR`, I’ll take a closer look and either document the rationale or adjust it accordingly.

Thanks again!

PR #820, opened 2026-03-31
[FI] : Pin pytest-asyncio below 1.0.0 to fix async test failures

### LINKED ISSUE #819: All async tests fail with RuntimeError: Runner.run() cannot be called from a running event loop
### Operating System

Windows 11

### Python Version

3.12.13

### PocketPaw Version

0.4.15

### Installation Method

uv sync (from source)

### Bug Description

All async tests fail with RuntimeError: Runner.run() cannot be called from a running event loop. This is caused by pytest-asyncio 1.3.0 being installed, which is incompatible with the existing test setup. The current constraint >=0.23.0 in pyproject.toml allows 1.x to be installed, but version 1.0.0+ changed how event loops are managed, breaking all 1348 async tests.

### Steps to Reproduce

1. Clone the repo: git clone https://github.com/pocketpaw/pocketpaw.git
2. Install dependencies: uv sync --dev
3. Run the tests: uv run pytest -v
4. All async tests fail with: RuntimeError: Runner.run() cannot be called from a running event loop

### Expected Behavior

All the async test should have passed without any runtime error related to the event loops

### Actual Behavior

1348 async tests fail immediately with:
RuntimeError: Runner.run() cannot be called from a running event loop

This affects every async test across the entire test suite including tests for agents, memory, tools, adapters, browser, and more

Image

Image

Image

### Relevant Files

pyproject.toml

### Before submitting

Kishor7624:
## Problem
All 1348 async tests fail with:
RuntimeError: Runner.run() cannot be called from a running event loop

Closes #819

## Root Cause
pytest-asyncio 1.0.0+ changed how it manages event loops, breaking compatibility with the existing test setup. The constraint `>=0.23.0` allows 1.x to be installed.

## Fix
Pin pytest-asyncio to `>=0.23.0,<1.0.0` in both:
- `[project.optional-dependencies]` dev section
- `[dependency-groups]` dev section

## Testing
Verified locally by running `uv run pytest -v` — the failure is reproduced with 1.3.0 installed.


DevRohit06:
## Code Review

**Verdict: Changes Are Incomplete — The Pin Does Not Take Effect**

---

### Critical: `uv.lock` Was Not Regenerated

The pyproject.toml pin was added correctly in both locations, but `uv.lock` was never updated to reflect the new constraint. As a result, CI and anyone running `uv sync --dev` will still install **pytest-asyncio 1.3.0** — the exact version this PR intends to block.

Evidence from the PR branch lock file ([`uv.lock` @ c88b39e](https://github.com/pocketpaw/pocketpaw/blob/c88b39ec9e1762b4862c47cfb555b1fd1e78e24c/uv.lock)):

```
name = "pytest-asyncio"
version = "1.3.0"            # ← still 1.3.0
...
{ name = "pytest-asyncio", specifier = ">=0.23.0" }   # ← <1.0.0 cap missing
```

The PR only changes `pyproject.toml` (1 file). The lock file must be regenerated after changing version constraints; without it the fix is inert.

**Required action:** Run `uv lock` after the pyproject changes and commit the updated `uv.lock`.

---

### The Root Cause Diagnosis Is Partially Incorrect

The PR body states the error is `RuntimeError: Runner.run() cannot be called from a running event loop`. This is a real error but its attribution to pytest-asyncio's event-loop management change is imprecise.

**What pytest-asyncio 1.0.0 actually removed** (from the [official changelog](https://github.com/pytest-dev/pytest-asyncio/releases/tag/v1.0.0)):

- The deprecated `event_loop` fixture was removed entirely.
- Scoped event loops are now created once per scope rather than once per test using that scope.

Neither of these produces `Runner.run() cannot be called from a running event loop` on its own. That error typically originates from code inside the project that calls `asyncio.run()` or creates an `asyncio.Runner` within an already-running loop — behaviour that became newly observable when pytest-asyncio 1.x changed how and when event loops are created. Pinning is the right short-term mitigation, but the underlying call site (likely something in `agents/` or `bus/` that calls `asyncio.run()` during test setup) should be investigated and fixed so the codebase can eventually migrate to pytest-asyncio 1.x.

---

### Minor: Duplicate Pin Location Is Correct, But Worth a Comment

The project has two dev dependency sections:

- `[project.optional-dependencies] dev` — used by `pip install pocketpaw[dev]`
- `[dependency-groups] dev` — used by `uv sync --dev`

Pinning both is the right call. It may be worth adding a short inline comment (`# pinned: incompatible with pytest-asyncio 1.x until asyncio.run() call sites are migrated`) to explain *why* the upper bound exists, so the next person who hits a "dependency too old" warning doesn't just delete the cap without reading the issue.

---

### Summary of Issues

| # | Severity | Finding |
|---|----------|---------|
| 1 | **Blocker** | `uv.lock` not regenerated — pin has zero effect; 1.3.0 is still resolved and installed |
| 2 | Medium | Root-cause statement is imprecise; the actual `asyncio.run()` call site in production code should be identified and fixed for a durable solution |
| 3 | Low | No explanatory comment on the upper-bound pin — future maintainers will not know why it exists |

---

The fix is one command away: run `uv lock` and include the updated lock file in this PR.

Kishor7624:
Hey @DevRohit06, thanks for the review — really helpful feedback!

I've fixed both issues:

1. Ran `uv lock` and committed the updated lock file - pytest-asyncio now resolves to 0.26.0 instead of 1.3.0, so the pin actually takes effect
2. Added an inline comment on both pin locations explaining why the upper bound exists so the next person doesn't just delete it

You make a good point about the root cause being more precise - I oversimplified it in the PR description. The pin is really just a short-term fix to unblock the test suite while the actual asyncio.run() call sites get properly migrated.

DevRohit06 (approved):
Lock file updated, pin takes effect, explanatory comments added. LGTM.

PR #822, opened 2026-03-31
[FI] fix: use AsyncMock for neonize preflight test to avoid real network call

### LINKED ISSUE #821: Bug: test_preflight_check_passes_when_reachable makes real network call instead of mocking
### Operating System

Windows 11

### Python Version

Python 3.11.9

### PocketPaw Version

0.4.15

### Installation Method

uv sync (from source)

### Bug Description

In tests/test_neonize_adapter.py, the test test_preflight_check_passes_when_reachable 
uses return_value=None instead of AsyncMock for patching _tcp_probe. 
Since _tcp_probe is an async function, this causes the test to make a real 
network call to web.whatsapp.com:443, failing in offline/CI environments.

### Steps to Reproduce

1. Disconnect from internet
2. Run: uv run pytest tests/test_neonize_adapter.py::test_preflight_check_passes_when_reachable
3. Test fails with ConnectionError

### Expected Behavior

Test should pass without any network access using AsyncMock

### Actual Behavior

ConnectionError: Cannot reach web.whatsapp.com:443

Image

### Relevant Files

tests/test_neonize_adapter.py

### Before submitting

PrinceSharma402:
## What does this PR do?
Fixes a test that was making real network calls to web.whatsapp.com:443 during testing.

## Related Issue
Fixes #821

## Changes Made
- `tests/test_neonize_adapter.py`: Changed `return_value=None` to `new_callable=AsyncMock` for patching `_tcp_probe` since it is an async function

## How to Test
1. Run `uv run pytest tests/test_neonize_adapter.py::test_preflight_check_passes_when_reachable -v`
2. Test should pass without any network access

## Evidence of Testing
Before: FAILED — ConnectionError: Cannot reach web.whatsapp.com:443
After:  PASSED in 0.25s — no network call made

## Checklist
image

image

### DIFF
--- a/tests/test_neonize_adapter.py
+++ b/tests/test_neonize_adapter.py
@@ -294,6 +294,7 @@ async def test_bus_integration(bus):
 
 async def test_preflight_check_passes_when_reachable():
     """Preflight check succeeds when the host is reachable."""
-    with patch("pocketpaw.bus.adapters.neonize_adapter._tcp_probe", return_value=None):
+    with patch("pocketpaw.bus.adapters.neonize_adapter._tcp_probe", new_callable=AsyncMock):
+
         # Should not raise
         await NeonizeAdapter._preflight_connectivity_check()


DevRohit06 (commented):
## Code Review — First Review

**Commit reviewed:** 950952410b26984ea964eb255329a07671f03c1f

---

### Summary

The fix is correct and necessary. `_tcp_probe` is a synchronous function that is dispatched via `loop.run_in_executor(None, _tcp_probe, ...)` inside `_preflight_connectivity_check`. Patching it with the old `return_value=None` made `patch` produce a plain `MagicMock`, which is not awaitable. When `run_in_executor` executed the mock it would return `None` synchronously — that happened to pass in some environments but was silently wrong. On environments where the executor's future is correctly awaited, a plain non-coroutine mock can allow the real socket call to slip through.

Switching to `new_callable=AsyncMock` ensures the patched object is an `AsyncMock` instance, which is what the `await asyncio.wait_for(...)` machinery in `_preflight_connectivity_check` expects when the patch intercepts the executor path.

---

### Issues Scored ≥ 80

#### [Score: 90] `AsyncMock` is the right tool but for the wrong reason — the patch target is a sync function

`_tcp_probe` itself is **not** an `async def`; it is a plain synchronous function:

```python
def _tcp_probe(host: str, port: int, timeout: float) -> None:
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    ...
```

It is called via `loop.run_in_executor(None, _tcp_probe, host, port, timeout)`. The async part is `run_in_executor` returning an awaitable — `_tcp_probe` itself is never awaited directly.

With `new_callable=AsyncMock` you are patching a sync function with an async mock. When `run_in_executor` is called with the `AsyncMock` object as its callable, the executor calls it in a thread pool and gets back a coroutine object rather than `None`. A coroutine object that is never awaited will then trigger a `RuntimeWarning: coroutine '...' was never awaited` in the test output and may not actually represent a clean pass.

**The correct fix** for a sync function called inside `run_in_executor` is to also patch `run_in_executor` so that it short-circuits the executor entirely, or to use `MagicMock(return_value=None)` and patch `asyncio.get_running_loop().run_in_executor` as well. For example:

```python
async def test_preflight_check_passes_when_reachable():
    """Preflight check succeeds when the host is reachable."""
    with patch("pocketpaw.bus.adapters.neonize_adapter._tcp_probe") as mock_probe:
        mock_probe.return_value = None  # sync; executor calls it, gets None
        # Should not raise
        await NeonizeAdapter._preflight_connectivity_check()
```

This is exactly what the original code tried to do. The reason the original code made a real network call was almost certainly because the `patch` target path was not being resolved before `_tcp_probe` was imported into the module's namespace (i.e., the module had already bound `_tcp_probe` locally by the time `patch` ran). The fix should verify the patch path is correct (e.g., confirm `pocketpaw.bus.adapters.neonize_adapter._tcp_probe` is the right string) rather than switching to `AsyncMock`.

**If** the test now passes with `AsyncMock` without a `RuntimeWarning`, it is because the executor happened to swallow the coroutine object silently. That is fragile behavior that may break on a different Python/asyncio version.

---

#### [Score: 80] Missing blank line introduced by the diff violates `ruff` format rules

The diff adds a blank line after the `with patch(...)` statement:

```python
    with patch("pocketpaw.bus.adapters.neonize_adapter._tcp_probe", new_callable=AsyncMock):

        # Should not raise
        await NeonizeAdapter._preflight_connectivity_check()
```

A blank line between the `with` header and its body is not idiomatic Python and will be flagged/reformatted by `ruff format`. Run `uv run ruff format tests/test_neonize_adapter.py` and commit the result.

---

### Minor / Not Scored

- The PR fixes a real problem (#821 — actual network calls in tests). The intent is solid.
- `asyncio_mode = "auto"` is set in `pyproject.toml`, so the test function itself does not need `@pytest.mark.asyncio` — that part is fine.
- There is no companion test for the failure path (`test_preflight_check_raises_when_unreachable`). Adding one would give meaningful coverage of the `ConnectionError` branch.

---

### Verdict

The fix addresses a real flakiness/leakage bug but the chosen mechanism (`AsyncMock` for a sync probe function) is subtly incorrect. The blank line added by the patch also needs to be cleaned up. The PR needs a small revision before merge.

DevRohit06 (changes requested):
AsyncMock wrong for sync _tcp_probe function — use MagicMock(return_value=None). See comment.

prakashUXtech (changes requested):
The intent is right — the original patch was leaking real network calls into the test suite, and that needs to be fixed. But `AsyncMock` is the wrong tool here.

`_tcp_probe` is a synchronous function dispatched via `loop.run_in_executor`. The executor calls it in a thread pool and expects a plain return value, not a coroutine. Patching it with `AsyncMock` means the executor receives a coroutine object, never awaits it, and you'll get `RuntimeWarning: coroutine was never awaited` — the test appears to pass but is silently broken.

The correct fix is `MagicMock(return_value=None)`, which is what the original code tried. The original likely failed because the import binding was already resolved before `patch` ran. Check that the patch target path is the full module path where `_tcp_probe` is defined (i.e., `pocketpaw.bus.adapters.neonize_adapter._tcp_probe`), not where it might be imported from elsewhere.

```python
async def test_preflight_check_passes_when_reachable():
    """Preflight check succeeds when the host is reachable."""
    with patch("pocketpaw.bus.adapters.neonize_adapter._tcp_probe", return_value=None):
        await NeonizeAdapter._preflight_connectivity_check()
```

Also, the blank line between the `with` header and its body (`with patch(...):\n\n    await ...`) should be removed — ruff will flag it.

DevRohit06:

PR #829, opened 2026-03-31
[FI] fix(installer): --backend flag missing openai_agents, google_adk and other backends

Adarsh-Kumar6534:
## Problem

The `--backend` CLI argument in `installer/installer.py` only accepts 3 choices:
```python
choices=["claude_agent_sdk", "pocketpaw_native", "open_interpreter"]
```

But PocketPaw supports 6 backends (all defined in the `BACKENDS` dict in the same file):
- `openai_agents`
- - `google_adk`
- - `codex_cli`
- - `opencode`
- - `copilot_sdk`
Anyone doing a headless/CI install with `--backend openai_agents` gets:
```
error: argument --backend: invalid choice: 'openai_agents'
```

## Fix

Derive the choices directly from the `BACKENDS` dict:
```python
choices=list(BACKENDS.keys()),
```

This also future-proofs the flag - any new backend added to `BACKENDS` is automatically valid without needing a second change here.

## Testing
```bash
python installer/installer.py --non-interactive --backend openai_agents --no-launch
# Before fix: argparse error
# After fix: proceeds normally
```


DevRohit06:
## Code Review

The fix is correct and the motivation is right — the hardcoded choices list was genuinely stale and the dynamic derivation is a better pattern. Approving the direction, but there are three issues that should be resolved before merging.

---

### Bug (blocking): `deep_agents` still missing from `BACKENDS`

The `BACKENDS` dict in `installer/installer.py` lists six backends. The runtime registry in [`src/pocketpaw/agents/registry.py`](https://github.com/pocketpaw/pocketpaw/blob/71e21a77f3f3751bf9dceda6f3dec305a192e1e5/src/pocketpaw/agents/registry.py#L19-L27) registers seven — `deep_agents` (LangChain/LangGraph) is present in the registry but absent from `BACKENDS`.

Because this PR derives choices from `BACKENDS.keys()`, `--backend deep_agents` will still produce an argparse error after the fix. The PR description lists six backends without mentioning `deep_agents` at all, suggesting this was missed rather than intentional.

**Fix:** add `deep_agents` to `BACKENDS`:
```python
BACKENDS = {
    "claude_agent_sdk": "Claude Agent SDK (recommended)",
    "openai_agents": "OpenAI Agents SDK",
    "google_adk": "Google ADK (Gemini)",
    "codex_cli": "Codex CLI (OpenAI)",
    "opencode": "OpenCode (External Server)",
    "copilot_sdk": "Copilot SDK",
    "deep_agents": "Deep Agents (LangChain/LangGraph)",  # add this
}
```

---

### Bug (blocking): `--llm-provider` has the same class of bug, unfixed

At line 1267 (unchanged by this PR):
```python
choices=["anthropic", "openai", "ollama", "auto"],
```

But `LLM_PROVIDERS` in the same file defines two additional values: `gemini` and `openai_compatible`. Running `--llm-provider gemini` or `--llm-provider openai_compatible` in a headless install fails with an argparse error — the exact same bug pattern as the one this PR is fixing.

The fix is identical in form: `choices=list(LLM_PROVIDERS.keys())`.

---

### Dead code (non-blocking): stale `pocketpaw_native` guard at line 1072

```python
if self.backend == "pocketpaw_native" and "native" not in self.extras:
    self.extras.append("native")
```

`pocketpaw_native` is not in `BACKENDS` (so it cannot be selected interactively or via `--backend` after this fix) and is a legacy alias in the runtime registry that silently redirects to `claude_agent_sdk`. This branch can never execute. It should be removed to avoid confusion — a future reader might think `pocketpaw_native` is still a valid selection.

---

### Summary

| Issue | Severity | Line |
|---|---|---|
| `deep_agents` absent from `BACKENDS`, still blocked by argparse | Blocking | L223–230 |
| `--llm-provider` choices hardcoded, `gemini`/`openai_compatible` fail | Blocking | L1267 |
| `pocketpaw_native` guard is dead code | Non-blocking | L1072 |

The one-line fix in this PR is sound — please expand it to cover the two blocking issues above before merging.

Adarsh-Kumar6534:
Thanks for the thorough review! I've addressed all three points — added deep_agents to BACKENDS, applied the same dynamic derivation fix to --llm-provider, and removed the stale pocketpaw_native guard. Ready for re-review.

DevRohit06 (approved):
All three issues fixed — deep_agents added, --llm-provider dynamic, dead code removed.

PR #832, opened 2026-03-31
[FI] fix: expose get_by_type on MemoryManager to remove private _store access

### LINKED ISSUE #831: Bug: MemoryManager._store accessed directly in dashboard.py — missing public get_by_type method
### Operating System

Windows 11

### Python Version

3.12.11

### PocketPaw Version

0.4.15

### Installation Method

uv sync (from source)

### Bug Description

`dashboard.py` line 1372 accesses the private `_store` attribute of 
`MemoryManager` directly to call `get_by_type()`:

    items = await manager._store.get_by_type(MemoryType.LONG_TERM, limit=limit)

This was noted by the maintainer with a `# TODO` comment. `MemoryManager` 
should expose a public `get_by_type()` method so callers don't need to 
reach into private internals.

### Steps to Reproduce

Search dashboard.py for manager._store.get_by_type — line 1372

### Expected Behavior

MemoryManager exposes a public get_by_type() method

### Actual Behavior

dashboard.py accesses manager._store directly, bypassing encapsulation

# From dashboard.py line 1368-1372:
# Access store directly for filtered query, or use get_by_type if exposed
# Manager doesn't expose get_by_type publicly in facade
# For now, let's use _store hack or add method to manager?
items = await manager._store.get_by_type(MemoryType.LONG_TERM, limit=limit)

### Relevant Files

src/pocketpaw/memory/manager.py
src/pocketpaw/dashboard.py

### Before submitting

Karan20P:
## What does this PR do?
Adds a public `get_by_type` method to `MemoryManager` so callers don't
need to access the private `_store` attribute directly. Updates dashboard.py
to use the new public method instead of the `_store` hack noted in the TODO comment.

## Related Issue
Fixes #831

## Changes Made
- `src/pocketpaw/memory/manager.py`: Added public `get_by_type()` method that delegates to `self._store.get_by_type()`
- `src/pocketpaw/dashboard.py`: Updated to use `manager.get_by_type()` instead of `manager._store.get_by_type()`

## How to Test
1. Run: `uv run pytest tests/ -k "memory" -v`
2. Confirm 192 tests pass
3. Run: `uv run ruff check src/pocketpaw/memory/manager.py src/pocketpaw/dashboard.py`
4. Confirm no linting errors

## Evidence of Testing
1 failed, 192 passed, 1 skipped (pre-existing Windows permission error, unrelated to this change)
All ruff checks passed.

## Checklist

### DIFF
--- a/src/pocketpaw/api/v1/memory.py
+++ b/src/pocketpaw/api/v1/memory.py
@@ -34,7 +34,7 @@ async def get_long_term_memory(limit: int = Query(50, ge=1, le=500)):
     from pocketpaw.memory import MemoryType, get_memory_manager
 
     manager = get_memory_manager()
-    items = await manager._store.get_by_type(MemoryType.LONG_TERM, limit=limit)
+    items = await manager.get_by_type(MemoryType.LONG_TERM, limit=limit)
     return [
         MemoryEntry(
             id=item.id,
--- a/src/pocketpaw/memory/manager.py
+++ b/src/pocketpaw/memory/manager.py
@@ -635,6 +635,8 @@ class MemoryManager:
 
         # Fall back to standard context
         return await self.get_context_for_agent(sender_id=sender_id)
+    
+    
     async def get_by_type(
         self,
         memory_type: MemoryType,
--- a/src/pocketpaw/memory/protocol.py
+++ b/src/pocketpaw/memory/protocol.py
@@ -64,6 +64,7 @@ class MemoryStoreProtocol(Protocol):
         self,
         memory_type: MemoryType,
         limit: int = 100,
+        user_id: str | None = None,
     ) -> list[MemoryEntry]:
         """Get all memories of a specific type."""
         ...
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -553,3 +553,4 @@ async def test_get_by_type_forwards_user_id(tmp_path):
     manager._store.get_by_type.assert_awaited_once_with(
         MemoryType.LONG_TERM, limit=5, user_id="user123"
     )
+    


DevRohit06:
## Code Review — First Review

**Commit:** a8ccd3d6747821451e35c35d70882978339049d7

The direction is right — removing `_store` leakage from call sites is a genuine improvement. A few issues need addressing before merge.

---

### [BLOCKER] Incomplete fix — `api/v1/memory.py` still calls `_store` directly

The PR description says it removes private `_store` access, and `dashboard.py` is updated, but a parallel call site was missed:

**`src/pocketpaw/api/v1/memory.py:37`**
```python
items = await manager._store.get_by_type(MemoryType.LONG_TERM, limit=limit)
```

This is the exact same pattern the PR is fixing. The new `manager.get_by_type()` should be used here too. A search for `_store.get_by_type` in the codebase returns two results — the PR only addressed one.

---

### [HIGH] Protocol/type contract mismatch with `user_id` parameter

The `MemoryStoreProtocol` definition (`src/pocketpaw/memory/protocol.py:63`) does not include `user_id`:

```python
async def get_by_type(
    self,
    memory_type: MemoryType,
    limit: int = 100,
) -> list[MemoryEntry]:
```

Both `FileMemoryStore` and `Mem0Store` accept `user_id` only via `**kwargs`, which is a runtime convention, not a protocol contract. The new public method introduces a typed `user_id: str | None = None` parameter that it passes as a keyword arg to `self._store`. Any future store that implements `MemoryStoreProtocol` strictly (without `**kwargs`) will raise a `TypeError` at runtime when `user_id` is non-None, and mypy won't catch it because the protocol doesn't declare the parameter.

Two cleaner paths:
1. Add `user_id: str | None = None` to `MemoryStoreProtocol.get_by_type` to make the contract explicit (preferred — also updates both implementations to drop `**kwargs` in favour of the named param).
2. Keep `user_id` off the manager method for now (matching the protocol), since the only current call site in `dashboard.py` doesn't pass `user_id`.

---

### [MINOR] Missing blank lines before new method — ruff will reject this

The diff inserts `get_by_type` with no blank line separating it from the preceding `get_semantic_context` return statement. PEP 8 / ruff rule `E302` requires two blank lines between top-level async methods in a class. The PR description says ruff passes, but the diff clearly shows trailing content with only one blank line before `clear_session` as well. Please re-run `uv run ruff check src/pocketpaw/memory/manager.py` on the actual committed file.

---

### [MINOR] Tests mock the private attribute they're supposed to hide

```python
manager._store.get_by_type = AsyncMock(return_value=[])
```

The tests correctly verify delegation behaviour, but they patch `_store` directly — the attribute the PR argues callers should not touch. This is fine for unit tests (you need to mock the underlying store somehow), but consider using `unittest.mock.patch.object` on the store instance returned by `_create_store()` instead of monkey-patching `_store` after construction. It's a minor style note, not a blocker.

---

### [MINOR] Missing newline at end of file in `tests/test_memory.py`

The diff ends with `\ No newline at end of file` on the last line of the new test. Ruff/pre-commit will flag this (`W292`).

---

### Summary

| Severity | Issue |
|---|---|
| **Blocker** | `api/v1/memory.py:37` still calls `_store.get_by_type` directly |
| **High** | `user_id` param not in `MemoryStoreProtocol` — runtime contract not enforced |
| **Minor** | Blank line formatting (ruff E302) |
| **Minor** | Tests patch `_store` directly; missing newline at EOF (ruff W292) |

The core change (exposing the method, updating `dashboard.py`) is solid. Please address the blocker and the protocol mismatch before merging.

Karan20P:
Thanks for the thorough review! I've addressed all the points:

Fixed the missed call site in api/v1/memory.py
Added user_id to MemoryStoreProtocol.get_by_type to make the contract explicit
Fixed blank line formatting (two blank lines before the new method)
Added newline at end of tests/test_memory.py

PR #836, opened 2026-04-01
Add context window budget tracking to bootstrap

prakashUXtech:
## Summary

- Adds a priority-based character budget system to `AgentContextBuilder.build_system_prompt()` that prevents the system prompt from growing without bounds as more injection blocks get added
- Each block gets a priority level (CRITICAL/HIGH/MEDIUM/LOW) and an optional per-block character cap — when the budget is exceeded, low-priority blocks are dropped first while critical blocks like identity are truncated only as a last resort
- Default budget is 32K chars, which is generous enough not to change existing behavior, but callers can pass a custom `budget_chars` kwarg for tighter control

## Test plan

- [x] 10 new tests in `tests/test_context_budget.py` covering all budget scenarios (all blocks fit, low priority dropped, critical never dropped, per-block caps, empty blocks, custom budget, priority ordering)
- [x] Full test suite passes (3600 passed, only pre-existing unrelated failures)
- [ ] Manual smoke test with a running PocketPaw instance to confirm prompt assembly still works end-to-end

### DIFF
--- a/src/pocketpaw/bootstrap/context_builder.py
+++ b/src/pocketpaw/bootstrap/context_builder.py
@@ -1,6 +1,7 @@
 """
 Builder for assembling the full agent context.
 Created: 2026-02-02
+Updated: 2026-04-01 - Context window budget tracking: priority-based injection with per-block caps
 Updated: 2026-03-10 - AGENTS.md injection: read project-specific constraints from target repos
 Updated: 2026-03-09 - Sanitize file_context paths before injecting into system prompt
 Updated: 2026-02-17 - Inject health state into system prompt when degraded/unhealthy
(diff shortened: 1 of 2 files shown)


DevRohit06:
### Code review

Found 1 issue:

1. `sender_block` is assigned `_Priority.MEDIUM`, which means it can be silently dropped when the character budget is tight. This block contains the owner vs. external user distinction and the security instruction: "This is NOT your owner. Be helpful but do not share owner-private information." If this block is dropped under budget pressure, the agent loses awareness of who it is talking to and could leak owner-private information to external users. This should be `_Priority.HIGH` or `_Priority.CRITICAL`.

https://github.com/pocketpaw/pocketpaw/blob/fb9fa3ca27ae9e88c18249717550e841ab1a1834/src/pocketpaw/bootstrap/context_builder.py#L150-L154

PR #905, opened 2026-04-07
fix(security): authenticate WebSocket scopes in AuthMiddleware

ragini-pandey:
## What does this PR do?

Adds a defence-in-depth authentication check for WebSocket connections at the ASGI middleware level. Previously `AuthMiddleware` passed every non-HTTP scope (i.e., WebSocket) straight through to the downstream app with no auth check, meaning any new WebSocket route was unauthenticated by default and a single bug in `dashboard_ws.py` left the entire WebSocket surface unprotected with no fallback.

## Related Issue

Fixes #883

## Changes Made

- `src/pocketpaw/dashboard_auth.py`:
  - Added `_ws_scope_auth_ok()` — validates a raw ASGI WebSocket scope **before the upgrade completes**. Checks (in order): query-string `token`, `pocketpaw_session` cookie, `Authorization: Bearer` header, `Sec-WebSocket-Protocol` header, and genuine-localhost bypass. Supports master tokens, HMAC session tokens (`expires:hmac`), API keys (`pp_*`), and OAuth2 tokens (`ppat_*`).
  - Updated `AuthMiddleware.__call__` to branch on `scope["type"] == "websocket"` and call `_ws_scope_auth_ok()`. Unauthenticated connections are closed with code `4003` before the upgrade completes. `lifespan` scopes still pass through unconditionally.
  - Removed `/ws`, `/v1/ws`, `/api/v1/ws` entries from the `_auth_dispatch` exempt list — WebSocket scopes no longer reach `_auth_dispatch`.
- `tests/test_dashboard_security.py`:
  - Added `TestWebSocketMiddlewareAuth` with 11 tests.

## How to Test

1. Install dependencies: `uv sync --dev`
2. Run the new test class directly:
   ```
   uv run pytest tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth -v
   ```
3. Run the full security test file to check for regressions:
   ```
   uv run pytest tests/test_dashboard_security.py -v
   ```
4. Run all non-e2e tests:
   ```
   uv run pytest --ignore=tests/e2e
   ```

## Evidence of Testing

```
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_ws_no_token_rejected PASSED [ 78%]
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_ws_valid_query_token PASSED [ 80%]
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_ws_invalid_query_token PASSED [ 82%]
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_ws_valid_cookie PASSED [ 85%]
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_ws_valid_bearer_header PASSED [ 87%]
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_ws_valid_session_token PASSED [ 89%]
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_ws_localhost_bypass PASSED [ 91%]
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_ws_sec_websocket_protocol PASSED [ 93%]
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_middleware_closes_unauthenticated_ws PASSED [ 95%]
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_middleware_passes_authenticated_ws PASSED [ 97%]
tests/test_dashboard_security.py::TestWebSocketMiddlewareAuth::test_middleware_passes_lifespan_through PASSED [100%]

============================== 47 passed in 1.23s ==============================

$ uv run ruff check .
All checks passed!
```

## Checklist

### DIFF
--- a/src/pocketpaw/dashboard_auth.py
+++ b/src/pocketpaw/dashboard_auth.py
@@ -131,24 +131,132 @@ async def verify_token(
 
 
 # ---------------------------------------------------------------------------
-# HTTP auth middleware (registered by dashboard.py via app.add_middleware)
+# WebSocket scope authentication (used by AuthMiddleware — issue #883)
+# ---------------------------------------------------------------------------
+
+
+def _ws_scope_auth_ok(scope: dict) -> bool:
+    """Return True if the raw ASGI WebSocket *scope* carries valid auth.
+
+    Checks — in order — query-string ``token`` param, ``Cookie`` header
+    (``pocketpaw_session``), ``Authorization: Bearer …`` header,
+    ``Sec-WebSocket-Protocol`` header, and genuine-localhost bypass.
+
+    This runs *before* the WebSocket upgrade completes so that
+    unauthenticated connections are rejected immediately.
+    """
+    from urllib.parse import parse_qs
+
+    current_token = get_access_token()
+
+    # --- helpers -----------------------------------------------------------
+
+    def _tok_ok(t: str | None) -> bool:
+        if not t:
+            return False
+        if hmac.compare_digest(t, current_token):
+            return True
+        if ":" in t and verify_session_token(t, current_token):
+            return True
+        if t.startswith("pp_") and not t.startswith("ppat_"):
+            try:
+                from pocketpaw.api.api_keys import get_api_key_manager
+
+                if get_api_key_manager().verify(t) is not None:
+                    return True
+            except Exception:
+                pass
+        if t.startswith("ppat_"):
+            try:
+                from pocketpaw.api.oauth2.server import get_oauth_server
+
+                if get_oauth_server().verify_access_token(t) is not None:
+                    return True
+            except Exception:
+                pass
+        return False
+
+    # --- extract headers as a dict (lower-cased keys) ----------------------
+    headers: dict[str, str] = {}
+    for raw_name, raw_val in scope.get("headers", []):
+        headers[raw_name.decode("latin-1").lower()] = raw_val.decode("latin-1")
+
+    # 1. Query-string token
+    qs = scope.get("query_string", b"").decode("latin-1")
+    params = parse_qs(qs)
+    token_vals = params.get("token", [])
+    if token_vals and _tok_ok(token_vals[0]):
+        return True
+
+    # 2. HTTP-only session cookie
+    cookie_header = headers.get("cookie", "")
+    for morsel in cookie_header.split(";"):
+        morsel = morsel.strip()
+        if morsel.startswith("pocketpaw_session="):
+            cookie_token = morsel.split("=", 1)[1]
+            if _tok_ok(cookie_token):
+                return True
+
+    # 3. Authorization header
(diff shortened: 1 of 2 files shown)


DevRohit06 (approved):
## PR #905 Review — WebSocket auth in AuthMiddleware (defence-in-depth)

This is a well-executed security fix. The design is sound: authenticate WebSocket scopes at the ASGI middleware level before the upgrade completes, so any new WS route is protected by default.

### What's done well

- **Correct ASGI flow**: Consuming `websocket.connect` via `await receive()` then sending `websocket.close` with code `4003` is the right way to reject before upgrade.
- **Comprehensive token validation**: Checks query-string token, cookie, Bearer header, `Sec-WebSocket-Protocol`, and localhost bypass — matching all existing auth methods.
- **`hmac.compare_digest`** for master token comparison — prevents timing attacks.
- **`_ScopeProxy`** for `_is_genuine_localhost` is a clean adapter. Verified that `_PROXY_HEADERS` uses lowercase keys (`"cf-connecting-ip"`, `"x-forwarded-for"`), so the lowered header dict works correctly.
- **Exempt list cleanup**: Removing `/ws`, `/v1/ws`, `/api/v1/ws` from `_auth_dispatch` is correct since WS scopes never reach HTTP dispatch.
- **Excellent test coverage**: 11 tests covering all auth methods, rejection behavior, pass-through, and lifespan scoping.
- **Lifespan pass-through** preserved (handled by `scope["type"] != "http"` fallthrough).

### Minor notes (non-blocking)

**1. `_ScopeProxy` inline class creation**

```python
self.client = type("C", (), {"host": s.get("client", ("", 0))[0]})()
```

This works but is a bit clever. A `namedtuple` or simple dataclass would be more readable:
```python
from collections import namedtuple
_Client = namedtuple("Client", ["host"])
self.client = _Client(host=s.get("client", ("", 0))[0])
```

Not blocking — the current approach is functionally correct.

**2. Consider: what if `scope["client"]` is `None`?**

ASGI spec says `client` can be `None` (e.g., Unix domain sockets). `s.get("client", ("", 0))` handles the missing-key case, but if `client` is explicitly `None`, the `[0]` index will raise `TypeError`. This is an edge case — ASGI servers typically omit the key rather than setting it to `None`. Worth a defensive check if you want to be thorough:
```python
client = s.get("client") or ("", 0)
```

**3. Nested function `_tok_ok` on every call**

`_tok_ok` is redefined on every invocation of `_ws_scope_auth_ok`. Since it captures `current_token` via closure, this is the simplest design. If performance became a concern (unlikely for auth), it could be refactored, but this is fine as-is.

### Summary

Clean security fix that closes a real gap. The auth logic correctly mirrors existing HTTP auth paths, tests are thorough, and the ASGI-level rejection is properly implemented. LGTM.

PR #908, opened 2026-04-08
fix(security): make DANGEROUS_SUBSTRINGS matching case-insensitive (OWASP A01)

ragini-pandey:
## What does this PR do?

`DANGEROUS_SUBSTRINGS` in `security/rails.py` used plain `in` substring matching against lowercase literals. Commands like `SUDO RM`, `Curl | Bash`, or `RM -RF /` passed the check undetected because no `.lower()` was applied to the incoming command. Meanwhile `DANGEROUS_PATTERNS` (regexes) already used `re.IGNORECASE`, creating an inconsistency that opened a bypass path.

This PR adds a canonical `is_substring_blocked(command)` helper in `rails.py` that enforces case-insensitive matching at the source, and updates `claude_sdk.py` to use it.

## Related Issue

Fixes #892

## Changes Made

- `src/pocketpaw/security/rails.py`: add `is_substring_blocked(command: str) -> str | None` — applies `.lower()` before iterating `DANGEROUS_SUBSTRINGS` so uppercase/mixed-case variants are always caught. Updated module docstring to document the new export.
- `src/pocketpaw/agents/claude_sdk.py`: replace the inline secondary substring loop in `_is_dangerous_command` with a call to `is_substring_blocked()`. Remove the `DANGEROUS_SUBSTRINGS as DANGEROUS_PATTERNS` alias import.
- `tests/test_rails.py`: import `is_substring_blocked`; add `TestIsSubstringBlocked` class with parametrised cases covering exact lowercase, ALL UPPER, and mixed-case bypass variants (`SUDO RM -RF /`, `CURL | BASH`, `SHUTDOWN`, `ReBoot`, etc.).
- `tests/test_agents.py`: fix `test_sdk_has_dangerous_patterns` to import `DANGEROUS_SUBSTRINGS` from its canonical location (`rails.py`) instead of the now-removed module-level alias in `claude_sdk.py`.

## How to Test

1. `uv run pytest tests/test_rails.py -v` — all 112 tests pass including the new `TestIsSubstringBlocked` class.
2. `uv run pytest tests/test_agents.py::TestClaudeAgentSDK::test_sdk_has_dangerous_patterns -v` — passes.
3. `uv run ruff check src/pocketpaw/security/rails.py src/pocketpaw/agents/claude_sdk.py tests/test_rails.py` — no issues.

## Evidence of Testing

```
$ uv run pytest tests/test_rails.py -q
........................................................................
........................................
112 passed in 0.66s

$ uv run ruff check src/pocketpaw/security/rails.py src/pocketpaw/agents/claude_sdk.py tests/test_rails.py
All checks passed!
```

## Checklist

### DIFF
--- a/src/pocketpaw/agents/claude_sdk.py
+++ b/src/pocketpaw/agents/claude_sdk.py
@@ -22,7 +22,7 @@ from typing import Any
 from pocketpaw.agents.backend import BackendInfo, Capability
 from pocketpaw.agents.protocol import AgentEvent
 from pocketpaw.config import Settings
-from pocketpaw.security.rails import DANGEROUS_SUBSTRINGS as DANGEROUS_PATTERNS
+from pocketpaw.security.rails import is_substring_blocked
 from pocketpaw.tools.policy import ToolPolicy
 
 logger = logging.getLogger(__name__)
@@ -214,12 +214,10 @@ class ClaudeSDKBackend:
             if pattern.search(command):
                 return pattern.pattern
 
-        # Secondary: substring matching (catches simple literal fragments)
-        command_lower = command.lower()
-        for pattern in DANGEROUS_PATTERNS:
-            if pattern.lower() in command_lower:
-                return pattern
-        return None
+        # Secondary: substring matching (catches simple literal fragments).
+        # is_substring_blocked() applies .lower() on both sides so that
+        # uppercase variants like "SUDO RM" are caught (OWASP A01).
+        return is_substring_blocked(command)
 
     # Patterns that indicate an OS-level "open file" command.
     _FILE_OPEN_PATTERNS = [
--- a/src/pocketpaw/security/rails.py
+++ b/src/pocketpaw/security/rails.py
@@ -4,11 +4,14 @@ Shared dangerous-command patterns — single source of truth.
 Every security rail in the codebase (Guardian, ShellTool, pocketpaw_native,
 claude_sdk) imports from here.  **Do not define ad-hoc pattern lists elsewhere.**
 
-Two exports:
+Exports:
   DANGEROUS_PATTERNS           – raw regex strings (case-insensitive intent)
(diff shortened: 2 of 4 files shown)


DevRohit06 (approved):
## PR #908 Review — Case-insensitive DANGEROUS_SUBSTRINGS matching

Clean, well-structured security hardening. The canonical `is_substring_blocked()` helper is a good defensive pattern — it centralizes the case-insensitive matching so call sites can't accidentally reintroduce a case-sensitive bypass. Tests are thorough with parametrized case variants.

**Approved** with minor notes:

### Minor note: implicit lowercase coupling

`is_substring_blocked()` only lowercases the *command* (`command.lower()`), relying on `DANGEROUS_SUBSTRINGS` entries being lowercase. The previous code in `claude_sdk.py` actually did `.lower()` on *both* sides:

```python
# Old (defensive on both sides):
if pattern.lower() in command_lower:

# New helper (assumes substrings are lowercase):
if sub in command_lower:
```

This is fine today since all entries are lowercase, but adding `sub.lower()` would make it bulletproof against a future contributor accidentally adding `"RM -RF /"` to the list. Very minor — not blocking.

### Honesty check on the bug description

The PR describes this as a bypass where `SUDO RM`, `Curl | Bash` etc. "passed the check undetected." However, the existing `claude_sdk.py` code already applied `.lower()` to both sides (lines 218-220), so the bypass didn't exist in the *current* call site. The real value here is the **defensive refactor** — ensuring no *future* call site can skip the lowering. Worth clarifying in the commit message, but not a blocker.

### Everything else looks good
- `rails.py` helper is well-documented with docstring and module-level comment
- `claude_sdk.py` change is a clean simplification
- `test_agents.py` import fix is correct
- New `TestIsSubstringBlocked` tests cover uppercase, mixed-case, safe commands, and return type
- PR targets `dev`, linting passes, single focused commit

PR #916, opened 2026-04-10
feat(ee): RBAC + ABAC guards module

prakashUXtech:
## Summary

Authorization backbone for PocketPaw Enterprise. Composable FastAPI dependencies that stack like NestJS guards — attach to any route with `Depends()`.

**Three layers, one pattern:**

- **RBAC** — `WorkspaceRole` (member/admin/owner) and `PocketAccess` (view/comment/edit/owner) enums with numeric-level guard functions
- **ABAC** — `evaluate_policy()` runs a 4-check chain: plan feature gates, action-to-role mapping (16 actions), agent creator ceiling, per-role tool whitelists
- **FastAPI deps** — `require_role()`, `require_pocket_access()`, `require_plan_feature()`, `require_policy()` factories

**Usage on routes:**

```python
# Simple role check
@router.delete("/pockets/{id}", dependencies=[Depends(require_role("admin"))])

# Plan-gated feature
@router.post("/automations", dependencies=[Depends(require_plan_feature("automations"))])

# Full ABAC evaluation
@router.post("/agents/{id}/run", dependencies=[Depends(require_policy("agent.run"))])
```

**Agent permission model:** agents inherit their creator's role as a ceiling, further scoped by pocket-level access and an explicit tool whitelist (members get web_search/research/memory; admins get everything).

## Files

| File | Purpose |
|------|---------|
| `ee/__init__.py` | Package root, FSL 1.1 license, module map |
| `ee/guards/rbac.py` | WorkspaceRole, PocketAccess, Forbidden, check functions |
| `ee/guards/policy.py` | PolicyContext + PolicyResult dataclasses |
| `ee/guards/abac.py` | Plan gates, action-role mapping, tool whitelists, evaluate_policy() |
| `ee/guards/deps.py` | FastAPI dependency factories |
| `ee/guards/__init__.py` | Re-exports for clean imports |
| `tests/test_ee_guards.py` | 67 tests — enums, guards, policy eval, FastAPI deps |

## Test plan

- [x] `uv run pytest tests/test_ee_guards.py -v` — 67 passed in 0.20s


DevRohit06 (commented):
## Review — feat(ee): RBAC + ABAC guards module

Clean, well-structured authorization module. The layered design (enums → guard functions → policy evaluation → FastAPI deps) is solid. 72 tests with good coverage. A few issues to address:

---

### 🔴 Security — Default-open for unknown actions

**`abac.py:173`** — `evaluate_policy()` returns `allowed=True` for any action not in `ACTION_ROLES`. A typo like `"worksapce.delete"` silently passes all checks. Default-allow is dangerous for an authorization system.

**Suggestion:** Default-deny, or at minimum log unknown actions at `WARNING` level so they're visible in production.

---

### 🔴 Bug — `require_policy` reads `agent_id` from query params only

**`deps.py:336`** — `agent_id = request.query_params.get("agent_id")`. If the agent ID comes from a path parameter (`/agents/{agent_id}/run`) or request body, the ceiling check is silently skipped. The `agent_context` on `request.state` already has the agent_id — read from there instead:

```python
agent_ctx = getattr(request.state, "agent_context", None)
if agent_ctx:
    agent_id = agent_ctx.get("agent_id")
    creator_role_str = agent_ctx.get("creator_role")
    ...
```

This also removes the redundant `agent_id` comparison on line 340.

---

### 🟡 Design — `_feature_for_action` mapping is implicit

**`abac.py:159-170`** — Only 5 prefixes are mapped. Actions like `pocket.create` or `agent.run` bypass plan gates entirely. This is probably intentional (core features on all plans), but the implicit skip-if-unmapped behavior should be documented with a comment for future maintainers.

---

### 🟡 `require_role` workspace mismatch edge case

**`deps.py:280`** — `membership.get("workspace_id")` returns `None` if the key is missing, which still fails the `!= ws_id` check correctly. But `membership.get("role", "")` on line 332 defaults to `"member"` (not empty string) — the inconsistency in default handling between the two deps could mask bugs. Consider normalizing.

---

### 🟢 Positives

- Frozen dataclasses with `slots=True` — good for immutability and perf
- `StrEnum` + `from_str` handles case-insensitive DB values cleanly
- Plan feature table uses set superset invariant (`enterprise ⊇ business ⊇ team`), well-tested
- Dotted error codes (`workspace.insufficient_role`, `plan.feature_denied`) are consistent and machine-parsable
- FastAPI dep tests properly inject middleware state mirroring production patterns
- Agent ceiling as defense-in-depth layer is a good pattern

---

**Verdict:** Core design is solid. The default-open unknown actions and agent_id query-param-only issues should be fixed before merge. The rest can be follow-ups.

PR #925, opened 2026-04-11
Modify existing pockets instead of creating duplicates

prakashUXtech:
## Summary

When a user opens a pocket and asks the agent to change its layout, the agent now updates the existing pocket instead of creating a new one with mock data.

**Root cause:** `CreatePocketTool` always generated a new UUID and `_create_pocket_and_session()` always inserted. The agent knew which pocket was open (via `<current-pocket>` tag) but had no tool path to update it.

**Changes:**

- `CreatePocketTool` now accepts optional `pocket_id` parameter. When provided, reuses the ID and emits `"updated"` event instead of `"created"`
- New `_update_pocket()` function in agent loop handles `"updated"` events by calling `PocketService.update()` with the new rippleSpec
- System prompt updated to teach the agent: "when `<current-pocket>` is present, pass its `pocket_id` to `create_pocket` to update in place"
- Frontend counterpart in paw-enterprise#61 handles `pocket_updated` SSE events

## Test plan

- [x] 37/37 pocket tool tests passing
- [ ] Open existing pocket, ask agent to change layout — verify same pocket ID, no duplicate
- [ ] New pocket creation still works (no pocket_id = new UUID as before)
- [ ] Widget add/remove mutations unaffected

### DIFF
--- a/src/pocketpaw/agents/loop.py
+++ b/src/pocketpaw/agents/loop.py
@@ -178,11 +178,54 @@ async def _create_pocket_and_session(spec: dict, session_key: str) -> str | None
         return None
 
 
+async def _update_pocket(spec: dict) -> str | None:
+    """Update an existing pocket's rippleSpec in MongoDB. Returns pocket _id or None."""
+    try:
+        from ee.cloud.models.user import User
+        from ee.cloud.pockets.schemas import UpdatePocketRequest
+        from ee.cloud.pockets.service import PocketService
+
+        # Extract pocket_id from the spec's lifecycle.id
+        pocket_id = spec.get("lifecycle", {}).get("id", "")
+        if not pocket_id:
+            pocket_id = spec.get("metadata", {}).get("pocket_id", "")
+        if not pocket_id:
+            logger.warning("Cannot update pocket — no pocket_id in spec")
+            return None
+
+        # Find the cloud pocket by lifecycle.id match
+        from ee.cloud.models.pocket import Pocket
+
+        pocket = await Pocket.find_one({"rippleSpec.lifecycle.id": pocket_id})
+        if not pocket:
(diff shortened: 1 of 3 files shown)


prakashUXtech:
Closing — superseded by #1085 (DevRohit's pocket-specialist squash).

**Why.** #925 was the coarse-grained shape: agent passes `pocket_id` to `CreatePocketTool`, the loop calls `PocketService.update()` with a whole new rippleSpec, one `pocket_updated` SSE event fires. #1085 introduces a strictly better shape for the same problem:

- First-class `pocket_specialist__edit` mirrors create — no more reusing `CreatePocketTool` with an optional flag.
- Granular UI/state mutations (5 node ops + 4 state ops in `ee/cloud/pockets/spec_ops.py` + `state_ops.py`) replace whole-spec rewrites.
- SSE events carry only the changed subtree, not the full new spec.

Both PRs touch the same three files (`agents/loop.py`, `api/v1/pockets.py`, `tools/builtin/pocket.py`); #1085 reshapes them so heavily that a rebase wouldn't preserve meaningful logic.

The companion paw-enterprise#61 also drifted — it was tagged as the frontend counterpart for `pocket_updated` SSE handling, but it's actually a `data_sources` feature for pocket specs (parallel `fetch()` resolver + wiki redesign). The two PRs are independent now.

**What to preserve.** If #1085 doesn't already teach the agent (via system prompt) to invoke `pocket_specialist__edit` when `<current-pocket>` is in context, that's a one-line follow-up in a future PR. Otherwise nothing — the problem #925 solved is fully covered by the new architecture.

Verified `origin/ee` doesn't already have #925's content: `CreatePocketTool` still generates an internal uuid, no `_update_pocket` exists in `loop.py`. So this is a genuine close, not a duplicate-with-merge.

PR #927, opened 2026-04-12
Dynamic Ripple widget docs for pocket agents via kb-go

prakashUXtech:
## What this does

When a user asks PocketPaw to build a pocket, the agent now gets relevant widget documentation searched from the Ripple knowledge base — instead of relying solely on the static system prompt that lists widget names without detail.

## How it works

Added `_get_ripple_widget_context()` to `pocket_chat_stream`. Before the message hits the agent, it:

1. Searches `kb search "<user message>" --scope ripple --context --limit 3` (~10ms)
2. Gets back the top 3 relevant articles (e.g., "sales dashboard" returns metric, chart, table docs)
3. Appends them inside a `<ripple-widget-reference>` tag after the static pocket system context

The agent then sees both the structural rules (how to invoke pocket tools, formats, data rules) AND deep widget knowledge (props, types, constraints, JSON examples) specific to what the user is asking to build.

## Failure handling

Falls back silently on any failure — missing binary, missing scope, timeout, non-zero exit. kb is a nice-to-have enhancement, never a blocker. Same pattern as the existing `_get_kb_context()` in context_builder.py.

## Companion PR

This pairs with [qbtrix/ripple-svelte#17](https://github.com/qbtrix/ripple-svelte/pull/17) which adds the widget knowledge base (10 articles, 34 widgets, 173 concepts) to the Ripple repo.

## Test plan

- [x] `test_context_budget.py` — all 15 tests pass
- [x] Syntax validated via `ast.parse`
- [x] Graceful fallback: returns empty string when kb binary missing, scope empty, or timeout
- [ ] Manual: verify agent generates better specs with widget context vs without

### DIFF
--- a/src/pocketpaw/api/v1/pockets.py
+++ b/src/pocketpaw/api/v1/pockets.py
@@ -1,7 +1,8 @@
 # Pocket chat router — dedicated endpoint for pocket creation.
-# Updated: 2026-04-01 — _prepare_pocket_spec now handles UISpec v1.0, multi-pane specs,
-#   and flat widget specs. System prompt teaches all three formats with UISpec as default.
-#   Debug logging added to trace pocket event pipeline.
+# Updated: 2026-04-12 — Added dynamic Ripple widget knowledge injection via kb-go.
+#   _get_ripple_widget_context() searches the 'ripple' kb scope for widget docs
+#   relevant to the user's request. Results are appended to the static pocket system
+#   context so agents get deep, targeted knowledge about the widgets they need.
 
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


prakashUXtech:
Read the diff. Verdict: LGTM, merging after CI clears.

The kb-go integration is a straight subprocess call — `asyncio.create_subprocess_exec` with the binary path from `settings.kb_binary`, 3-second timeout, returncode guarded. FileNotFoundError, TimeoutError, and the bare Exception fallback all return `""`, so a missing binary, missing `ripple` scope, or empty kb output leaves the prompt untouched. This matches `_get_kb_context` in `bootstrap/context_builder.py` exactly, which is the right call — one failure mode, one recovery pattern.

The only nit was a ruff format miss (one blank line between the function and `_POCKET_SYSTEM_CONTEXT`). Pushed `54ede3c` to fix it so lint goes green.

Observations worth tracking as follow-ups, not blockers:

- No journal emission for `kb.source.ingested` / `kb.article.compiled`. PocketPaw is read-side here (calling `kb search`), not ingest-side, so there's nothing to emit yet. When ingest lands inside pocketpaw, the events should flow through the existing SystemEvent channel so the dashboard Activity panel can show them.
- The 3s timeout is inherited from the context_builder pattern. For a chat-path call gating the first token, 3s is a lot of tail. If kb-go p99 is comfortably under 50ms on warm indexes, worth tightening to 1s.
- `user_message` goes to kb-go as a positional search arg, which is fine for kb-go's internal search but means the full user turn lands in its logs if kb logging is ever enabled. Not a concern today, flagging for the security-auditor pass.

Test suite green locally (3991 passed, 51 skipped). Ship it.

PR #928, opened 2026-04-12
Capture corrections between proposal and approval

prakashUXtech:
## Why

The Instinct pipeline already captured what the agent proposed and what the human ultimately decided. It did not capture the delta between them — the edits a rep made to a draft before hitting approve. That delta is the single cheapest learning signal we have, and it was being discarded every time.

This PR lands the first part of the Correction Loop: the data model, storage, and endpoint surface. PR-B will wire the captured corrections into soul-protocol so the agent learns from them on the next proposal. PR-C (paw-enterprise) adds the Edit & Approve UI.

### `ee/instinct/correction.py` (new)
- `CorrectionPatch` and `Correction` Pydantic models.
- `compute_patches(before, after)` structural diff across fields a human would actually edit: `title`, `description`, `recommendation`, `category`, `priority`, and the top-level keys of `parameters`. `context` is intentionally skipped — it's reasoning metadata, not action content, and will be handled by the Move 2 decision-trace collector.
- `summarize_correction(action, patches)` formats a deterministic, LLM-free recall key for the soul bridge to consume in PR-B.

### `ee/instinct/store.py`
- New `instinct_corrections` SQLite table with indexes on `pocket_id` and `action_id`.
- `record_correction()` persists the row and logs a `correction_captured` audit event carrying the patch paths.
- `get_corrections_for_pocket` / `get_corrections_for_action` / `count_corrections_by_path` read paths.

### `ee/instinct/router.py`
- `POST /instinct/actions/{id}/approve` accepts an optional `ApproveRequest` body with edited fields. When edits differ from the stored proposal, the server diffs the two, persists a `Correction`, and writes the edits back to the action before transitioning its status to approved.
- Response is now `ApproveResponse { action, correction | null }` so callers always know whether learning was captured. No-body POST still approves unchanged.
- `GET /instinct/corrections?pocket_id=...|action_id=...` exposes captured deltas for the UI and the agent tool in PR-B.

## Test plan

- [x] `uv run pytest tests/cloud/test_ee_correction.py` — 21 new tests, all passing
- [x] `uv run pytest tests/ --ignore=tests/e2e --ignore=tests/test_frontend_syntax.py` — 3991 passed, 0 failed
- [x] `uv run ruff check ee/instinct/ tests/cloud/test_ee_correction.py` — clean

## Follow-ups

- **PR-B** (this workstream) — soul bridge + `instinct_corrections` agent tool + `context_builder` hook so past edits shape the next proposal.
- **PR-C** (paw-enterprise) — Edit & Approve UI. Will also update `src/lib/core/runtime/api.ts` to consume the new `ApproveResponse` shape (existing callers only use the returned object at runtime for optimistic updates; types need refresh).

## Context

Design doc: `paw-enterprise/docs/enterprise/CORRECTION-LOOP.md`.

### DIFF
--- a/ee/instinct/__init__.py
+++ b/ee/instinct/__init__.py
@@ -1,8 +1,16 @@
 # Instinct — decision pipeline for Paw OS.
 # Created: 2026-03-28 — Actions, approvals, audit log.
 # Updated: 2026-03-30 — Exported ActionStatus, ActionCategory, ActionPriority, AuditCategory.
-# The decision loop: Agent proposes -> Human approves -> Action executes -> Feedback captured.
+# Updated: 2026-04-12 (Move 1 PR-A) — Exported Correction, CorrectionPatch, compute_patches.
+# The decision loop: Agent proposes -> Human approves (optionally edits) ->
+# Action executes -> Correction captured -> Soul learns.
 
+from ee.instinct.correction import (
+    Correction,
+    CorrectionPatch,
+    compute_patches,
+    summarize_correction,
+)
 from ee.instinct.models import (
     Action,
     ActionCategory,
@@ -24,5 +32,9 @@ __all__ = [
     "ActionTrigger",
     "AuditCategory",
     "AuditEntry",
+    "Correction",
+    "CorrectionPatch",
     "InstinctStore",
+    "compute_patches",
+    "summarize_correction",
 ]
--- a/ee/instinct/correction.py
+++ b/ee/instinct/correction.py
@@ -1,0 +1,101 @@
+# ee/instinct/correction.py — Correction Loop data types.
+# Created: 2026-04-12 (Move 1 PR-A) — Captures the diff between what the agent
+# proposed and what the human approved so it can be used as a learning signal.
+# Pairs with soul-protocol to form the correction loop: proposal → edit → soul
+# remembers → next proposal improves.
+
+from __future__ import annotations
+
+from datetime import datetime
+from typing import Any
+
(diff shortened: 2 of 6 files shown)


prakashUXtech:
## Self-review

First part of the correction loop. `CorrectionPatch` + `Correction` Pydantic models, new corrections table, `/approve` that accepts optional edits and diffs server-side.

**Notes**
- `ee/instinct/router.py` `_persist_edits` reaches into `store._db_path` to run a raw UPDATE (~L321). That's a layer violation — the store should expose `update_action_fields()` so the helper doesn't crack open private state.
- `ee/instinct/store.py` `count_corrections_by_path` scans up to 1000 corrections in Python and counts path hits. Fine at current volume, but a `SELECT COUNT(*)` with a JSON-path or LIKE filter wins before the table grows.
- `GET /instinct/corrections` caps at `limit=500` but has no offset / cursor. Enough for "show recent in a pocket", not enough for long history.
- `compute_patches` diffs `parameters` at the top level only. A nested dict change produces one patch with the full before/after object — documented in code but worth surfacing in the API response schema.
- Response shape changed from `Action` to `ApproveResponse { action, correction | null }`. Any existing external caller needs updating. PR-C handles the frontend.

**Follow-up issues**
- Extract \`update_action_fields()\` into \`InstinctStore\` — replace the raw-SQL helper.
- Swap Python-side path counting for SQL.
- Add pagination cursor on \`/instinct/corrections\`.

PR #929, opened 2026-04-12
Route captured corrections into soul + expose a read tool

prakashUXtech:
## Why

Part two of the correction loop. [PR #928](https://github.com/pocketpaw/pocketpaw/pull/928) captured the diff between proposal and approval; this PR feeds that signal into soul-protocol and makes it available to the agent on the next draft.

Stacks on top of `feat/correction-loop-models` — ready to merge into `dev` once PR-A lands.

### `ee/instinct/correction_soul_bridge.py` (new)
- `CorrectionSoulBridge.record(correction, action)` observes the edit as a soul `Interaction` with the full patch list in `agent_output` — enters the episodic tier with a recall-friendly body.
- Counts how many times the same `path` has been edited in this pocket. On the **third match**, synthesizes a short deterministic rule (no LLM call on the hot path) and stores it as a procedural memory at importance 7.
- Degrades silently when the soul is absent, or when any step of the soul handoff raises. Approval must never fail because the soul is sick. Corrections still persist to SQLite either way.

### `ee/instinct/router.py`
- `/approve` forwards to the bridge after `record_correction` succeeds. Best-effort, never blocks the response.

### `src/pocketpaw/tools/builtin/instinct_corrections.py` (new)
- New `instinct_corrections` agent tool. The agent calls it with a `pocket_id` before proposing a new action and gets back a formatted list of recent edits with patch paths. Pre-applying those patterns is the whole point — the draft reflects the user's history before it reaches the Tray.
- Registered alongside `instinct_propose` / `instinct_pending` / `instinct_audit` in the enterprise tool block. All 7 agent backends pick it up.

## Test plan

- [x] 11 new tests in `tests/cloud/test_correction_soul_bridge.py`:
  - observe payload shape (summary + patches in agent_output)
  - 3x-same-path promotion fires exactly once
  - promotion tracks different paths independently
  - graceful no-soul degradation
  - observe exception is swallowed
  - tool empty state, populated state, ee-missing message, parameter schema
- [x] `uv run pytest tests/ --ignore=tests/e2e --ignore=tests/test_frontend_syntax.py` — 3991 passed, 0 failed
- [x] Optional `soul_protocol` dep stubbed via autouse fixture so tests run in the base dev env

## Note on context injection

The original plan called for a `context_builder.py` step 4f to auto-inject corrections. We skipped it this PR because soul-protocol's existing `SoulBootstrapProvider` already pulls in the episodic memories we wrote via `soul.observe()`, and the `instinct_corrections` tool provides an explicit fallback when soul isn't loaded. Adding a third injection path would duplicate the signal and couple `context_builder` to the instinct store unnecessarily.

## Follow-ups

- **PR-C** (paw-enterprise) — Edit & Approve UI + `api.ts` type refresh for the new `ApproveResponse` shape.

## Context

Design doc: `paw-enterprise/docs/enterprise/CORRECTION-LOOP.md`.

### DIFF
--- a/ee/instinct/correction_soul_bridge.py
+++ b/ee/instinct/correction_soul_bridge.py
@@ -1,0 +1,151 @@
+# ee/instinct/correction_soul_bridge.py — Wires Corrections into soul-protocol.
+# Created: 2026-04-12 (Move 1 PR-B) — Turns each captured human edit into a soul
+# observation, and promotes repeated edits on the same field into a procedural
+# memory. No new soul primitive — uses soul.observe() + soul.remember() as-is.
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+from ee.instinct.correction import Correction, CorrectionPatch
+from ee.instinct.models import Action
+
+logger = logging.getLogger(__name__)
+
+_PROMOTION_THRESHOLD = 3  # Same path edited N times → procedural memory
+_PROCEDURAL_IMPORTANCE = 7
+_EPISODIC_IMPORTANCE = 5
+
+if TYPE_CHECKING:
+    from ee.instinct.store import InstinctStore  # noqa: F401
+
+
+class CorrectionSoulBridge:
+    """Connects captured Corrections to soul-protocol's learning hooks.
+
+    Call `record(correction, action)` right after persisting the Correction
+    in the store. The bridge:
+
+    - Always observes the edit as an Interaction so it enters the soul's
+      episodic tier with a recall-friendly summary.
+    - Counts how many times the same `path` has been edited in this pocket.
+      On the third match, synthesizes a short rule and stores it as a
+      procedural memory (importance 7).
+
+    The bridge degrades silently when the soul is not loaded — corrections
+    still persist to SQLite, and the agent tool can still read them back.
+    Nothing in the request path fails because soul is down.
+    """
+
+    def __init__(self, soul_manager: object, store: object) -> None:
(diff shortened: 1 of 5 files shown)


prakashUXtech:
## Self-review

Part two of the correction loop. `CorrectionSoulBridge` routes each captured diff into soul-protocol, new `instinct_corrections` agent tool reads recent corrections back for the next proposal.

**Notes**
- `ee/instinct/correction_soul_bridge.py` `_synthesize_rule` embeds \`correction.actor\` and both \`patch.before\` / \`patch.after\` verbatim into a string stored as a *procedural* memory. Procedural memories are durable and travel with \`.soul\` exports. If before/after carries PII or adversarial content, it lands in the soul unsanitized. Cap each value and strip control characters before embedding.
- \`src/pocketpaw/tools/builtin/instinct_corrections.py\` returns raw exception text: \`f"Error loading corrections: {exc}"\`. Exception strings can leak internal paths. Log the detail, return a generic message.
- Promotion fires on \`count == 3\` exactly. If the table is pre-seeded or a test inserts rows without going through the bridge, the threshold can be missed. \`count > 0 and count % 3 == 0\` is more robust.
- Graceful degradation when soul is absent looks right — tests cover the no-op path.
- \`InstinctCorrectionsTool\` appended at the end of \`_EE_TOOLS\`; the list is otherwise loosely grouped by feature. Minor consistency.

**Follow-up issues**
- Sanitize actor + patch values before writing to procedural soul memory.
- Replace raw exception exposure in \`instinct_corrections.execute\`.

PR #930, opened 2026-04-12
ReasoningTrace + TraceCollector + fabric snapshots

prakashUXtech:
## Why

The audit log already records that a decision happened — actor, event, timestamp. It never recorded *why*. This PR lays the foundation so every proposal carries its reasoning context.

Stacks on `feat/correction-loop-soul` ([#929](https://github.com/pocketpaw/pocketpaw/pull/929)) to keep the diff clean. First of three PRs that land Move 2 (Decision Traces).

### `ee/instinct/trace.py`
- `ToolCallRef` — tool invocation captured during reasoning: tool name, stable args fingerprint, 200-char result preview, duration. Hash lets repeated calls dedupe without losing the first result.
- `ReasoningTrace` — the top-level structure stored inside `AuditEntry.context["reasoning_trace"]`. Holds referenced `fabric_query` / `soul_recall` / `kb_article` IDs, `tool_calls`, and prompt version + backend + model.
- `FabricObjectSnapshot` — immutable snapshot of a Fabric object at decision time, keyed by `(object_id, audit_id)`. The trace stores IDs; the snapshot lets a compliance reviewer reproduce inputs three months later.

### `ee/instinct/trace_collector.py`
- `TraceCollector` async context manager (will wrap `instinct_propose` in PR-B). Subscribes to `SystemEvent`s on enter, aggregates `fabric_query` / `soul_recall` / `kb_inject` / `tool_start` / `tool_end` (or `tool_result`), unsubscribes on exit — even when the body raises.
- Tool calls deduped on `(tool, args_hash)`. Reference lists deduped on exit, preserving first-seen order.
- Unknown types and malformed payloads are silently ignored. The collector never fails the surrounding proposal.

### `InstinctStore`
- `record_fabric_snapshot` / `get_snapshots_for_audit` / `get_snapshots_for_object`.
- New `instinct_fabric_snapshots` SQLite table, indexed on both `audit_id` and `object_id`.

## Test plan

- [x] 19 new tests in `tests/cloud/test_decision_traces.py`:
  - model round-trip
  - subscribe/unsubscribe on enter + exit (including exception path)
  - every event-type branch (`fabric_query`, `soul_recall`, `kb_inject`, `tool_start`+`tool_end`, `tool_result` alias)
  - tool-call dedup and 200-char truncation
  - malformed-event tolerance
  - snapshot read paths (audit-scoped ordered oldest-first, object-scoped newest-first)
- [x] `uv run pytest tests/ --ignore=tests/e2e --ignore=tests/test_frontend_syntax.py` — 3991 passed

## Follow-ups

- **PR-B** — Wrap `instinct_propose` agent tool with `TraceCollector`, serialize trace into `AuditEntry.context.reasoning_trace` at proposal time, snapshot referenced Fabric objects, and add `GET /instinct/audit/{id}?hydrate=1` for the UI.
- **PR-C** (paw-enterprise) — Why? drawer in `AuditLogPanel` with Data / Memory / Knowledge tabs.

## Context

Design doc: `paw-enterprise/docs/enterprise/DECISION-TRACES.md`.

### DIFF
--- a/ee/instinct/__init__.py
+++ b/ee/instinct/__init__.py
@@ -22,6 +22,8 @@ from ee.instinct.models import (
     AuditEntry,
 )
 from ee.instinct.store import InstinctStore
+from ee.instinct.trace import FabricObjectSnapshot, ReasoningTrace, ToolCallRef
+from ee.instinct.trace_collector import TraceCollector
 
 __all__ = [
     "Action",
@@ -34,7 +36,11 @@ __all__ = [
     "AuditEntry",
     "Correction",
     "CorrectionPatch",
+    "FabricObjectSnapshot",
     "InstinctStore",
+    "ReasoningTrace",
+    "ToolCallRef",
+    "TraceCollector",
     "compute_patches",
     "summarize_correction",
 ]
--- a/ee/instinct/store.py
+++ b/ee/instinct/store.py
@@ -25,6 +25,7 @@ from ee.instinct.models import (
     AuditCategory,
     AuditEntry,
 )
+from ee.instinct.trace import FabricObjectSnapshot
 
 SCHEMA_SQL = """
 CREATE TABLE IF NOT EXISTS instinct_actions (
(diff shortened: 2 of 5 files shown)


prakashUXtech:
## Self-review

Data layer for decision traces. `ReasoningTrace` + `ToolCallRef` + `FabricObjectSnapshot` Pydantic models, `TraceCollector` async-context-manager subscribing to the bus, new `instinct_fabric_snapshots` table with indexes on `(audit_id)` and `(object_id)`.

**Notes**
- \`ee/instinct/trace_collector.py\` \`_record_tool_end\` dedupes \`(tool, args_hash)\` and keeps the FIRST result. For deterministic tools that's ideal; for non-deterministic tools called twice with the same args, the second result is silently dropped. Document the choice or flip to last-write-wins.
- \`_hash_args\` truncates SHA-256 to 16 hex chars (64 bits). Collision probability is negligible for per-trace dedup, but add a comment acknowledging the trade-off so a future reader doesn't "widen it for safety."
- \`ee/instinct/trace.py\` \`FabricObjectSnapshot.created_at\` uses naive \`datetime.now()\`. Rest of the codebase is also naive, so no drift, but a sweep to timezone-aware \`datetime.now(UTC)\` is worth filing.
- \`__aexit__\` mutates \`self.trace.*\` lists (dedup) on exit. A caller reading the trace inside the \`async with\` block sees undeduped state. Not a bug, worth a docstring note.
- 19 tests cover lifecycle + every event-type branch + malformed tolerance. Good coverage.

**Follow-up issues**
- Document or change the first-wins tool-call dedup policy.
- Timezone-aware datetimes across \`ee/instinct/\` in one sweep.

PR #931, opened 2026-04-12
Persist reasoning traces and hydrate audit entries

prakashUXtech:
## Why

PR-A ([#930](https://github.com/pocketpaw/pocketpaw/pull/930)) landed the data types. This PR wires them into the proposal path and adds the read endpoint the UI will consume.

Stacks on `feat/decision-traces-models`.

### Store
- `InstinctStore.propose()` now takes optional `reasoning_trace` and `fabric_snapshots` kwargs.
- Trace is serialized into `AuditEntry.context["reasoning_trace"]` when present.
- Snapshots are persisted via `record_fabric_snapshot()` with their `audit_id` rewritten to the freshly created audit row — callers don't need to know the audit ID ahead of time.

### Router
- `POST /instinct/actions` accepts `reasoning_trace` and `fabric_snapshots` on the body so direct HTTP callers (and the agent tool) can attach decision inputs at propose time.
- **`GET /instinct/audit/{id}`** is the new read endpoint.
  - `hydrate=0` (default): entry + decoded trace if stored.
  - `hydrate=1`: also expands referenced Fabric IDs into immutable snapshots (captured at decision time) and live state (current object). Reviewers can compare what the agent saw against what the object looks like now.
- `HydratedAuditEntry` response: `{ entry, reasoning_trace, fabric_snapshots, fabric_current }`.
- `fabric_current` is best-effort; if ee/ isn't installed or an object was deleted, its entry is omitted.

### FabricQueryTool
- Emits `SystemEvent(event_type="fabric_query", data={"object_id": ..., "object_type": ...})` for each returned object.
- When a `TraceCollector` is active (agent-loop wiring — follow-up), these land in the trace automatically.
- In every other context it's a no-op; any failure is logged at debug and swallowed so tool calls never break because telemetry is sick.

## Test plan

- [x] 8 new tests in `tests/cloud/test_decision_traces_wiring.py`:
  - `propose()` with trace — lands in audit context
  - `propose()` with snapshots — re-keyed to audit row
  - `propose()` without trace — legacy callers still work
  - `POST /instinct/actions` — accepts both fields
  - `hydrate=0` — decoded trace, no expansion
  - `hydrate=1` — snapshots returned
  - unknown audit ID → 404
  - audit entry without trace → hydrates empty
- [x] `uv run pytest` — 3991 passed

## Follow-ups

- **Agent-loop integration** — wrap `instinct_propose` and the agent turn in a `TraceCollector` so fabric/soul/kb events that happen during reasoning are captured automatically. Deferred to a follow-up PR to keep this one surgical.
- **PR-C** (paw-enterprise) — Why? drawer in `AuditLogPanel` consuming this endpoint.

## Context

Design doc: `paw-enterprise/docs/enterprise/DECISION-TRACES.md`.

### DIFF
--- a/ee/instinct/router.py
+++ b/ee/instinct/router.py
@@ -6,6 +6,11 @@
 #   When present, the server diffs the stored proposal against the edits, persists
 #   a Correction, then approves. GET /instinct/corrections exposes corrections
 #   scoped to a pocket or an action so the UI and agents can read them back.
+# Updated: 2026-04-13 (Move 2 PR-B) — POST /instinct/actions accepts an optional
+#   reasoning_trace + fabric_snapshots body so callers (and the agent tool) can
+#   attach decision inputs at propose time. GET /instinct/audit/{id}?hydrate=1
+#   returns the audit entry with the trace's referenced IDs expanded into Fabric
+#   object snapshots, making the "Why?" drawer possible in the UI.
 
 from __future__ import annotations
 
@@ -29,6 +34,7 @@ from ee.instinct.models import (
     ActionTrigger,
     AuditEntry,
 )
+from ee.instinct.trace import FabricObjectSnapshot, ReasoningTrace
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


prakashUXtech:
## Self-review

Wires traces into the propose path and adds the hydration endpoint. `store.propose()` now accepts `reasoning_trace` + `fabric_snapshots`, new `GET /instinct/audit/{id}?hydrate=1` returns decoded trace + snapshots + live Fabric state. `FabricQueryTool` emits `fabric_query` SystemEvents.

**Notes**
- \`ee/instinct/router.py\` \`get_audit_entry\` calls \`store.query_audit(limit=1000)\` and linear-scans to find the matching id. O(N) on every hydration hit, hard-capped at the most recent 1000 entries — older audits become un-hydratable. Add \`InstinctStore.get_audit_by_id(audit_id: str)\` with \`WHERE id = ?\`. Blocking for any real audit volume.
- Route order: \`GET /instinct/audit\` (query), \`GET /instinct/audit/{audit_id}\` (new), \`GET /instinct/audit/export\` (existing). Because \`/export\` is registered AFTER the \`{audit_id}\` param route in file order, verify FastAPI routes \`/export\` correctly — a regex constraint on \`{audit_id}\` excluding \`export\` is safer than trusting order.
- \`propose()\` mutates \`snapshot.audit_id\` in place before persisting. Pydantic v2 models are mutable, but callers reusing a snapshot instance will be surprised. Copy + assign instead.
- \`_fetch_current_fabric\` does N individual \`get_object\` calls. Fine at N < 10; add a bulk method once traces reference more objects.
- \`FabricQueryTool\` emits events unconditionally. No subscribers in the common case — cheap but wasteful.

**Follow-up issues**
- \`InstinctStore.get_audit_by_id\` — replace the full-scan hydration.
- Verify route ordering / add regex constraint on \`{audit_id}\`.
- Bulk Fabric get for hydration.

PR #932, opened 2026-04-12
Paw Print backend — models + async SQLite store

prakashUXtech:
## Why

Kicks off Move 3 — the customer-facing widget. Palantir's decision loop ends at the operator. Paw OS closes the loop to the customer: events on a brand-embedded widget flow into a Pocket, Instinct nudges the owner, the approved action flows back to the widget. This PR is the backend foundation of that loop.

Stacks on `feat/decision-traces-wiring` (pocketpaw #931).

### `ee/paw_print/models.py`
- `PawPrintBlock` — tagged-union render primitive: `text` / `image` / `list` / `button` / `form` / `divider`. No raw HTML, no arbitrary JS. Forward-compatible.
- `PawPrintSpec` — what the widget bundle renders. `<= 64 blocks per spec`, `<= 50 items per list`.
- `PawPrintWidget` — owner-facing model. Validators lowercase + dedupe `allowed_domains` (cap 20), enforce positive rate limits, and mint per-widget tokens via `secrets`.
- `PawPrintEvent` — inbound signal. Type required + stripped. `payload_size()` helps PR-B enforce the 4 KB cap.
- `PawPrintEventMapping` — how a widget event becomes a Fabric object (template interpolation lives in PR-B).

### `ee/paw_print/store.py`
- Async SQLite tables for widgets + events, matching `InstinctStore`'s shape.
- Indexes on `(widget_id, timestamp DESC)` and `(widget_id, customer_ref)` — keeps PR-B's rate limiter cheap.
- CRUD + `update_spec` + `rotate_token` (mints a fresh `pp_tok_…`).
- Append-only event log with `recent_events`, `count_events_since`, and the composite `within_rate_limit(overall_per_min, per_customer_per_min, customer_ref)` primitive.

## Test plan

- [x] 19 tests in `tests/cloud/test_paw_print_backend.py`:
  - block + spec caps enforced
  - domain normalization + dedup + cap
  - rate-limit + token validation
  - CRUD round-trip including update_spec, rotate_token, delete idempotency
  - event ordering, window counting
  - overall-vs-per-customer rate-limit interaction
- [x] `uv run pytest` — 3991 passed
- [x] `uv run ruff check` — clean

## Follow-ups

- **PR-B** — HTTP router (`GET /paw-print/spec/{id}`, WS `/paw-print/ws/{id}`, `POST /paw-print/events/{id}`), domain/origin enforcement, Guardian hook, event → Fabric mapping.
[...]

PR #933, opened 2026-04-12
Paw Print — HTTP surface + event ingest + Fabric mapping

prakashUXtech:
## Why

PR-A landed the data layer. This PR is the HTTP face of Paw Print: spec serving for the embedded widget, owner-authed CRUD for widget management, and the inbound event pipeline that turns customer clicks into Fabric objects.

Stacks on `feat/paw-print-backend` (#932).

## Endpoints

All mounted at `/paw-print/`.

**Owner surface — require `X-Paw-Print-Token` header:**
- `POST   /widgets` — create
- `GET    /widgets` — list (filter by `pocket_id` / `owner`)
- `GET    /widgets/{id}`
- `PATCH  /widgets/{id}/spec`
- `POST   /widgets/{id}/rotate-token`
- `DELETE /widgets/{id}`
- `GET    /widgets/{id}/events`

**Public surface — CORS-gated per widget, no token:**
- `GET /paw-print/spec/{widget_id}` — returns `PawPrintSpec` with `Access-Control-Allow-Origin` echoed back
- `POST /paw-print/events/{widget_id}` — the ingest hot path

## Security ordering on event ingest

1. Widget exists (404 otherwise)
2. `Origin` header is on the widget's allowlist (403 otherwise). Empty allowlist disables the check — fine for local dev, **must** be set in production. Host-match ignores ports and paths.
3. Payload size ≤ `MAX_PAYLOAD_BYTES` (4 KB). 413 otherwise.
4. Rate limits — both overall-per-minute and per-customer-per-minute. 429 otherwise.
5. Guardian (existing input sanitizer) screens the JSON payload. Missing guardian → skip check (telemetry must never become a dependency that breaks customer interactions).

After that the event is persisted, and when the widget has a matching `event_mapping` entry, `{{ placeholder }}` fields are interpolated against `{payload, customer_ref}` and a Fabric object is created. Whole-placeholder templates preserve raw types; mixed strings fall back to stringified substitution.

## Test plan

- [x] 19 tests in `tests/cloud/test_paw_print_ingest.py`:
  - Widget CRUD with token auth (create/get/rotate/delete/list-events) — 401 without header, 204 on delete, 404 after
  - Spec endpoint across allowed / disallowed / missing origin + empty-allowlist opens up
  - Event ingest: happy path, origin rejection, oversized payload 413, per-customer rate limit firing on the 4th call, Guardian rejection returns `accepted=false`, Fabric mapping creates an object
  - `_interpolate` across full-placeholder / mixed string / missing-path
- [x] `uv run pytest` — 3991 passed

## Follow-ups

- **PR-C** — `paw-print-widget` vanilla JS bundle (new repo).
- **PR-D** — `PawPrintPanel.svelte` admin UI in paw-enterprise.
- Router must still be mounted in `ee.api` — PR-D wires it into the live app.

## Context

Design doc: `paw-enterprise/docs/enterprise/PAW-PRINT-MVP.md`.

### DIFF
--- a/ee/paw_print/router.py
+++ b/ee/paw_print/router.py
@@ -1,0 +1,389 @@
+# ee/paw_print/router.py — HTTP surface for the Paw Print widget layer.
+# Created: 2026-04-13 (Move 3 PR-B) — Spec serving (public, CORS-gated),
+# widget CRUD (owner-authed via access_token), event ingest (rate-limited,
+# domain-enforced, Guardian-screened, Fabric-mapped). The widget.js bundle
+# built in PR-C consumes these endpoints.
+
+from __future__ import annotations
+
+import json
+import logging
+import re
+from typing import Any
+
+from fastapi import APIRouter, Header, HTTPException, Query, Request
+from fastapi.responses import JSONResponse
+from pydantic import BaseModel, Field
+
+from ee.paw_print.models import (
+    MAX_PAYLOAD_BYTES,
+    PawPrintEvent,
+    PawPrintEventMapping,
+    PawPrintSpec,
+    PawPrintWidget,
+)
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(tags=["PawPrint"])
+
+_PLACEHOLDER_RE = re.compile(r"\{\{\s*([a-zA-Z0-9_.]+)\s*\}\}")
+
+
+def _store():
+    from ee.api import get_paw_print_store
+
+    return get_paw_print_store()
+
+
+def _require_owner_token(widget: PawPrintWidget, header_token: str | None) -> None:
+    if not header_token or header_token != widget.access_token:
+        raise HTTPException(status_code=401, detail="Invalid or missing access token")
+
+
+def _origin_allowed(widget: PawPrintWidget, origin: str | None) -> bool:
+    """Match an inbound Origin header against the widget's allowed_domains.
+
+    Empty `allowed_domains` disables the check — useful for local demos but
+    must be set in production. The match is host-only so ports and paths don't
+    matter: `https://brewco.com:443/menu` matches `brewco.com`.
+    """
+    if not widget.allowed_domains:
+        return True
+    if not origin:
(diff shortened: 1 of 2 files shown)


prakashUXtech:
## Self-review

HTTP surface for paw-print. Public spec + event endpoints (CORS-gated), owner-authed widget CRUD, five-step ingest pipeline (widget → origin → size → rate limit → guardian → persist → map).

**Notes**
- \`ee/paw_print/router.py\` \`_require_owner_token\` does \`header_token != widget.access_token\`. Plain string comparison is not constant-time. Swap for \`hmac.compare_digest()\`. Blocking before any public deploy.
- \`POST /paw-print/widgets\` (create) has NO auth. Any caller can mint a widget under any \`pocket_id\` with any \`owner\` string. The token-on-the-widget pattern only kicks in after creation. Needs a pre-creation gate (session auth, or caller-matches-owner check) before this goes live.
- No WebSocket endpoint in this PR despite the design doc calling for one. Polling is the fallback until a future PR lands \`/paw-print/ws/{id}\`.
- \`_apply_event_mapping\` wraps the Fabric create in a bare \`except Exception\` that only \`logger.exception\`s. Customer-facing events silently fail to map without any surfaced signal. At minimum return \`fabric_object_id=None\` AND emit an audit event so operators can detect mapping failures.
- \`_pass_through_guardian\` accepts the event when the guardian raises — correct degradation, but no counter / ops signal when the security layer silently falls off.
- \`_origin_allowed\` strips ports + paths. \`https://brewco.com:9999\` matches \`brewco.com\`. Documented behavior, worth a comment so operators know the scope.
- 19 tests — every security step covered.

**Follow-up issues**
- \`hmac.compare_digest\` in \`_require_owner_token\` — blocking.
- Auth on \`POST /paw-print/widgets\` before first deploy.
- Audit event when guardian is unavailable.
- Surface Fabric mapping failures explicitly.
- File an issue for the WebSocket push endpoint.

PR #936, opened 2026-04-13
feat(ee): retrieval log sink + HTTP + CLI

prakashUXtech:
The companion runtime layer for soul-protocol's `RetrievalTrace` primitive ([soul-protocol #161](https://github.com/qbtrix/soul-protocol/pull/161)). Provides one place where every recall / search / match writes a receipt, plus the three reads off that file: HTTP, CLI, and the upcoming graduation policy + Why? drawer.

Move 4 PR-B. Targets `dev`. PR-C (graduation) and PR-D (UI) follow.

### \`ee/retrieval/log.py\`
- \`RetrievalLog\` — async-safe JSONL sink at \`~/.pocketpaw/retrieval.jsonl\` (override via \`POCKETPAW_RETRIEVAL_LOG\`). \`asyncio.Lock\` around appends serialises concurrent writers in-process. Cross-process safety intentionally out of scope until multi-process surfaces.
- \`read()\` — actor / source / pocket_id / since / until / limit filters, newest-first, malformed-line tolerant, missing-file safe.
- \`tail()\` reverses for terminal display. \`stats()\` for ops counters. \`clear()\` is test-only.
- Process singleton via \`get_log()\` + \`reset_log_for_tests()\`.

### \`ee/retrieval/models.py\`
- \`RetrievalLogEntry\` wraps the raw \`RetrievalTrace\` dict with paw-runtime context (ingestion timestamp, process tag, optional session id).
- Convenience accessors (\`actor()\`, \`source()\`, \`query()\`, \`pocket_id()\`, \`candidate_ids()\`) save callers from re-deserialising the embedded dict.

### \`ee/retrieval/router.py\`
- \`GET /retrieval/log\` — same filter set as the sink + \`limit\` (1-1000). Returns \`{ entries, total }\`.
- \`GET /retrieval/stats\` — \`{ total, actors, pockets }\`.

### \`pocketpaw retrieval\` CLI
- \`pocketpaw retrieval tail [--actor X] [--source Y] [--pocket-id Z] [--limit N] [--json]\` — newest-last terminal view.
- \`pocketpaw retrieval stats [--json]\` — counters.
- Registered as an early command alongside \`errors\` / \`logs\`.

## Test plan

- [x] 19 new tests in \`tests/cloud/test_retrieval_log.py\`:
  - Append path (dict + Pydantic + bad input)
  - 20-way concurrent appends preserving line integrity
  - Every filter axis (actor / source / pocket_id / time window)
  - Malformed-line tolerance
  - Missing-file safety
  - Tail ordering
  - Stats counters on empty + populated log
  - Singleton + env-var override
  - HTTP envelope, filter passthrough, stats endpoint
- [x] \`pytest tests/\` — 3991 passed (no regressions)
- [x] \`ruff check\` — clean
- [x] Smoke-tested \`uv run pocketpaw retrieval stats\`

## Pairs with

- [soul-protocol #161](https://github.com/qbtrix/soul-protocol/pull/161) — spec + emission
- PR-C (graduation policy) — reads this log
- PR-D (Why? drawer fourth tab) — surfaces traces inline

## Context

Design doc: paw-enterprise \`docs/enterprise/RETRIEVAL-TRACE.md\` (paw-enterprise PR #67).

### DIFF
--- a/ee/retrieval/__init__.py
+++ b/ee/retrieval/__init__.py
@@ -1,0 +1,8 @@
+# Retrieval log — paw-runtime sink for RetrievalTrace events from soul + kb + skills.
+# Created: 2026-04-13 (Move 4 PR-B) — JSONL appender, filtered reader, HTTP endpoint,
+# CLI tail. Consumers: graduation policy (PR-C), Why? drawer (PR-D), compliance export.
+
+from ee.retrieval.log import RetrievalLog, get_log
+from ee.retrieval.models import RetrievalLogEntry
+
+__all__ = ["RetrievalLog", "RetrievalLogEntry", "get_log"]
--- a/ee/retrieval/log.py
+++ b/ee/retrieval/log.py
@@ -1,0 +1,172 @@
+# ee/retrieval/log.py — Async-safe JSONL sink + filtered reader.
+# Created: 2026-04-13 (Move 4 PR-B) — Append-only file at
+# ~/.pocketpaw/retrieval.jsonl. Reads stream the file with optional filters.
+# Once line counts cross 10K-100K, swap the reader for an SQLite FTS index.
+
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+import os
+from collections.abc import AsyncIterator
+from datetime import datetime
+from pathlib import Path
+from typing import Any
+
+from ee.retrieval.models import RetrievalLogEntry
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_PATH = Path.home() / ".pocketpaw" / "retrieval.jsonl"
+
+
+class RetrievalLog:
+    """Async-safe JSONL sink for RetrievalTrace events.
+
+    Single-process serialisation via an asyncio.Lock — protects against
+    interleaved writes from concurrent recall callers in the same runtime.
+    Cross-process safety is intentionally NOT provided here (run paw-runtime
+    once per host); add file-locking later if multi-process emerges.
+    """
+
+    def __init__(self, path: str | Path | None = None) -> None:
+        self._path = Path(path) if path else DEFAULT_PATH
+        self._lock = asyncio.Lock()
+        self._path.parent.mkdir(parents=True, exist_ok=True)
+
+    @property
+    def path(self) -> Path:
+        return self._path
+
+    async def append(
+        self,
+        trace: Any,
+        *,
+        session_id: str | None = None,
+    ) -> RetrievalLogEntry:
+        """Append a trace to the log. ``trace`` may be a Pydantic model or a dict."""
+        trace_dict: dict[str, Any]
+        if hasattr(trace, "model_dump"):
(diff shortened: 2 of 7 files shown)


prakashUXtech (commented):
Solid foundation — the JSONL sink with asyncio.Lock, malformed-line tolerance, and the 20-way concurrent append test are all well done. Three things to fix before this merges.

**Blocking: datetime timezone mismatch**

`RetrievalLogEntry.ingested_at` is set with `datetime.now()` (naive, local time). FastAPI will parse a `since` or `until` query param as timezone-aware if the client sends ISO 8601 with a Z suffix (e.g. `?since=2026-04-13T10:00:00Z`). The comparison in `_matches()` at `ee/retrieval/log.py:150-152` then throws `TypeError: can't compare offset-naive and offset-aware datetimes`. Fix: use `datetime.now(tz=timezone.utc)` throughout and strip the tzinfo on comparison, or accept only naive UTC everywhere and document that constraint.

**Blocking: HTTP endpoints have no auth**

`ee/retrieval/router.py` defines two GET routes with no `Depends(current_user)` or equivalent. The routes aren't wired into `mount_cloud()` yet so they're not live, but that also means there's no PR that adds them to the auth surface — which means they could get mounted later without review. Either add the auth dependency now (matching the pattern in `ee/cloud/kb/router.py`) or add a prominent `# TODO: mount via mount_cloud() with current_user dep before exposing` comment so the next PR can't miss it.

**Non-blocking: PII in query strings**

User query text flows verbatim into `retrieval.jsonl` via the `trace` dict. That's intentional for observability, but there's no mention of it in the retention docs and no scrubbing path. Worth a note in the PR description or a `# NOTE: query strings may contain PII` comment near `_write()` so the compliance export (listed as a future consumer) knows to handle it.

**Minor nit**

`tail()` is correct but the semantics are easy to misread. `read(limit=n)` returns the n newest entries newest-first; reversing gives oldest-of-those-n first. A one-line docstring clarification ("returns up to n entries, oldest of those first") would help the next reader.

Test coverage is thorough — the filter axis tests, the singleton/env-override tests, and the HTTP fixture pattern are all solid.

prakashUXtech (commented):
test

prakashUXtech:

The RFC retires the retrieval log as a separate JSONL store. `retrieval.query` becomes a journal event, and the log surface turns into a filter view over the journal. Merging this ships a store that gets torn out in Phase 3.

Wait for the journal engine (Workstream A slice 2) to land, then rebuild the HTTP + CLI layers as filter views. The datetime tz and unauthed-endpoint blockers from review stop mattering in the new shape.

prakashUXtech:
Superseded by #954 — retrieval log + graduation rewritten as journal projection (same pattern as #953 for Fabric). All retrieval events now flow through the org journal; the HTTP + CLI surfaces live in ee/retrieval/router.py backed by projection queries. Closing; ship logic moved to #954.

PR #937, opened 2026-04-13
feat(ee): graduation policy over the retrieval log

prakashUXtech:
Generalises the bespoke 3×-same-path rule from \`correction_soul_bridge\` into a tier-promotion policy that scans the retrieval log and emits \`GraduationDecision\` objects when access counts cross thresholds. Apply mode is opt-in and uses \`soul.remember()\` for the higher-tier copy — no spec addition needed.

Move 4 PR-C. Stacks on [#936](https://github.com/pocketpaw/pocketpaw/pull/936).

### \`ee/graduation/models.py\`
- \`GraduationDecision\` (\`memory_id\`, \`kind\`, \`access_count\`, \`from_tier\`→\`to_tier\`, \`reason\`).
- \`GraduationReport\` (decisions + scan metadata + dry_run flag).

### \`ee/graduation/policy.py\`
- \`scan_for_graduations(log, *, window_days, episodic_threshold, semantic_threshold, actor, pocket_id, dry_run)\` — reads the log via \`ee/retrieval\`, counts per memory_id within the window, emits decisions when counts cross the threshold matching the candidate's tier. Last-seen tier wins so the report reflects current state.
- \`apply_decisions(soul, decisions)\` — writes higher-tier copies via \`soul.remember()\`, looks up original content via wide recall. Per-decision errors are logged and counted; never raises upward (graduation must not break the runtime). Soul without \`remember\`/\`recall\` is a no-op.
- Defaults: 30-day window, 10 accesses for episodic→semantic, 50 for semantic→core.
- \`MemoryType\` import is wrapped — when soul-protocol isn't installed in the dev env (optional dep), the apply path passes tier as a string. Production resolves to the real enum.

### \`ee/graduation/router.py\`
- \`POST /graduation/scan\` returns a \`GraduationReport\` (dry-run by default). Body tunes windows + thresholds + filters.
- \`POST /graduation/apply\` takes \`{ decisions: [...] }\` and returns \`{ applied, skipped }\`. 503 when no soul manager is loaded.

## Test plan

- [x] 12 new tests in \`tests/cloud/test_graduation_policy.py\`:
  - Empty log → no decisions
  - Episodic threshold crossing → \`episodic_to_semantic\`
  - Sub-threshold → no decision
  - Semantic threshold crossing → \`semantic_to_core\`
  - Outside-window entries excluded
  - Actor filter isolates counts
  - Apply success path with mocked soul (verifies \`remember\` called with target tier + bumped importance)
  - Apply skips missing memory
  - Apply swallows per-decision exceptions
  - Soul-without-apis returns empty
  - Scan endpoint returns decisions
  - Apply endpoint 503s when no soul manager
- [x] \`pytest tests/\` — 3991 passed (no regressions)
- [x] \`ruff check\` — clean

## Follow-up

Once both Move 1 PR-B (correction_soul_bridge) and this PR land in dev, a cleanup PR will collapse the bespoke 3×-same-path rule in \`_maybe_promote_to_procedural\` into a call to \`scan_for_graduations\` with \`window_days=30, episodic_threshold=3\`. Tracked separately so neither PR blocks the other.

## Pairs with

- soul-protocol [#161](https://github.com/qbtrix/soul-protocol/pull/161) — \`RetrievalTrace\` spec + emission
- pocketpaw [#936](https://github.com/pocketpaw/pocketpaw/pull/936) — sink + HTTP + CLI

## Context

Design doc: paw-enterprise \`docs/enterprise/RETRIEVAL-TRACE.md\` (paw-enterprise PR #67).

### DIFF
--- a/ee/graduation/__init__.py
+++ b/ee/graduation/__init__.py
@@ -1,0 +1,11 @@
+# Graduation — promote frequently-accessed soul memories to higher tiers.
+# Created: 2026-04-13 (Move 4 PR-C) — Reads the retrieval log written by
+# ee/retrieval (PR-B), counts accesses per memory_id within a window,
+# emits GraduationDecision objects when access crosses a threshold.
+# Apply mode uses soul.remember() to write the higher-tier copy; the
+# original entry is marked superseded via the spec's superseded field.
+
+from ee.graduation.models import GraduationDecision, GraduationReport
+from ee.graduation.policy import scan_for_graduations
+
+__all__ = ["GraduationDecision", "GraduationReport", "scan_for_graduations"]
--- a/ee/graduation/models.py
+++ b/ee/graduation/models.py
@@ -1,0 +1,45 @@
+# ee/graduation/models.py — GraduationDecision + report types.
+# Created: 2026-04-13 — Decisions describe the change without making it.
+# Apply mode reads decisions and mutates the soul; dry-run (default) just
+# emits them so an operator can review before promoting.
+
+from __future__ import annotations
+
+from datetime import datetime
+from typing import Literal
+
+from pydantic import BaseModel, Field
+
+GraduationKind = Literal["episodic_to_semantic", "semantic_to_core", "promote_procedural"]
+
+
+class GraduationDecision(BaseModel):
+    """One memory crossed an access threshold — propose a tier change."""
+
+    memory_id: str
+    actor: str = ""
+    pocket_id: str | None = None
+    kind: GraduationKind
+    access_count: int
+    window_days: int
+    from_tier: str | None = None
+    to_tier: str
+    reason: str = ""
+
+    def short(self) -> str:
+        """One-line summary for terminal output."""
+        from_label = self.from_tier or "?"
+        return (
+            f"[{self.kind}] {self.memory_id} {from_label}→{self.to_tier} "
+            f"({self.access_count} accesses in {self.window_days}d)"
+        )
+
+
+class GraduationReport(BaseModel):
+    """The output of one scan."""
+
+    decisions: list[GraduationDecision] = Field(default_factory=list)
+    scanned_entries: int = 0
+    window_days: int = 30
+    dry_run: bool = True
+    generated_at: datetime = Field(default_factory=datetime.now)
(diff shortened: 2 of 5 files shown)


prakashUXtech (commented):
The policy logic is sound — threshold semantics are clear, the dry_run default is the right call for a mutation-adjacent feature, and the exception-swallowing in apply_decisions with per-decision isolation is correct behavior (graduation must not kill the runtime). A few things need attention.

**Blocking: apply endpoint mutates the soul with no auth and no audit trail**

POST /graduation/apply writes new memories to the soul via soul.remember(). The router has no Depends(current_user) — same gap as the retrieval router in #936. More critically, apply_decisions() has no AuditEvent call. Soul mutation is exactly the kind of operation that should appear in audit.jsonl with severity WARNING. Add an AuditEvent.create() call in apply_decisions() for each successful graduation, mirroring the pattern in pocketpaw/security/audit.py. The auth dep can wait until mount_cloud() wires this in, but the audit log entry should be here from day one.

**Non-blocking: _lookup_memory_content is O(n) per decision**

_lookup_memory_content() does soul.recall("", limit=500) to find a single memory by ID. apply_decisions() calls it in a loop — 50 decisions means 50 full wide recalls. This is fine for the current volume but will get noticed at scale. A dict built from a single recall pass at the start of apply_decisions() would fix it. Flagging now so it doesn't get buried.

**Non-blocking: empty tier treated as episodic is too broad**

In _decide(), tier in {"episodic", ""} means any candidate with no tier field gets counted toward episodic promotion. The retrieval log can contain kb/fabric results that have no tier — those aren't soul memories and shouldn't graduate. Consider requiring an explicit tier field (skip if missing) rather than defaulting to episodic. The comment in the code acknowledges this but frames it as intentional; worth a follow-up issue at minimum.

**Stacking note**

This PR targets feat/retrieval-log-sink, not dev. It can't merge until #936 lands. That's the right call for a stacked PR — just noting it for the merge queue.

The coupling to #936 is clean: scan_for_graduations() takes a RetrievalLog as a parameter, tests stub it directly, and the graduation router imports only get_log(). No sink internals leak into the policy layer.

Test coverage for the happy paths, apply skip/swallow, and the 503 endpoint are all there. The window test (test_outside_window_is_ignored) is particularly thorough — writing backdated entries directly to the file is the right approach.

prakashUXtech:

Stacks on #936's JSONL log, so the data source disappears when the log becomes a journal filter. The missing-AuditEvent blocker from review closes by construction once graduation emits `graduation.applied` journal events instead of a side-channel audit.

Wait for the journal engine and the #936 rewire. The scoring and threshold logic ports over; the I/O surface gets replaced.

prakashUXtech:
Superseded by #954 — graduation policy now reads from retrieval.query projection, threshold/decision logic ported verbatim. apply_decisions emits graduation.applied journal events. Closing; ship logic moved to #954.

PR #938, opened 2026-04-13
feat(policy): scope on Fabric objects + visibility engine

prakashUXtech:
Adds the runtime side of the RBAC/ABAC primitive landed in [soul-protocol #162](https://github.com/qbtrix/soul-protocol/pull/162). FabricObject gains a \`scope: list[str]\` field, FabricQuery accepts an optional \`scopes\` filter, and a new \`ee/policy/\` module exposes the one function paw-runtime uses across every retrieval path.

Move 5 PR-B. Targets \`dev\`. PR-C (scope chip UI) and PR-D (E2E enforcement test) follow.

### \`ee/policy/engine.py\`
- \`visible(entity, user_scopes)\` — pure function, defensive about where scope lives (attribute / dict key / empty). Empty either side passes through.
- \`filter_visible(entities, user_scopes)\` — returns \`(kept, hidden_count)\` so paw-runtime can record how many entries got filtered into the retrieval log (ties Move 4 + Move 5).
- \`decide(entity, user_scopes)\` — \`PolicyDecision\` dataclass with \`reason\` field for audit explanations.
- \`DEFAULT_ALLOW_UNSCOPED\` toggle for tenants that want strict mode.
- Local \`_granted\` helper mirrors soul-protocol's \`match_scope\` semantics — no hard dep on soul-protocol being installed.

### \`ee/fabric/models.py\`
- \`FabricObject.scope\` (list[str], default empty).
- \`FabricQuery.scopes\` (caller's scope set; default empty = no filter).

### \`ee/fabric/store.py\`
- \`fabric_objects\` table gets a \`scope TEXT\` column with default \`'[]'\`.
- \`create_object\` accepts either positional (legacy: \`type_id\` + properties) OR a pre-built \`FabricObject\` instance, plus an optional \`scope=[]\` kwarg on the legacy path.
- \`_row_to_object\` decodes scope JSON defensively (malformed rows fall back to \`[]\`).
- \`query()\` applies \`filter_visible(objects, q.scopes)\` after the SQL fetch — policy logic in one place.

## Test plan

- [x] 22 new tests in \`tests/cloud/test_policy_engine.py\` covering:
  - \`visible\` / \`filter_visible\` / \`decide\` across FabricObject + dict + duck-typed dataclass + string scope
  - Default allow-unscoped behaviour + strict mode override
  - Exact match, hierarchical glob, no-overlap denial
  - \`FabricObject.scope\` round-trip through model_dump
  - \`FabricQuery.scopes\` default
  - End-to-end: \`FabricStore.query()\` with three objects (sales/finance/unscoped) and a sales-scoped caller — confirms the right subset surfaces
- [x] \`pytest tests/\` — 3991 passed, 0 failed
- [x] \`ruff check\` — clean

## Pairs with

- soul-protocol [#162](https://github.com/qbtrix/soul-protocol/pull/162) — scope on \`MemoryEntry\` + \`Soul.recall(scopes=...)\` + \`spec.match_scope\`
- PR-C — scope chip UI
- PR-D — Playwright E2E enforcement test

## Context

Design referenced in the fleet vision brainstorm. Compliance narrative ("show me who saw what" = SQL) becomes one query against the retrieval log + policy decisions.

### DIFF
--- a/ee/fabric/models.py
+++ b/ee/fabric/models.py
@@ -54,6 +54,10 @@ class FabricObject(BaseModel):
     source_id: str | None = None
     created_at: datetime = Field(default_factory=datetime.now)
     updated_at: datetime = Field(default_factory=datetime.now)
+    # Move 5 PR-B — RBAC/ABAC scope tags. Empty list = visible to anyone.
+    # Hierarchical glob: org:sales:* matches org:sales:leads. Filtered at
+    # query time before results reach the LLM.
+    scope: list[str] = Field(default_factory=list)
 
 
 class FabricLink(BaseModel):
@@ -77,6 +81,10 @@ class FabricQuery(BaseModel):
     link_type: str | None = None
     limit: int = 50
     offset: int = 0
+    # Move 5 PR-B — caller's effective scope set. Empty = no scope filter
+    # (sees everything). Populated by paw-runtime from session auth before
+    # forwarding to FabricStore.query().
+    scopes: list[str] = Field(default_factory=list)
 
 
 class FabricQueryResult(BaseModel):
--- a/ee/fabric/store.py
+++ b/ee/fabric/store.py
@@ -38,7 +38,8 @@ CREATE TABLE IF NOT EXISTS fabric_objects (
     source_connector TEXT,
     source_id TEXT,
     created_at TEXT DEFAULT (datetime('now')),
-    updated_at TEXT DEFAULT (datetime('now'))
+    updated_at TEXT DEFAULT (datetime('now')),
+    scope TEXT DEFAULT '[]'
 );
 
 CREATE TABLE IF NOT EXISTS fabric_links (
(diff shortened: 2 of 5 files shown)


prakashUXtech (commented):
## #938 — scope on Fabric objects + visibility engine

Good foundation. The engine logic is clean and the test suite is thorough — the duck-typed entity support and strict-mode tests in particular are exactly what you want for a primitive like this. Two blockers before merge.

---

### Blocker 1: no schema migration for existing databases

`_ensure_schema` runs `CREATE TABLE IF NOT EXISTS fabric_objects ...`. On an existing install that already has the table (without the `scope` column), SQLite will skip the CREATE entirely. The new `scope TEXT DEFAULT '[]'` column never gets added. The `INSERT` in `create_object` then fails with `no such column: scope`.

`_row_to_object` defensively handles the missing column via `row["scope"] if "scope" in row.keys()`, which is good — but SELECTs work fine; it's the writes that break.

Fix: add an `ALTER TABLE` step to `_ensure_schema` that adds the column if it doesn't exist:

```python
async with aiosqlite.connect(self._db_path) as db:
    await db.executescript(SCHEMA_SQL)
    # Migrate pre-scope installs
    try:
        await db.execute(
            "ALTER TABLE fabric_objects ADD COLUMN scope TEXT DEFAULT '[]'"
        )
        await db.commit()
    except Exception:
        pass  # Column already exists — SQLite has no IF NOT EXISTS for ALTER
```

SQLite doesn't support `ALTER TABLE ... ADD COLUMN IF NOT EXISTS`, so the try/except is the idiomatic way. The `DEFAULT '[]'` ensures existing rows get the right default on read without a backfill.

---

### Blocker 2: `total` is the pre-filter count

The `query()` method counts all matching rows in SQL, then applies `filter_visible()` in Python. The returned `FabricQueryResult.total` is the pre-filter number. Callers using `total` for pagination (including the REST client hitting `/fabric/query`) will compute wrong page counts.

If the SQL returns 100 rows and the policy filters to 40, a client with `limit=50` thinks there's a second page with the remaining 50, but that second page may return 0 visible results (or worse, leak that hidden rows exist via their absence).

Options:
- Set `total` to `len(objects)` after filtering (simplest; breaks `total` as "database count" semantics but fixes pagination).
- Add a `visible_total` field to `FabricQueryResult` alongside the raw `total` (clean, additive, doesn't break existing consumers).
- Push scope filtering into SQL (harder but correct at scale — avoids pulling all rows into Python to count).

The second option adds a field to the response model without removing anything, so it's backwards-compatible:

```python
class FabricQueryResult(BaseModel):
    objects: list[FabricObject]
    total: int           # pre-filter SQL count (unchanged for existing consumers)
    visible_total: int = 0  # post-filter count for pagination
```

---

### Minor: `DEFAULT_ALLOW_UNSCOPED` as a mutable module global

Setting this at startup is a process-wide side effect. Because `visible()` and `filter_visible()` both default to `DEFAULT_ALLOW_UNSCOPED`, mutating it from the outside flips the default for all concurrent requests. It also isn't wired into the config system — nothing in `config.py` reads or sets it.

Two things worth doing before this ships:

1. Add a `POCKETPAW_POLICY_ALLOW_UNSCOPED` setting in `config.py` (Pydantic Field, bool, default True).
2. Read that setting at startup and set `engine.DEFAULT_ALLOW_UNSCOPED` once, or pass `allow_unscoped=settings.policy_allow_unscoped` at the call sites in `store.py` rather than relying on the global default.

---

### Nit

The lazy import of `filter_visible` inside `query()` is fine for avoiding circular imports at module load time, but it's imported on every call. A top-of-file `TYPE_CHECKING` guard + deferred import or a module-level conditional import would be cleaner once the circular dep concern is resolved.

prakashUXtech:
**Verdict:** REDESIGN (see RFC #164)

The schema-migration blocker flagged in review (CREATE TABLE IF NOT EXISTS misses the new scope column on existing DBs) is fixed by construction when Fabric writes redirect to `fabric.object.*` journal events — scope normalization happens at journal write time, not per-store. The pagination blocker (total vs visible count) resolves at the journal filter layer. The decision model (visible/filter_visible/decide/granted) is correct and portable; the write path is what changes.

Wait for the journal engine, then rewrite FabricStore as a journal projection. Keep the decision logic; retire the sqlite table.

prakashUXtech:
Superseded by #953 — Fabric journal projection. The new PR rewrites Fabric as an append-only journal projection over soul-protocol's EventEntry primitives, which fixes both #938's blockers (CREATE TABLE IF NOT EXISTS schema migration bug + total-vs-visible pagination leak) by construction. Decision logic ported verbatim.

PR #939, opened 2026-04-13
feat(connectors): IngestAdapter alias + IngestACL for source-side scope inheritance

prakashUXtech:
Adds the lean ingest-side primitive sketched in the fleet platform brainstorm. \`ConnectorProtocol\` stays as-is for the existing 9 adapters; a new \`IngestAdapter\` Protocol extends it with one method (\`permissions\`) that reports source-side ACLs so paw-runtime tags ingested Fabric objects with the inherited scope before they land.

The result: a private Slack channel's messages stay private inside Single Brain without flattening permissions to "anyone with the connector."

Move 7 PR-A. PR-B adds the fleet template runtime that discovers \`IngestAdapter\` implementations and wires the install flow.

### \`src/pocketpaw/connectors/protocol.py\`
- \`IngestACL\` dataclass — \`scope\` (list[str]), \`visibility\`, \`source_principal\`, open \`metadata\` dict. Empty defaults so legacy adapters that don't care about per-record scope are unaffected.
- \`IngestAdapter\` Protocol — extends \`ConnectorProtocol\` with a single \`permissions(pocket_id, record_id=None) -> IngestACL\` method.
  - \`record_id=None\` → connector-wide defaults (a Stripe-account-wide read scope).
  - Named \`record_id\` → per-document ACLs for systems like Slack channels or HubSpot deal-team membership.

### \`src/pocketpaw/connectors/__init__.py\`
- Re-exports \`IngestACL\` + \`IngestAdapter\` alongside the existing public surface.

## Test plan

- [x] 11 new tests in \`tests/test_ingest_adapter.py\`:
  - \`IngestACL\` defaults, scope passthrough, open metadata
  - Reference duck-typed \`FakeIngestAdapter\` implementing both protocols
  - Default \`permissions\` returns public scope when \`record_id\` is None
  - Per-record \`permissions\` returns private scope for \`private_*\` ids
  - \`permissions\` records the call args (pocket_id + record_id) so PR-B's installer can verify ACL inheritance
  - Public re-exports point at the same objects
  - Static check: \`IngestAdapter\` exposes every \`ConnectorProtocol\` method + \`permissions\`
- [x] \`pytest\` — passing
- [x] \`ruff check\` — clean

## Pairs with

- pocketpaw [#938](https://github.com/pocketpaw/pocketpaw/pull/938) — \`ee/policy\` engine. \`IngestACL.scope\` feeds straight into \`FabricObject.scope\`, then \`ee/policy.filter_visible\` applies \`match_scope\` at retrieval time. End-to-end: a private Slack channel ingested by a SlackIngestAdapter lands tagged \`org:eng:eyes-only\` and only callers granted that scope ever see those messages.

## Follow-ups

- Update existing 9 connectors to implement \`permissions()\` — non-breaking; defaults to "no per-record ACL" if missing.
- PR-B (next): fleet template runtime + paw fleet install CLI.
- PR-C: Install Fleet UI flow.

### DIFF
--- a/src/pocketpaw/connectors/__init__.py
+++ b/src/pocketpaw/connectors/__init__.py
@@ -7,6 +7,8 @@ from pocketpaw.connectors.protocol import (
     ActionSchema,
     ConnectionResult,
     ConnectorProtocol,
+    IngestACL,
+    IngestAdapter,
     SyncResult,
 )
 from pocketpaw.connectors.registry import ConnectorRegistry
@@ -16,6 +18,8 @@ __all__ = [
     "ConnectionResult",
     "ActionSchema",
     "ActionResult",
+    "IngestACL",
+    "IngestAdapter",
     "SyncResult",
     "ConnectorRegistry",
 ]
--- a/src/pocketpaw/connectors/protocol.py
+++ b/src/pocketpaw/connectors/protocol.py
@@ -1,5 +1,10 @@
 # ConnectorProtocol — interface for all data source adapters.
 # Created: 2026-03-27 — Protocol-based, async, adapter-agnostic.
+# Updated: 2026-04-13 (Move 7 PR-A) — Added IngestACL + IngestAdapter
+#   alias so the ingest side of the protocol carries source-side ACLs
+#   into Fabric. Adapters that emit documents tagged with inherited
+#   scope keep org permissions intact end-to-end (a private Slack
+#   channel's messages stay private inside Single Brain).
 
 from __future__ import annotations
 
@@ -70,6 +75,29 @@ class SyncResult:
     duration_ms: float = 0
 
 
+@dataclass
+class IngestACL:
+    """Source-side access control list inherited into Single Brain.
+
+    When an adapter pulls a document from a tenant system that already has
+    ACLs (a private Slack channel, a HubSpot deal restricted to one team,
+    a Notion page shared with a sub-list), the adapter reports those ACLs
+    here so paw-runtime tags the resulting Fabric object with the matching
+    scope tags before it lands. The brain inherits the source's permissions
+    instead of flattening them to "everyone with access to the connector."
+
+    Empty fields are intentional defaults — most ingests are global to the
+    pocket and need no per-record scoping. When ``scope`` is non-empty,
+    the runtime applies it as the Fabric object's ``scope`` list verbatim;
+    when ``visibility`` is set, callers can also surface a UI label.
+    """
+
+    scope: list[str] = field(default_factory=list)
+    visibility: str = ""  # "public" | "private" | "members" | "owner"
+    source_principal: str = ""  # The original ACL holder, e.g. "channel:#founders"
+    metadata: dict[str, Any] = field(default_factory=dict)
+
+
 class ConnectorProtocol(Protocol):
     """Interface for all connector adapters.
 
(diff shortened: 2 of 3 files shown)


prakashUXtech (commented):
## #939 — IngestAdapter + IngestACL for source-side scope inheritance

Clean, well-contained primitive. No blockers — the protocol boundary is correct, backwards compat is solid (existing adapters don't need to change until PR-B), and the tests cover the contract shape well. A few things to tighten up.

---

### Finding 1: `visibility` is an unvalidated free-form string

`IngestACL.visibility` is typed as `str` with a comment listing valid values: `"public" | "private" | "members" | "owner"`. Nothing enforces this. An adapter returning `"restricted"` or `"team-only"` will be accepted silently and downstream UI code in PR-C will get a value it doesn't know how to render.

Since `ConnectorStatus` and `TrustLevel` in the same file already use `StrEnum`, the pattern is established:

```python
class IngestVisibility(StrEnum):
    PUBLIC = "public"
    PRIVATE = "private"
    MEMBERS = "members"
    OWNER = "owner"
```

Then `IngestACL.visibility: IngestVisibility | str = ""` keeps it backwards-compatible for adapters that want to pass custom strings, while making the common values explicit and type-checkable.

---

### Finding 2: no audit trail when scope is inherited

When the fleet template runtime (PR-B) calls `permissions()` and tags a `FabricObject` with the returned scope, there's no hook here to log it. For a system whose compliance narrative is "show me who saw what," that attribution needs to exist at ingestion time, not just at retrieval time via `decide()`.

Not a blocker for this PR since the actual wiring is PR-B's responsibility, but worth documenting as a PR-B requirement: the installer should call `ee.policy.decide()` and write an `AuditEvent` when scope is applied to an ingested object. A TODO comment in `IngestACL` or `IngestAdapter` pointing to this would make it harder to miss.

---

### Nit: structural check in tests instead of `@runtime_checkable`

The test comment acknowledges that `isinstance(adapter, ConnectorProtocol)` won't work because the Protocol isn't `@runtime_checkable`. The `hasattr` fallback is fine for now. If the fleet runtime (PR-B) needs to discover `IngestAdapter`-capable connectors at runtime — the PR description says it will — you'll want `@runtime_checkable` on `IngestAdapter` (not necessarily `ConnectorProtocol`). Worth adding now so PR-B doesn't have to reach back and modify this file.

```python
from typing import runtime_checkable

@runtime_checkable
class IngestAdapter(ConnectorProtocol, Protocol):
    ...
```

---

### Nit: `source_principal` has no format contract

`IngestACL.source_principal` carries strings like `"channel:#founders"` or `"hubspot:deal-team:42"`. Fine for now, but if `decide()` audit logs ever surface this to operators, a free-form string is hard to parse programmatically. Consider documenting the expected format (`system:resource:id`) in the docstring, even if it stays a plain string.

prakashUXtech:
Updating earlier review: the Org Architecture RFC (soul-protocol #164) has now merged as part of v0.3.1. This PR's scope (IngestAdapter + IngestACL for copy-on-ingest) is stable direction — it provides the base class that concrete connectors (Drive, Salesforce, etc.) will extend. The SourceAdapter Protocol for live/Zero-Copy retrieval lives in soul-protocol and is complementary, not overlapping.

Lint fix applied in latest commit. Ready to merge once CI is green.

Issue #73, opened 2026-02-13
Add gemini integration in it for text becoause claude and gpt in expensive for student hence few user will use this


ANSHSINGH050404:
### Motivation

PocketPaw currently relies on Claude/GPT style agent backends which are expensive for students and hobby developers.
Because of the high API cost, many users cannot run the assistant continuously or experiment with it locally.

Google Gemini provides a generous free tier and significantly cheaper pricing, making the project more accessible — especially for learners and contributors from developing regions.

### Proposal

Add a new agent backend implementation using the Google Gemini API that integrates into the existing `agent_backend` abstraction.

The backend should:

* Support chat text generation
* Follow the same interface as existing backends
* Be selectable via configuration
* Work with memory and message bus without changes
* Appear in `/api/health` diagnostics as the active backend

### Example configuration

```
AGENT_BACKEND=gemini
GEMINI_API_KEY=your_api_key_here
GEMINI_MODEL=gemini-1.5-flash
```

### Expected behavior

* Users can switch providers without code changes
* Local development becomes affordable
* Students can run long-lived agents
* Improves project adoption and testing

### Additional Notes

This change does not replace existing providers — it only adds an optional backend.

I would like to work on implementing this feature.

ANSHSINGH050404:
@prakashUXtech  Please assign me this It will help us to gain more users

prithvirajz:
I am working on this.

prakashUXtech:
Moving to backlog — we're focused on installation reliability and bug fixes for v0.3. See #115 for how to contribute during stabilization. We'll revisit feature work once the install path is solid across all platforms.

PR #940, opened 2026-04-13
feat(fleet): installable bundle runtime + Sales Fleet template

prakashUXtech:
A FleetTemplate is a YAML manifest naming a soul template + pocket + connector list + scope tags. \`install_fleet()\` orchestrates the install using existing primitives (\`SoulFactory\`, \`ConnectorRegistry\`, pocket service) — no new runtime concepts. Sales Fleet ships as the first bundled example.

Move 7 PR-B. Stacks on [#939](https://github.com/pocketpaw/pocketpaw/pull/939). PR-C ships the Install Fleet UI.

### \`ee/fleet/models.py\`
- \`FleetConnector\` — name + config + optional flag.
- \`FleetTemplate\` — name, display_name, version, soul_template ref, pocket name + widgets, connector list, scope list, open metadata.
- \`FleetInstallStep\` + \`FleetInstallReport\` — per-step status (succeeded / skipped / failed) so partial installs are observable.

### \`ee/fleet/installer.py\`
- \`load_fleet(path_or_name)\` — reads YAML/JSON or resolves a bundled name (\`sales-fleet\` → \`src/pocketpaw/fleet_templates/sales-fleet.yaml\`).
- \`install_fleet(fleet, *, soul_factory, connector_registry, pocket_creator)\` — pure orchestrator. Each external dep is injectable; tests substitute fakes.
- Per-step exceptions are caught + logged + marked failed in the report so install never crashes the runtime.
- Optional connectors → "skipped" when missing; required → "failed" so admins see what to fix.
- Returns early on soul creation failure (no orphan pocket).

### \`src/pocketpaw/fleet_templates/sales-fleet.yaml\`
- Arrow soul + Pipeline pocket + HubSpot + Gong, scoped \`org:sales:*\`. Connectors marked optional so the demo install works without external API keys.

## Test plan

- [x] 15 new tests in \`tests/cloud/test_fleet_installer.py\` covering YAML + JSON manifest loading, bundled-by-name resolution, missing-file error, end-to-end orchestration with mocked deps, optional vs required connector behaviour, per-step exception capture, early return on soul failure, Sales Fleet bundled contract, report helpers.
- [x] \`pytest\` — passing
- [x] \`ruff check\` — clean

## Pairs with

- pocketpaw [#939](https://github.com/pocketpaw/pocketpaw/pull/939) — IngestAdapter + IngestACL. Future fleets can declare \`scope_inheritance: source\` on a connector to have ingested records adopt the source's ACLs verbatim.
- soul-protocol [#163](https://github.com/qbtrix/soul-protocol/pull/163) — soul template library that Sales Fleet's \`soul_template: arrow\` resolves against.

## Follow-ups

- More bundled fleets (Content Fleet → Flash, Recruiting Fleet → Cyborg, Research Fleet → Analyst).
- \`paw fleet install <name>\` CLI subcommand.
- PR-C: Install Fleet UI in paw-enterprise.


prakashUXtech (commented):
Good structure overall. The injectable-deps pattern in install_fleet makes this easy to test and keeps the orchestrator clean. 15 tests with solid coverage of partial failures and the optional/required connector distinction. A few things to address:

**load_fleet path handling**

When path_or_name is a string and the bundled lookup misses, the function falls through to Path(path_or_name) with no canonicalization. In an operator CLI context the blast radius is small, but if this ever gets wired to a REST endpoint or user-facing input, ../../../ traversal works as-is. Suggest adding a p.resolve() check or at minimum a note in the docstring that callers are responsible for sanitizing paths.

**PyYAML error message points at the wrong package**

ee/fleet/installer.py line ~55:
```
"Install with `pip install soul-protocol[engine]`."
```
PyYAML is a transitive dep of pocketpaw[soul], not soul-protocol directly. The user-facing message should say `pip install pocketpaw[soul]`.

**Minor: installer.py re-imports FleetInstallReport and FleetInstallStep from models via __init__.py**

The __init__.py re-exports FleetInstallReport and FleetInstallStep from installer.py, but those types are defined in models.py. Installer.py imports them from models.py, so the public API is consistent -- just note that the __init__ re-exports installer symbols that themselves wrap model symbols. Fine as-is, but if models grows it may be cleaner to re-export directly from models.

Verdict: minor nits, nothing blocking. Can merge after fixing the pip install message.

prakashUXtech (commented):
Good structure overall. The injectable-deps pattern in install_fleet makes this easy to test and keeps the orchestrator clean. 15 tests with solid coverage of partial failures and optional/required connector distinctions.

Three things to address before merge:

**load_fleet path handling (P2)**

When path_or_name is a string and the bundled lookup misses, it falls through to Path(path_or_name) without canonicalization. Traversal sequences like ../../../etc work as-is. Low risk for an operator CLI, but if this ever reaches a REST endpoint, it is a problem. Add p.resolve() before the exists() check, or document in the docstring that callers own sanitization.

**PyYAML error message points at the wrong package (P2)**

installer.py ~line 55: the ImportError tells users to run pip install soul-protocol[engine], but PyYAML lands via pocketpaw[soul]. Change the message to pip install pocketpaw[soul] so operators do not go chasing the wrong package.

**Dead constant DEFAULT_FADE_THRESHOLD in... wait, that is in 941. Disregard here.**

Minor: the __init__ re-exports FleetInstallReport and FleetInstallStep from installer.py even though those types live in models.py. Works fine, but the indirection is a bit confusing. Not blocking.

Overall: minor nits only. LGTM after fixing the error message.

prakashUXtech (commented):
Updating earlier review now that #939 is merged and this rebased cleanly on dev.

Scope check: the Org Architecture RFC (soul-protocol #164) shipped in v0.3.1 and the `soul org init --fleet` stub (#170) is already in place waiting for this PR, so the direction is stable — not a moving target anymore.

Re-reviewed the diff on the rebased branch:

- **PyYAML import-error message**: fixed on this branch. Error now points at `pocketpaw[soul]` (the extra that actually owns the PyYAML dep via `soul-protocol[engine]`), so operators who hit the error run the right install.
- **`load_fleet` path canonicalization**: still unguarded, still marked as a P2 on the original review. Consensus was this is operator-CLI only so the blast radius is bounded; if a REST surface picks up `load_fleet` later the caller is on the hook for sanitization. Fine to land and follow up when/if a REST route actually exposes it.
- **`__init__` re-export of `FleetInstallReport` / `FleetInstallStep`**: still goes through `installer.py` even though the types live in `models.py`. Non-blocking, just a nit.

Coverage on the 15 tests is solid — partial-failure reporting, required-vs-optional connector distinction, early return on soul failure, bundled Sales Fleet contract. Ruff clean. Full suite green locally (4002 passed, 51 skipped).

One thing to flag for follow-up work, not this PR: the installer doesn't emit `agent.spawned` journal events when it creates a soul, so the onboarding wizard stubs in #170 won't see the install happen through the journal yet. Worth a small follow-up to wire that in once the fleet install lands and the paw-enterprise #73 UI is pulling from the journal.

LGTM. Merging.

PR #941, opened 2026-04-13
feat(widget): graduation engine for UI elements

prakashUXtech:
Same primitive as memory graduation ([ee/graduation in Move 4 PR-C](https://github.com/pocketpaw/pocketpaw/pull/937)), applied to widgets. Captures user interaction with widgets in a JSONL log, scans the log to propose pin / fade / archive decisions.

This is the captain's "intelligent UI" thesis materialising as a concrete primitive — UI that reacts to use the same way memory does. The UI consumer (Move 8 PR-C) renders the proposals as Suggested Layout chips inside pocket chat.

Move 8 PR-A. PR-B adds agent widget suggestions over retrieval patterns. PR-C ships the UI surfaces.

### \`ee/widget_graduation/models.py\`
- \`WidgetInteraction\` — append-only log entry: widget_id, pocket_id, actor, action (\`open\` / \`edit\` / \`click\` / \`dismiss\` / \`remove\`), timestamp.
- \`WidgetDecision\` — proposed UI mutation: \`pin\` / \`fade\` / \`archive\` verdict + interaction count + window + reason.
- \`WidgetReport\` — full scan output.

### \`ee/widget_graduation/log.py\`
- \`WidgetInteractionLog\` — async-safe JSONL sink mirroring \`ee/retrieval/log.py\`.
- Filtered read (widget_id / pocket_id / actor / since / limit).
- Process singleton via \`get_widget_log()\` with \`POCKETPAW_WIDGET_LOG\` env override.

### \`ee/widget_graduation/policy.py\`
- \`scan_for_widget_decisions(log, *, window_days, pin_threshold, archive_days, pocket_id, actor)\` — three-verdict policy.
- **Pin** — widget hit \`pin_threshold\` promoting interactions in window. Promoting = open / edit / click. Dismiss + remove never promote.
- **Fade** — widget seen in archive window but no promoting hits in pin window.
- **Archive** — widget last touched more than \`archive_days\` ago.
- Defaults: 30-day pin window, 10 promoting interactions, 60-day archive ceiling.

## Test plan

- [x] 12 new tests in \`tests/cloud/test_widget_graduation.py\`:
  - Append + filtered read across pocket + widget axes
  - Concurrent appends preserve line integrity
  - Missing-file returns empty
  - Pin fires at exact threshold, sub-threshold does not
  - Only promoting actions count toward pin
  - Recent history with zero in-window use → fade
  - Old inactive widget → archive
  - Pocket + actor filters isolate decisions
- [x] \`pytest\` — passing
- [x] \`ruff check\` — clean

## Pairs with

- pocketpaw [#937](https://github.com/pocketpaw/pocketpaw/pull/937) — same memory-graduation primitive applied to a different signal source. The two scans share zero code by design but have identical shape (JSONL log → policy → decisions) so a future cleanup could factor a shared helper if patterns repeat across more signal types.

### DIFF
--- a/ee/widget_graduation/__init__.py
+++ b/ee/widget_graduation/__init__.py
@@ -1,0 +1,23 @@
+# Widget Graduation — same primitive as memory graduation, applied to widgets.
+# Created: 2026-04-13 (Move 8 PR-A) — Widgets a user opens often get pinned;
+# widgets ignored for a window get archived. Reads from the widget interaction
+# log and emits WidgetDecision objects (analogous to GraduationDecision in
+# ee/graduation). The captain's "intelligent UI" thesis materialises here:
+# the same memory-tier promotion mechanism, applied to UI elements.
+
+from ee.widget_graduation.log import WidgetInteractionLog, get_widget_log
+from ee.widget_graduation.models import (
+    WidgetDecision,
+    WidgetInteraction,
+    WidgetReport,
+)
+from ee.widget_graduation.policy import scan_for_widget_decisions
+
+__all__ = [
+    "WidgetDecision",
+    "WidgetInteraction",
+    "WidgetInteractionLog",
+    "WidgetReport",
+    "get_widget_log",
+    "scan_for_widget_decisions",
+]
--- a/ee/widget_graduation/log.py
+++ b/ee/widget_graduation/log.py
@@ -1,0 +1,106 @@
+# ee/widget_graduation/log.py — Append-only JSONL widget interaction log.
+# Created: 2026-04-13 (Move 8 PR-A) — Same JSONL pattern as ee/retrieval/log,
+# scoped to widget events. Lives at ~/.pocketpaw/widget-interactions.jsonl
+# (override via POCKETPAW_WIDGET_LOG). Read by the graduation policy.
+
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+import os
+from collections.abc import AsyncIterator
+from datetime import datetime
+from pathlib import Path
+
+from ee.widget_graduation.models import WidgetInteraction
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_PATH = Path.home() / ".pocketpaw" / "widget-interactions.jsonl"
+
+
+class WidgetInteractionLog:
+    """Async-safe append + streaming read for widget interactions."""
+
+    def __init__(self, path: str | Path | None = None) -> None:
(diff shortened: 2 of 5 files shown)


prakashUXtech (commented):
The graduation primitive translates cleanly from memory to widgets. JSONL log pattern is consistent with the rest of ee/, the asyncio.Lock + asyncio.to_thread write path is correct, and the policy logic is easy to follow.

Three things worth fixing before this lands:

**CI is UNSTABLE on this PR**

The merge state shows UNSTABLE. Confirm the failing check before merging -- widget graduation stacks nothing else yet but 942 depends on this branch.

**Three separate datetime.now() calls in scan_for_widget_decisions (P2 correctness)**

policy.py computes since, long_ago, and archive_window_start with three separate datetime.now() calls. In production these fire within microseconds, so it is not a real bug. In fast-running tests with mocked time they can diverge. Snapshot once at the top of the function:

now = datetime.now()
since = now - timedelta(days=window_days)
long_ago = now - timedelta(days=archive_days)
archive_window_start = now - timedelta(days=archive_days * 2)

**Widgets with sub-threshold promoting hits get no verdict (P2, document or decide)**

A widget with 7 opens in the window (below the default pin_threshold of 10) lands in promoting_in_window, so the fade/archive loop skips it entirely. It does not pin, fade, or archive -- it just drops from the report. This may be intentional (do not disturb an active widget), but it is not documented anywhere. Either add a comment in the policy or emit a dedicated verdict like ignore so the consumer knows the widget was evaluated.

**Dead constant DEFAULT_FADE_THRESHOLD = 0 (nit)**

It is exported and never read by the policy. Remove it or wire it into the logic if there is a future plan for it.

Overall: LGTM with the above addressed. The test suite covers the right cases -- pin-at-threshold, sub-threshold no-pin, non-promoting actions, fade, archive, pocket/actor filters.

prakashUXtech:

Reads widget interaction data from the same JSONL log pattern that #936 retires. Graduation decisions should emit to the journal as events for audit and replay, so both I/O sides change.

Wait for the journal engine and the widget-interaction event namespace (candidate: `widget.interaction.recorded`). Core heuristic logic ports; data source replaces.

prakashUXtech:
Superseded by #955 — widget graduation engine rewritten as journal projection. Pin/fade/archive thresholds + confidence ramp ported verbatim. All widget events now flow through the org journal (widget.interaction.recorded + widget.graduated). Closing; ship logic moved to #955.

PR #942, opened 2026-04-13
feat(widget): co-occurrence detector for agent widget suggestions

prakashUXtech:
Reads the retrieval log written by [ee/retrieval (Move 4 PR-B)](https://github.com/pocketpaw/pocketpaw/pull/936), groups queries from the same actor + pocket into 15-minute sessions, and surfaces \`SuggestedWidget\` proposals when a query pair recurs above a threshold. The agent feeds these proposals into the Instinct propose flow so the operator sees them as actions in The Tray rather than auto-applied UI mutations.

Move 8 PR-B. Stacks on [#941 PR-A](https://github.com/pocketpaw/pocketpaw/pull/941). PR-C ships the UI surfaces.

### \`ee/widget_suggestions/models.py\`
- \`PatternMatch\` — one detected co-occurring pair (signature, queries, count, pocket_id, recent actors).
- \`SuggestedWidget\` — id, pocket_id, title, description, widget_type, pattern, confidence, timestamp.
- \`SuggestionReport\` — full output of one detection pass.

### \`ee/widget_suggestions/detector.py\`
- \`detect_co_occurrence_patterns(log, *, window_days, threshold, pocket_id)\` — pure async function over any log that satisfies the ee/retrieval \`read\` shape.
- Sessions form when consecutive queries from the same actor + pocket are <15 minutes apart.
- Token-bag normalisation sorts within each query so word-order variants ("renewal discount" vs "discount renewal") collide and don't produce nonsense pair suggestions.
- Confidence scales \`count / (threshold * 3)\` capped at 1.0.
- Defaults: 7-day window, 3 co-occurrences.

## Test plan

- [x] 8 new tests in \`tests/cloud/test_widget_suggestions.py\`:
  - Pair above threshold → suggestion
  - Below threshold → no suggestion
  - Word-order variants suppressed (same question phrased two ways shouldn't suggest a side-by-side widget)
  - 60-min gap breaks session boundary
  - Different actors don't co-occur
  - Pocket filter isolates
  - Report carries window + threshold
  - Confidence saturates at 1.0
- [x] \`pytest\` — passing
- [x] \`ruff check\` — clean

## Pairs with

- pocketpaw [#941](https://github.com/pocketpaw/pocketpaw/pull/941) — widget graduation. Graduation reshuffles existing widgets; suggestions create new ones. Both feed the same "intelligent UI" surface in PR-C.

### DIFF
--- a/ee/widget_suggestions/__init__.py
+++ b/ee/widget_suggestions/__init__.py
@@ -1,0 +1,19 @@
+# Widget Suggestions — agent reads retrieval patterns and proposes new widgets.
+# Created: 2026-04-13 (Move 8 PR-B) — Companion to widget graduation: graduation
+# moves existing widgets up/down based on use, suggestions create NEW widgets
+# from observed retrieval co-occurrence ("user asks about X then Y three times
+# in a row → propose a widget that surfaces both side-by-side").
+
+from ee.widget_suggestions.detector import detect_co_occurrence_patterns
+from ee.widget_suggestions.models import (
+    PatternMatch,
+    SuggestedWidget,
+    SuggestionReport,
+)
+
+__all__ = [
+    "PatternMatch",
+    "SuggestedWidget",
+    "SuggestionReport",
+    "detect_co_occurrence_patterns",
+]
--- a/ee/widget_suggestions/detector.py
+++ b/ee/widget_suggestions/detector.py
@@ -1,0 +1,204 @@
+# ee/widget_suggestions/detector.py — Find co-occurring query patterns.
+# Created: 2026-04-13 (Move 8 PR-B) — Lean detector that reads the
+# retrieval log written by ee/retrieval (Move 4 PR-B), groups consecutive
+# queries from the same actor, and surfaces pairs that repeat above a
+# threshold within a window. The agent uses these as input to propose
+# new widgets via the existing Instinct propose flow.
+
+from __future__ import annotations
+
+import logging
+import re
+from collections import defaultdict
+from collections.abc import Iterable
+from datetime import datetime, timedelta
+from typing import Any
+
+from ee.widget_suggestions.models import (
+    PatternMatch,
+    SuggestedWidget,
+    SuggestionReport,
+)
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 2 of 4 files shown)


prakashUXtech (commented):
Clean implementation. The session-boundary logic is correct, the structural duck-typing via _entry_attr handles both dataclass and dict log entries well, and the test double (_FakeLog) is the right pattern for this kind of protocol-shaped dependency.

One correctness issue to fix:

**_normalise_query truncates before sorting (P1)**

detector.py line 166:
  tokens = _TOKEN_RE.findall(query.lower())
  return " ".join(sorted(tokens[:6]))

The slice tokens[:6] happens before sorted(). That means two queries that share a 7-token stem but differ only in the first six positions will normalise to different signatures even if they are semantically identical. The slice should come after sorting so the most significant tokens (alphabetically first) survive:

  return " ".join(sorted(tokens)[:6])

This is load-bearing for the deduplication guarantee the PR description promises. The existing test for word-order normalization still passes because that test uses short queries (under 6 tokens), so the bug is not caught.

**Nits**

The confidence formula min(1.0, count / (threshold * 3)) saturates at count = 9 for the default threshold of 3. That is a steep ramp -- a pair seen 4 times gets 0.44, while 9 times gets 1.0. Whether that scale is right depends on product intent, but it is worth a comment explaining the cap.

_unordered_pairs emits O(n^2) pairs per session. For the typical 2-3 query session this is fine. If sessions ever grow large (e.g., an automated agent running many queries), a session size cap would prevent a quadratic blowup. Not blocking, but worth a TODO.

Overall: one real bug in the truncation order. Fix that and this is good to merge.

prakashUXtech:

Reads from the retrieval log. With #936 held, this has no current data source. The `sorted(tokens)[:6]` correctness bug from review should still get fixed in the eventual rewrite.

Wait for the journal engine and the widget-interaction event shape (same dependency as #941).

prakashUXtech:
Superseded by #955 — co-occurrence detector rewritten as journal projection. The sorted(tokens[:6]) truncate-before-sort bug is fixed (now sorted(tokens)[:6]) with a regression guard test covering 8-token rotated inputs. Closing; ship logic moved to #955.

PR #951, opened 2026-04-16
ee-cloud: MongoDB memory backend + unified chat persistence

DevRohit06:
## Summary
- Replace file-backed pocket session storage with a MongoDB-backed `MemoryStoreProtocol` implementation that becomes the default memory backend in ee.
- Unify pocket and group messages into a single `messages` collection with a `context_type` discriminator.
- Fix the agent dispatch precedence bug that caused multiple `auto` agents to all reply when one was mentioned.

## What changed
- **Schema**: `Session` and `Message` get a `context_type: Literal["pocket","group"]` discriminator + `model_validator`. `Message` also gains `session_key`, `role`, and `workspace_id` for pocket rows. New `MemoryFactDoc` (`memory_facts` collection) holds LONG_TERM/DAILY entries with `workspace_id` + `user_id` indexes.
- **`ee/cloud/memory/`**: new module — `MongoMemoryStore` (full protocol + tenant-scoped helpers), `documents.py`, `bootstrap.register_default_backend()` which is called from `init_cloud_db` AFTER `init_beanie` and swaps `._store` in place on the existing `MemoryManager` singleton so cached refs (e.g. `agent_loop.memory`) auto-pick up Mongo.
- **`chat_persistence.py` rewritten**: `chat_id` IS the `Session.sessionId`; messages land as `context_type="pocket"` keyed by `session_key`. Fake-group auto-creation is gone.
- **`MongoMemoryStore` normalises** bus-style keys (`websocket:X`) to UI-style (`websocket_X`) on read/write so agent-loop writes line up with `Session.sessionId` for history reads. Also derives `sender_type` from `role` so assistant rows aren't stored as `sender_type="user"`.
- **Workspace scoping**: every persisted row stamps a `workspace_id` (resolved from `Session.workspace` for pocket rows, `entry.metadata["workspace_id"]` for facts). New `get_session_in_workspace` and `list_facts_in_workspace` adapter methods enforce a workspace filter at read time; protocol-level methods stay tenant-agnostic to preserve the OSS contract.
- **`SessionService.get_history`** reads from the unified messages collection — no file-memory fallback. `sessions/router` delegates to the service.
- **Agent dispatch** (`agent_bridge._should_agent_respond`): `silent` always opts out; if any agent is mentioned, only mentioned agents reply (regardless of mode); otherwise fall back to per-agent mode.
- **Observability**: `logger.debug` → `logger.exception`/`logger.warning` in chat_persistence + websocket_adapter so silent persistence failures surface in dev.

## Tests
- New `tests/cloud/memory/` suite (75 tests): protocol conformance, discriminator validators, session/facts paths, backend selection, workspace isolation, OSS isolation.
- New `tests/cloud/test_api_contracts.py` snapshots the response shapes for chat/groups and sessions endpoints so the schema rewrite is a storage-only change for clients.
- Fixture uses `mongomock-motor` so the suite runs in CI without a real MongoDB service. (Wraps `database.list_collection_names` to drop the `authorizedCollections` kwarg Beanie >=1.26 passes.)
- `scripts/smoke_*.py` and `scripts/diag_*.py` exercise each end-to-end path against a real local MongoDB.

## Test plan
- [x] `uv run pytest tests/cloud/memory/ tests/cloud/test_api_contracts.py` (75 tests, runs on mongomock-motor)
- [x] `uv run python scripts/smoke_mongo_memory.py` — protocol round-trips
- [x] `uv run python scripts/smoke_session_then_chat.py` — POST /sessions → chat → /history
- [x] `uv run python scripts/smoke_pocket_session.py` — sidebar pocket session flow
- [x] `uv run python scripts/smoke_chat_stream_keys.py` — bus-key normalization
- [x] `uv run python scripts/smoke_manager_reference.py` — cached MemoryManager picks up new store
- [x] `uv run python scripts/smoke_dm_and_group.py` — human DM + group end-to-end
- [x] `uv run python scripts/smoke_agent_dispatch.py` — mention precedence
- [ ] Reviewer: full `uv run pytest tests/cloud/` against local Mongo (HEAD baseline + this branch should match on pre-existing failures only)

## Notes
- Clean-slate assumption: no migration; existing dev DB can be dropped.
- Pairs with the paw-enterprise PR that switches the desktop client off the runtime API to the cloud sessions API.
- Auto-generated wiki under `docs/wiki/` is intentionally **not** included; the kb-rebuild hook will regenerate it on the merge commit.


prakashUXtech (changes requested):
**Verdict:** Request changes — several blocking issues, but the spec and implementation quality are genuinely good

**What it does:** Replaces file-backed pocket session storage in the ee cloud deployment with a MongoDB-backed MemoryStoreProtocol implementation. Unifies pocket and group messages into a single messages collection using a context_type discriminator. Also fixes an agent dispatch bug where multiple auto agents all replied when one was mentioned.

**Blocking issues:**

1. `ee/cloud/memory/mongo_store.py` — No workspace scoping on any query. The SPEC itself flags this (§9.2): "No workspace scoping at the adapter layer — callers must namespace their session_key and user_id across tenants." In a multi-tenant ee deployment, MongoMemoryStore.get_session() returns entries from any tenant whose session_key matches. This is a data isolation hole that should be resolved before merge, not deferred. Either add workspace_id to MemoryFactDoc and all queries now, or explicitly block this PR from running in multi-tenant mode until the follow-up lands.

2. `tests/cloud/memory/conftest.py` — The test fixture uses a real MongoDB on localhost (mongodb://localhost:27017). The SPEC says mongomock-motor, but the implementation uses a real connection. Tests will silently pass locally for anyone with Mongo running and silently fail (or not run at all) in CI where there is no Mongo service defined. CI for this PR shows only check-quality and security-scan — no test run against the new tests/cloud/memory/ suite. Add a mongomock-motor fixture as described in the SPEC, or add a MongoDB service container to the CI job.

3. `ee/cloud/memory/bootstrap.py:register_default_backend()` — This function writes directly to os.environ and then mutates _mm._manager._store in place. If this is called before Beanie is initialized (which init_cloud_db suggests is the case), MongoMemoryStore.save() will attempt Beanie document inserts without an active Motor connection and crash with an unhelpful error. The bootstrap function needs a guard that verifies Beanie is initialized before installing the store, or the store needs to defer its first Mongo call until Beanie is ready.

4. `tests/cloud/memory/test_backend_selection.py:96` — The GitHub security scanner flagged this file for sys.modules manipulation (__import__("sys").modules.pop(mod, None)). Module cache clearing in tests is a fragile pattern that can cause import-related side effects across the test suite depending on execution order. Use importlib.reload() or isolate the backend-selection test in a subprocess instead.

**Non-blocking findings:**

5. `ee/cloud/memory/mongo_store.py:_normalize_session_key()` — The channel whitelist (websocket, telegram, discord, slack, whatsapp, cli) is duplicated from knowledge of the bus internals. If a new adapter is added to the bus, this list will silently miss it and session keys from the new channel will not normalize. Consider deriving this from a canonical list in the bus module, or logging a warning when a key contains ":" but the prefix is unrecognized.

6. `SPEC.md` committed to the repo root — spec documents belong in docs/ or the PR description, not the root. The PR description already contains the full spec content. The committed SPEC.md will confuse contributors browsing the repo.

7. `src/pocketpaw/memory/manager.py` — The new mongodb branch does a lazy `from ee.cloud.memory.mongo_store import MongoMemoryStore` inside create_memory_store(). This is the right pattern for OSS/ee import hygiene. Verified: no top-level ee imports from src/pocketpaw/. Good.

8. PR description includes "Generated with Claude Code" attribution at the bottom — that should be removed before merge per project conventions.

**What is genuinely good:**

- The discriminator design (context_type) is clean. Using a single messages collection rather than two separate ones is the right call for query simplicity.
- _get_group_message_or_404 now checks context_type != "group", which correctly prevents pocket messages from being addressable via group chat routes.
- The agent dispatch fix (ee/cloud/shared/agent_bridge.py) is a clean one-shot change that solves the both-auto-agents-reply bug without touching the broader agent loop.
- The SPEC is thorough — the acceptance criteria and boundaries sections in particular make it clear what this PR does and does not do.
- MongoMemoryStore implements the full protocol with no NotImplementedError stubs.

**Test coverage:** Adequate in intent but the real-Mongo fixture makes the new tests non-runnable in standard CI. The contract tests are a strong addition once the fixture is fixed.

**Architecture fit:** Targets the ee branch, not dev — correct. No OSS behavior change. Follows the MemoryStoreProtocol adapter pattern. The bootstrap.py approach to swapping ._store in place is unusual but the comment explaining why is accurate. No deprecated JSONL patterns introduced.

**Action:** Request changes. Fix the workspace scoping hole and the test fixture before merge. The spec, structure, and code quality are solid — this is worth landing once those two items are addressed.

DevRohit06:
Pushed `59374666` addressing the review feedback. Walking through each item:

**Blocking — addressed:**

1. **Workspace scoping** — `MemoryFactDoc` and `Message` both gain a `workspace_id` field (with appropriate compound indexes). `MongoMemoryStore.save()` resolves it from the linked `Session.workspace` for pocket rows, or from `entry.metadata["workspace_id"]` for facts. Two new adapter helpers — `get_session_in_workspace()` and `list_facts_in_workspace()` — enforce a workspace filter at read time and explicitly exclude untagged legacy rows so cross-tenant leakage is impossible by construction. Protocol-level methods stay tenant-agnostic (preserves the `MemoryStoreProtocol` contract for OSS callers; ee callers use the scoped helpers). Coverage in `tests/cloud/memory/test_workspace_isolation.py` (4 stamping cases + 3 read-isolation cases including the untagged-row exclusion).

2. **Test fixture** — switched to `mongomock-motor` (added to dev deps). The whole `tests/cloud/memory/` suite (75 tests) now runs in 1.34s without a real MongoDB. Wraps `database.list_collection_names` to drop the `authorizedCollections` kwarg Beanie >=1.26 passes (mongomock-motor doesn't accept it).

3. **`register_default_backend` ordering** — moved to AFTER `init_beanie` in `init_cloud_db`. The bootstrap is a true no-op until Beanie is ready, so the first MongoMemoryStore call can never race a not-yet-initialised collection.

4. **`sys.modules` manipulation** — replaced with source-inspection of `manager.py` to verify the `ee.*` import stays lazy. Same guarantee, no module-cache surgery.

**Non-blocking — addressed:**

5. `_normalize_session_key` now logs a `warning` when it sees a colon-shaped key with an unknown channel prefix. The `_KNOWN_BUS_CHANNELS` set has a comment pointing at `pocketpaw.bus.events.Channel` so future bus additions update both. Cleaner than silent fall-through.

6. `SPEC.md` deleted from repo root. The PR description carries the same content.

7. PR description rewritten without the Claude attribution.

**Genuinely good — kept:** discriminator design, narrowed `_get_group_message_or_404`, single-shot agent dispatch fix.

Re-review requested.

DevRohit06:
Review feedback addressed in commit [`59374666`](https://github.com/pocketpaw/pocketpaw/commit/59374666) (before #971 / #972 stacked on top).

**Blocking items — fixed:**

1. ✅ Workspace scoping — `MemoryFactDoc` + `Message` now carry `workspace_id`; every read uses the tenant-scoped helpers (`get_session_in_workspace`, `list_facts_in_workspace`). New `tests/cloud/memory/test_workspace_isolation.py` covers stamping + cross-tenant filtering.
2. ✅ Test fixture now uses `mongomock-motor` (added to dev deps). Wraps `database.list_collection_names` to drop the `authorizedCollections` kwarg Beanie ≥1.26 passes. Full `tests/cloud/memory/` + `tests/cloud/test_api_contracts.py` runs in CI — 88 tests passing locally.
3. ✅ Init ordering fixed: `init_cloud_db` now calls `register_default_backend()` **after** `init_beanie`, so `MongoMemoryStore` can't race an un-initialised collection.
4. ✅ `tests/cloud/memory/test_backend_selection.py` replaces `sys.modules` manipulation with source inspection of `pocketpaw/memory/manager.py` to verify the `ee.*` import stays lazy.

**Non-blocking — also done:**

5. ✅ `_normalize_session_key` logs a warning on unknown channel prefixes instead of silently passing through.
6. ✅ `SPEC.md` at repo root removed.

Running the suite at current HEAD ($(git rev-parse --short HEAD)):
\`\`\`
uv run pytest tests/cloud/memory/ tests/cloud/test_api_contracts.py
= 88 passed in 5.99s =
\`\`\`

PR #965, opened 2026-04-16
feat(uploads): pluggable StorageAdapter + OSS/EE upload routers with bulk support

DevRohit06:
## Summary

- New `src/pocketpaw/uploads/` module with a `StorageAdapter` protocol, `LocalStorageAdapter` (atomic writes, path-traversal safety), `UploadService` (validation, magic-byte sniff, bulk uploads), JSONL metadata store, and error hierarchy.
- OSS `POST/GET/DELETE /api/v1/uploads` router registered in the v1 aggregator.
- EE `ee/cloud/uploads/` — `FileUpload` Beanie doc, workspace-scoped `MongoFileStore`, `EEUploadService`, and EE router mounted in `mount_cloud`.
- Bulk upload support from day one (up to 50 files per request, partial-success response).

## Key decisions

- **Adapter is dumb bytes-in/bytes-out** — no metadata, auth, or mime logic. S3 swap later requires no changes to the service or routers.
- **Opaque `file_id`** — clients never see storage keys. Keeps local/S3 interchangeable.
- **Magic-byte sniff wins over `Content-Type`** — guards against mime-type spoofing. ZIP containers (docx/xlsx) keep the caller's declared mime.
- **EE wraps OSS via `_NullMeta`** — EE gets all validation/sniff logic for free and persists to Mongo itself.
- **Owner-only access in v1** — cross-user within a workspace returns 404 (don't leak existence). Group-member access is a follow-up.

## Test plan

- [x] 61/61 upload-suite tests green (`uv run pytest tests/uploads/ tests/cloud/uploads/`)
- [x] mypy clean on `src/pocketpaw/uploads` + `ee/cloud/uploads`
- [x] ruff format + check clean
- [ ] Manual: upload single / bulk / over-size / bad-mime via `curl`
- [ ] Manual: cross-workspace 404 (EE)
- [ ] Manual: backend restart preserves uploads + metadata (OSS JSONL reload)

## Out of scope (follow-up)

- Frontend wiring in paw-enterprise ChatInput → `POST /api/v1/uploads`
- Chat-member access check on EE downloads (currently owner-only)
- `S3StorageAdapter` (`ee/cloud/uploads/s3.py`) using the same contract
- Orphan cleanup job (`.tmp` files older than 1h; metadata without blob)
- Migrate existing ad-hoc upload sites (avatar, knowledge ingestion, soul imports) to `UploadService`
- AV/malware scanning hook on `put`

### DIFF
--- a/ee/cloud/__init__.py
+++ b/ee/cloud/__init__.py
@@ -39,8 +39,10 @@ def mount_cloud(app: FastAPI) -> None:
     app.include_router(sessions_router, prefix="/api/v1")
 
     from ee.cloud.kb.router import router as kb_router
+    from ee.cloud.uploads.router import router as uploads_router
 
     app.include_router(kb_router, prefix="/api/v1")
+    app.include_router(uploads_router, prefix="/api/v1")
 
     # User search endpoint — used by group settings, pocket sharing
     from ee.cloud.models.user import User as UserModel
--- a/ee/cloud/models/__init__.py
+++ b/ee/cloud/models/__init__.py
@@ -13,7 +13,18 @@ from ee.cloud.models.pocket import Pocket, Widget, WidgetPosition
 from ee.cloud.models.session import Session
 from ee.cloud.models.user import OAuthAccount, User, WorkspaceMembership
 from ee.cloud.models.workspace import Workspace, WorkspaceSettings
-from ee.cloud.uploads.models import FileUpload
+
+# Lazy import to avoid circular imports
+FileUpload: type = None  # type: ignore[assignment]
+
+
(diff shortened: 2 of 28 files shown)


DevRohit06:
Reviewed — changes look solid. Delete ordering hardened so a mid-op crash leaves an orphan blob (reclaimable) rather than a dangling record. Added X-Content-Type-Options: nosniff to both download responses. 60 unit tests green, CI clean. Known follow-ups (not blockers): tighter MIME-sniff for text/* types, orphan-blob cleanup on mid-batch Mongo failure in EE. Merging.

PR #967, opened 2026-04-16
feat(uploads): URL→disk-path resolver + chat wiring so agents see attached files

DevRohit06:
## Summary

Closes the gap that blocks agents from seeing files the user attaches. The frontend sends `MediaAttachment.url = "/api/v1/uploads/{id}"`, the chat endpoint passes those strings verbatim into `InboundMessage.media`, and the agent loop injects them into the prompt as `[Media files on disk: /api/v1/uploads/u1]` — which the agent's Read tool cannot open. This PR wires a resolver so those URLs become real on-disk paths before the prompt is built.

Replaces closed PR #966 (original base `feat/upload-adapter` was merged + deleted); now targets `ee` directly.

## Changes

- `StorageAdapter.local_path(key) -> Path | None` — new protocol method. `LocalStorageAdapter` returns the resolved path when the blob exists; future remote adapters return `None` so callers can fall back to `open()`.
- `pocketpaw/uploads/resolver.py` — `parse_upload_url`, `UploadResolver`, `resolve_media_paths`. Unresolvable upload URLs are dropped with a warning log; non-upload strings pass through.
- `ee/cloud/uploads/resolver.py` — workspace-scoped counterpart over MongoFileStore. Cross-tenant lookups return None (no existence leak across workspaces).
- `api/v1/chat.py:_send_message` — resolves `ChatRequest.media` before publishing the inbound message.
- `adapter.local_path()` wrapped in try/except in both resolvers so OSError from permission/remount issues can't crash `_send_message`.

## Agent pipeline after this PR

1. Frontend uploads → `{ url: "/api/v1/uploads/u1", ... }` in `MediaAttachment` (paw-enterprise PR #77).
2. `POST /chat/stream` — `ChatRequest.media = ["/api/v1/uploads/u1", ...]`.
3. `_send_message` → `resolve_media_paths` → `InboundMessage.media = ["/abs/path/to/blob", ...]`.
4. `AgentLoop` appends `[Media files on disk: <path>]` and the agent's Read tool opens it.

## Test plan
- [x] `tests/uploads/test_resolver.py` — 13 unit tests covering parse, resolve, soft-delete, missing-on-disk
- [x] `tests/cloud/uploads/test_resolver.py` — 6 EE tests including cross-tenant isolation
- [x] `tests/test_api_chat.py::TestSendMessageResolvesMedia` — integration test for `_send_message` resolution
[...]

PR #968, opened 2026-04-16
fix(uploads): resolve media URLs from EE Mongo when OSS JSONL misses

DevRohit06:
## Problem

Uploads in self-hosted EE flow through the workspace-scoped `ee.cloud.uploads.router` (writes to MongoDB), but chat still goes through the OSS `POST /api/v1/chat/stream` endpoint (no auth, no workspace context). The resolver merged in #967 only reads the OSS `JSONLFileStore`, so every EE-backed upload gets silently dropped before reaching the agent loop.

**Concrete repro:** user uploads a screenshot via `ChatInput` → Mongo has the record (`file_uploads` collection) → user sends chat with the returned URL → resolver's JSONL lookup misses → `logger.warning("dropping unresolvable upload entry: ...")` → agent gets `InboundMessage.media = []` → agent responds "you didn't attach an image."

## Fix

- `MongoFileStore.get_unscoped(file_id)` — find a live record by id without workspace filter. Documented as for single-user self-hosted use; multi-tenant cloud chat should continue to use `get_scoped` with an authenticated workspace.
- `pocketpaw.uploads.resolver.resolve_media_paths_any(media)` — async chain: try OSS JSONL first; on miss, fall back to EE Mongo; on double miss, drop with a warning log. EE is imported lazily so pure-OSS installs stay unaffected.
- `api/v1/chat.py:_send_message` awaits the async chain.

## Test plan
- [x] `tests/cloud/uploads/test_resolver_fallback.py` — 4 new tests (fallback hit, double-miss drop, soft-delete respected, workspace-less get)
- [x] `pytest tests/uploads/ tests/cloud/uploads/ tests/test_api_chat.py` → 74 passed
- [x] Ruff + mypy clean on touched files (pre-existing mypy noise in `chat.py:170-171` is unrelated)

## Known limits

`get_unscoped` ignores tenant isolation — acceptable here because the OSS chat endpoint has no auth context to check a workspace against, and the file_id is a 128-bit random UUID (not guessable). A proper multi-tenant fix is to route chat through a workspace-aware EE endpoint; that's a larger refactor outside this hotfix.

### DIFF
--- a/ee/cloud/uploads/mongo_store.py
+++ b/ee/cloud/uploads/mongo_store.py
@@ -30,6 +30,24 @@ class MongoFileStore:
             FileUpload.workspace == workspace,
             FileUpload.deleted_at == None,  # noqa: E711 beanie needs literal None
         )
+        return self._to_record(doc)
+
+    async def get_unscoped(self, file_id: str) -> FileRecord | None:
+        """Find a live record by file_id without workspace filter.
+
+        Intended for call sites that lack tenant context (e.g. the OSS chat
+        bridge in single-user self-hosted deployments). Multi-tenant cloud
+        chat flows should use ``get_scoped`` with an authenticated workspace
+        and never call this.
+        """
+        doc = await FileUpload.find_one(
+            FileUpload.file_id == file_id,
+            FileUpload.deleted_at == None,  # noqa: E711
+        )
+        return self._to_record(doc)
+
+    @staticmethod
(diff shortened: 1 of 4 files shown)


DevRohit06:
Reviewed — direct fix for the EE-upload / OSS-chat gap hit in local testing. Fallback is lazy-imported so pure OSS is unchanged; tenant-safety trade-off documented in the PR body. Merging.

PR #983, opened 2026-04-19
fix(security): require workspace admin on fleet install (P0)

prakashUXtech:
**P0 SECURITY. Please gate merge on review, don't auto-merge on green.**

## Summary
- `POST /fleet/install` had no auth guard at all. Any authenticated user could install a fleet (and spawn the agents + pockets) into any workspace, including workspaces they weren't in. Flagged in cluster-D-reality.md line 108.
- The route now requires `current_active_user`, so unauth callers get 401. The caller also has to be owner or admin of the target `workspace_id`, otherwise 403.
- `workspace_id` is a new required field on `InstallFleetRequest`. `template_name`, `journal`, and `actor` stay the same.
- Registered `fleet.install` in the ACTIONS matrix at `WorkspaceRole.ADMIN` and swapped the handler over to `check_workspace_action` so denials hit `log_denial` automatically. The RBAC matrix test picks the new row up without any extra wiring.

## Test plan
- [x] `tests/ee/test_fleet_router.py::TestInstallFleetAuth`: unauthenticated → 401, non-member → 403, member below admin → 403, admin → 200, owner → 200. Each deny case asserts the installer mock was never called, so a 403 can't silently spawn anything.
- [x] `TestInstallFleet` cases updated to send `workspace_id` + admin header, so the happy paths exercise the guard too.
- [x] `tests/cloud/test_rbac_matrix.py` parametrises over ACTIONS, so `fleet.install` is covered for every `WorkspaceRole` automatically.
- [x] `cd pocketPaw && uv run pytest tests/ee/ -k fleet -q`: 28 passed.
- [x] `cd pocketPaw && uv run pytest tests/ee/ tests/cloud/test_rbac_matrix.py -q`: 227 passed.
- [x] `/security-review` run on the diff.

## Security review
- Layer 5 (audit log): first draft called `resolve_workspace_role` directly, which skipped `log_denial`. Rewired through `check_workspace_action` so every 403 is audited.
- Layer 6 (file sandbox): `ee/fleet/installer.py::load_fleet` has a pre-existing path-traversal surface. An unknown template name falls through to `Path(name)` and reads whatever file is on disk. Out of scope for this P0, the guard limits the blast radius to workspace admins who are already trusted, but the loader should be clamped to the bundled directory. Filing a follow-up.
- Nothing else relevant here. No shell, no new deps, no new user input channels.

## Notes
- Not a breaking change for existing fields. `InstallFleetPanel.svelte` will need a one-line `workspace_id: currentWorkspace.id` when it's wired in. Not blocking right now: the panel isn't currently mounted on any route, so nothing real hits the 422 the server would now return.

### DIFF
--- a/ee/fleet/installer.py
+++ b/ee/fleet/installer.py
@@ -13,6 +13,18 @@
 # summary at the end. The journal parameter is opt-in so existing callers
 # (tests, CLI without an org) keep working unchanged. Emission errors are
 # logged and swallowed — the journal is observability, not control flow.
+# Updated: 2026-04-19 (fix/fleet-install-auth-guard) — P0 path-traversal
+# clamp. ``load_fleet`` previously fell through to ``Path(path_or_name)``
+# for any string that did not match a bundled template, which let a
+# workspace admin pass ``"../../etc/passwd"`` and have the server read +
+# attempt to parse arbitrary files. String inputs (the only code path the
+# REST router can reach) are now resolved against ``_BUNDLED_DIR`` and
+# rejected with a generic ``FileNotFoundError`` if they escape that
+# directory — the error never echoes the attempted filesystem path, so a
+# 4xx response leaks nothing about disk state. ``Path`` instances keep
+# the unclamped behaviour because they only come from trusted
+# programmatic callers (tests, scripts); the router never reaches that
+# branch.
 
 from __future__ import annotations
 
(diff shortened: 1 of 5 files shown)


prakashUXtech:
**P0 SECURITY — Request changes on one item, then approve.**

The auth gap is closed correctly and the tests cover every case that matters for a P0 fix. Auth is checked before the filesystem is touched, every 403 is audited via `log_denial`, and the happy paths for both admin and owner are pinned. This is solid work.

One item needs to be resolved before merge. Everything else is noted as nits or confirmed clean.

---

**Must fix before merge**

`load_fleet` in `ee/fleet/installer.py` has a path traversal surface that this PR correctly flags in its own security review notes, but does not guard against. The function logic is:

1. If the argument matches a bundled fleet name (`_BUNDLED_DIR / f"{path_or_name}.yaml"` exists), load from there — safe.
2. Otherwise, fall through to `Path(path_or_name)` and read whatever is on disk.

The auth guard added in this PR limits who can reach step 2 to workspace admins, which does reduce the blast radius significantly. But a workspace admin can still pass `template_name: "../../secrets"` (or any other relative path) and get the server to read an arbitrary file and attempt to parse it as YAML/JSON. The error message on a parse failure leaks path information. The fix is one line in `_load_from_path` — resolve the path against `_BUNDLED_DIR` and check that the resolved path starts with `_BUNDLED_DIR`. The PR body acknowledges this as a known gap and says a follow-up will be filed. For a P0 landing in `ee`, I'd rather see the clamp land here in the same commit, or the follow-up issue created and linked before merge. Filing the issue now costs less than a second round of review.

---

**Nice to have**

- `_require_fleet_install` re-raises `GuardForbidden` as `HTTPException(403, detail=exc.code)`. The `exc.code` values (`workspace.not_member`, `workspace.insufficient_role`) are machine-readable strings, which is consistent with every other ee cloud router. Good — no change needed, just confirming the pattern is right.
- The `TestInstallFleetAuth` class names are descriptive and the deny cases each assert `patch_install_fleet.call_count == 0`. That assertion is the most important part of a security regression test and it's there. Good.
- `workspace_id` is a plain `str` field on `InstallFleetRequest` with no length or format validation. A very long string or a string with special characters won't cause injection here because it's only passed to `resolve_workspace_role` for a list scan, but adding `min_length=1` and `max_length=128` to the `Field()` would make the contract explicit and reject obviously malformed inputs at the Pydantic layer.

---

**Clean**

- Auth is enforced before `load_fleet` is called — the 403 leaks nothing about template state.
- `fleet.install` is registered at `WorkspaceRole.ADMIN` in `ACTIONS` with `deny_code="workspace.insufficient_role"`. The RBAC matrix test picks this up automatically (confirmed by the PR body: 227 passed).
- `current_active_user` is the first `Depends` in the handler signature — FastAPI resolves it before `get_journal`, so an unauthenticated caller never triggers a journal open.
- The fake `current_active_user` override in tests uses `dependency_overrides`, which is the canonical FastAPI pattern. Header-driven membership (`X-Test-Workspaces`) is readable and covers all relevant role levels.
- All existing `TestInstallFleet` cases now send `workspace_id` and the admin header — no pre-existing happy path is left unguarded.
- File header comment is updated on both changed Python files following the project convention.

prakashUXtech:
Clamp pushed in 2c4b93c.

`load_fleet` now resolves any string input against `_BUNDLED_DIR.resolve()` and rejects anything that falls outside that directory with a generic `FileNotFoundError`. The error message echoes only the user's own input string, never the resolved filesystem path, so a 404 body leaks nothing about disk state. `Path` instances keep the old behaviour because those only come from trusted programmatic callers (tests, scripts); the REST router always passes strings, so the untrusted path is clamped end to end.

Four new tests in `TestLoaderBundledNameClamp` pin the contract the reviewer called out:

- `load_fleet("sales-fleet")` still loads the bundled template.
- `load_fleet("../../etc/passwd")` raises `FileNotFoundError` with no filesystem path leaked.
- `load_fleet("/etc/passwd")` raises `FileNotFoundError`.
- `load_fleet("nonexistent")` raises `FileNotFoundError`.

Plus a belt-and-braces `load_fleet("..")` case so a single `..` segment cannot sneak through.

Tests: `tests/ee/` 113 passed, fleet-scoped `-k fleet` 28 passed, original 15 loader/orchestration tests still green. Closes the gap acknowledged in the PR body — filing #984 as resolved.

PR #969, opened 2026-04-16
feat(agent): inject filename/mime/size into prompt so agents know what's attached

DevRohit06:
## Summary

Before this PR the agent prompt injected \`[Media files on disk: /abs/chat/202604/abc.png]\` — a storage-key path the model had to Read just to learn what kind of file it got. Now:

\`\`\`
Attached files:
- screenshot.png (image/png, 404.5 KB) at /abs/chat/202604/abc.png
\`\`\`

The agent sees filename + mime + size up front and can decide whether to Read (text/code), treat as an image, or ignore — without opening the bytes first.

## Changes

- \`pocketpaw/uploads/resolver.py\` gets a \`ResolvedMedia(path, record)\` dataclass and \`resolve_media_with_records\` that yields path + FileRecord pairs. Passthrough entries (non-upload strings) get \`record=None\`.
- \`api/v1/chat.py:_send_message\` uses the richer form to populate \`InboundMessage.metadata[\"media_info\"]\` — list of \`{path, filename, mime, size}\` dicts, one per resolved upload.
- \`agents/loop.py\` renders the richer block when \`media_info\` is present and falls back to the old plain-path line when it isn't — so adapters that don't know about uploads (Telegram/Discord/Slack/WhatsApp) keep their existing behavior.
- \`_format_bytes\` helper for human-readable sizes in the prompt.

## Test plan
- [x] \`tests/uploads/test_resolver.py\` — 3 new tests: OSS hit yields record, passthrough has \`None\` record, unresolvable drops.
- [x] \`tests/test_api_chat.py\` — 1 new test confirming \`media_info\` populated for resolved URLs and absent for passthroughs.
- [x] \`tests/test_agent_loop_media.py\` — 4 tests for the size formatter (B / KB / MB / GB).
- [x] \`pytest tests/uploads/ tests/cloud/uploads/ tests/test_api_chat.py tests/test_agent_loop_media.py\` → 102 passed.
- [x] Ruff + mypy clean on touched files.

## Not in scope
- Native vision blocks for Claude / OpenAI / Gemini (per-backend work).
- Persisting \`attachments\` on the \`Message\` doc so reloaded history shows them — follow-up on the feat/ee-mongo-memory-backend track.

PR #986, opened 2026-04-19
fix(chat): forward channel attachments into agent prompt

prakashUXtech:
## Summary

Channel agents were missing the filename/mime/size block that DM agents already get from PR #969. Drop a PDF in a group chat and the agent answered as if no file existed.

The `message.sent` event payload didn't carry attachments, and `_run_agent_response` had no attachments parameter. So they never reached the prompt passed to `pool.run`.

## What changed

- `ee/cloud/chat/message_service.py`: include `body.attachments` in the `message.sent` emit. Additive change. Existing subscribers ignore unknown keys.
- `ee/cloud/shared/agent_bridge.py`: `_dispatch_agent_responses` now reads `data["attachments"]` and passes it to `_run_agent_response`, which appends an `Attached files:` block to the user prompt before calling `pool.run`. Shape matches the DM path in `src/pocketpaw/agents/loop.py` (filename, mime, readable size, URL).
- `pool.run`'s signature is unchanged.

## Test plan

- [x] New test `tests/cloud/shared/test_agent_bridge_attachments.py` fails on pre-fix code with `TypeError: _run_agent_response() got an unexpected keyword argument 'attachments'` and passes on this branch.
- [x] `uv run pytest tests/cloud/shared/ tests/cloud/chat/`: passing tests stay green. Two pre-existing failures on other files are unrelated (hardcoded Windows path; audit export matcher).
- [ ] Manual smoke: drop a file into a group with an auto-mode agent and check the reply references the filename or type.

### DIFF
--- a/ee/cloud/chat/message_service.py
+++ b/ee/cloud/chat/message_service.py
@@ -1,6 +1,9 @@
 # Refactored: Split from service.py — contains MessageService class and message-related
 # helper functions. Added create_agent_message() static method for use by agent_bridge
 # instead of creating Message documents directly.
+# 2026-04-19: ``message.sent`` event now carries ``attachments`` so agent_bridge
+# can surface filename/mime/size into the channel agent prompt (fixes silent
+# attachment drop on the channel path — DM path already had this).
 
 """Chat domain — message business logic (CRUD, reactions, threads, pins, search)."""
 
@@ -195,6 +198,12 @@ class MessageService:
                 "sender_type": "user",
                 "content": body.content,
                 "mentions": body.mentions,
+                # Attachments ride on the event so ``agent_bridge`` can inject
+                # filename / mime / size into the agent prompt. Mirrors the DM
+                # path's file-awareness contract; raw dicts match
+                # ``body.attachments``'s shape and downstream handlers no-op on
+                # unknown keys.
+                "attachments": body.attachments,
                 "workspace_id": group.workspace,
                 **reply_meta,
             },
--- a/ee/cloud/shared/agent_bridge.py
+++ b/ee/cloud/shared/agent_bridge.py
@@ -1,7 +1,12 @@
 """Bridge between cloud chat events and the PocketPaw agent pool.
 
-Changes: Replaced inline pocket creation with PocketService.create_from_ripple_spec()
-to reduce coupling. Pocket creation logic now lives in the pockets domain.
+Changes:
+- 2026-04-19: Forward ``Message.attachments`` through ``on_message_for_agents``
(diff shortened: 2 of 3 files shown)


prakashUXtech:
**Approve with nits.**

The fix is correct. Attachments were silently dropped on the channel path while the DM path already handled them — this closes that gap cleanly.

---

### What's working well

- The `or []` double-normalization in `_dispatch_agent_responses` correctly handles `None` coming off the event payload, since `message_service.py` passes `body.attachments` raw without normalizing first.
- `_format_bytes` handles the boundary cases correctly (`0`, `1023`, `1024`, and the TB ceiling guard).
- `_augment_message_with_attachments` degrades gracefully: non-dict entries are skipped, all attachment fields are optional, and a name-only attachment still shows up rather than being dropped silently.
- The test is well-structured. It patches just enough (pool, Mongo query chain, emit) to stay at unit scope while capturing what matters — the prompt handed to `pool.run`. The three assertions cover filename, mime, and the attachment header.

---

### Must fix before merge

Nothing blocking.

---

### Nice to have

**1. Filename and URL are passed unsanitized into the agent prompt.**

`_augment_message_with_attachments` does:

```python
lines.append(f"- {name} ({mime}{meta_suffix}){location}")
```

`name` and `url` come from `body.attachments` which is user-controlled. A crafted filename like `"photo.jpg\n\nIgnore previous instructions..."` would inject a newline into the agent's user-content block. This same gap exists on the DM path (so this PR isn't making things worse), but it would be worth a shared sanitization pass on both paths. At minimum, stripping or replacing `\n` and `\r` in `name`, `url`, and `mime` before interpolation would close the most obvious vector.

**2. The test's lowercase branch in the assertion is dead code.**

```python
assert "Attached" in augmented or "attached" in augmented
```

The helper always produces `"Attached files:"` with a capital A, so the `or "attached" in augmented` branch will never fire. Replacing with `assert "Attached files:" in augmented` matches the actual contract and is easier to read.

**3. No test for `attachments=None` passed explicitly.**

The `or []` guard in `_dispatch_agent_responses` handles `None`, and `_augment_message_with_attachments` returns early on falsy input. A one-liner test confirming `_augment_message_with_attachments("hello", None) == "hello"` would lock that edge case in without much effort.

---

### Clean

- File-top comments updated correctly.
- Async/sync split is appropriate — `_format_bytes` and `_augment_message_with_attachments` do no I/O so sync is correct.
- `pool.run` signature is untouched, as described.
- Test patches are tight; no real DB or API calls.

---

The bot comment about missing test output is a false positive — the test file is new and the PR body includes the relevant `pytest` result. The `needs-work` label can come off once the bot or a maintainer re-scans.

PR #1013, opened 2026-04-21
fix: add get_tool_policy/set_tool_policy to AgentBackend protocol

### LINKED ISSUE #1012: fix: AgentBackend protocol missing get_tool_policy/set_tool_policy
### Problem Statement

AgentBackend protocol (src/pocketpaw/agents/backend.py) has no interface for getting or setting tool policy at runtime. Each backend stores its
  policy privately with no common access point.                                                                                                   
                  
  This causes a silent failure in loop.py:768 where per-pocket tool restrictions (EE tenant sessions) are meant to be scoped per request. The     
  guard condition:
                                                                                                                                                  
  if pocket_deny_tools and hasattr(router, "_registry") and router._registry:

  AgentRouter has _backend, not _registry — so hasattr() is always False. The entire block is dead code. Per-pocket tool policy is never applied, 
  never restored, and no error is raised.

### Proposed Solution

- Add get_tool_policy() / set_tool_policy() to the AgentBackend protocol
  - Implement on all 7 backends with appropriate adapters (backends with cached tool lists must invalidate cache on set_tool_policy())
  - Update tool_bridge.py build functions to accept an optional policy parameter so backends pass self._policy instead of reconstructing from settings                                                                                                                                        
  - Fix the dead-code block in loop.py to use router._backend.get_tool_policy() / set_tool_policy()

- src/pocketpaw/agents/backend.py (add get_tool_policy() / set_tool_policy() to protocol)
  - src/pocketpaw/agents/claude_sdk.py (implement protocol methods)                                                                             
  - src/pocketpaw/agents/openai_agents.py (implement protocol methods, invalidate _custom_tools cache on set)
  - src/pocketpaw/agents/google_adk.py (implement protocol methods, invalidate cache on set, fix _build_mcp_toolsets to use self._policy )        
  - src/pocketpaw/agents/deep_agents.py ( implement protocol methods, invalidate both caches on set, fix _build_mcp_tools to use self._policy )    
  - src/pocketpaw/agents/codex_cli.py (implement protocol methods)                                                     
  - src/pocketpaw/agents/copilot_sdk.py (implement protocol methods )                                                   
  - src/pocketpaw/agents/opencode.py — implement protocol methods)                                                  
  - src/pocketpaw/agents/tool_bridge.py (add optional policy param to 3 build functions )                                                         
  - src/pocketpaw/agents/loop.py (replace broken hasattr(router, "_registry") with router._backend + protocol method calls)
  - tests/test_backend_protocol.py (add TestToolPolicyProtocol covering all 7 backends )                                                          
  - tests/test_google_adk_backend.py  (fix _make_backend() helper and 2 tests to use set_tool_policy()) 

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Before submitting

TejasriPacharu:
> **Before opening this PR:**

## What does this PR do?

Extends the `AgentBackend` protocol with `get_tool_policy()` and `set_tool_policy()` methods and implements them across all 7 backends. Fixes a dead-code bug in `loop.py` where per-pocket tool restrictions were silently never applied because the code guarded on `router._registry` which never existed — now uses `router._backend` with the new protocol methods.

## Related Issue
#1012

## Fixes
- Dead-code bug in `loop.py` where per-pocket `deny_tools` were silently never applied due to `hasattr(router, "_registry")` always returning `False`
- Race condition where concurrent pocket requests could clobber each other's tool policy mid-request
- Fallback backends running with unscoped policy when a pocket had `deny_tools` configured
- Third-party backends without the protocol methods causing `AttributeError` at runtime
- Store-only backends (`codex_cli`, `copilot_sdk`, `opencode`) silently accepting per-pocket deny rules with no indication they were not enforced

## Changes Made

- `src/pocketpaw/agents/backend.py`: Added `get_tool_policy()` and `set_tool_policy()` to `AgentBackend` protocol
- `src/pocketpaw/agents/claude_sdk.py`: Implemented protocol methods (trivial — `self._policy` already existed)
- `src/pocketpaw/agents/openai_agents.py`: Added `self._policy`, Pattern B adapter (invalidates `_custom_tools` cache on set)
- `src/pocketpaw/agents/google_adk.py`: Added `self._policy`, Pattern B adapter, `_build_mcp_toolsets` now uses `self._policy`
- `src/pocketpaw/agents/deep_agents.py`: Added `self._policy`, Pattern B adapter (invalidates both `_custom_tools` and `_mcp_tools` caches on set), `_build_mcp_tools` now uses `self._policy`
- `src/pocketpaw/agents/codex_cli.py`: Added `self._policy`, store-only adapter (external CLI)
- `src/pocketpaw/agents/copilot_sdk.py`: Added `self._policy`, store-only adapter (external CLI)
- `src/pocketpaw/agents/opencode.py`: Added `self._policy`, store-only adapter (external server)
- `src/pocketpaw/agents/tool_bridge.py`: Added optional `policy` param to `build_openai_function_tools`, `build_adk_function_tools`, `build_deep_agents_tools`
- `src/pocketpaw/agents/loop.py`: Replaced broken `hasattr(router, "_registry")` guard with `scoped_tool_policy()` context manager on `AgentRouter`
- `src/pocketpaw/agents/router.py`: Added `_policy_lock` + `scoped_tool_policy()` async context manager; `hasattr` guard for third-party backends
- `tests/test_backend_protocol.py`: Added `TestToolPolicyProtocol` covering all 7 backends with existence, round-trip, and cache-invalidation assertions
- `tests/test_google_adk_backend.py`: Updated `_make_backend()` to initialise `_policy`; updated 2 tests to use `set_tool_policy()` instead of mutating settings directly

## Review Feedback Addressed

- `loop.py`: Replaced manual save/restore pattern with `scoped_tool_policy()` context manager on `AgentRouter` — no more raw `get_tool_policy()`/`set_tool_policy()` calls scattered in loop logic
- `router.py`: Added `_policy_lock` + `scoped_tool_policy()` async context manager — serializes concurrent pocket requests eliminating the policy clobber race condition, and scopes primary + all fallback backends in one place
- `router.py`: Added `hasattr` guard on the active backends list so third-party backends missing the protocol methods do not `AttributeError` at runtime
- `codex_cli.py`, `copilot_sdk.py`, `opencode.py`: Added `logger.debug` + comment on store-only setters to surface the external-process limitation
- `tests/test_backend_protocol.py`: Strengthened `TestToolPolicyProtocol` with round-trip assertions and cache-invalidation assertions for `openai_agents`, `google_adk`, and `deep_agents`

## How to Test

1. Run `uv run pytest tests/test_backend_protocol.py -v` — all 35 cases should pass
2. Run `uv run pytest tests/test_google_adk_backend.py -v` — all cases should pass
3. Run `uv run pytest --ignore=tests/e2e` — full suite should pass
4. Run `uv run ruff check .` — should be clean

## Evidence of Testing

image
image
image
image

## Checklist


DevRohit06 (commented):
## Review

### Overview
Fixes a real bug in `loop.py` where per-pocket `deny_tools` were silently never applied because the guard checked `router._registry` (which doesn't exist on `AgentRouter` — only `router._backend` does). Formalizes tool-policy access as two protocol methods and implements them across all 7 backends.

### What's good
- **Root cause fix is correct.** `AgentRouter` exposes `_backend`, not `_registry` (`router.py:27`).
- **Consistent pattern across backends** — every backend now initializes `self._policy` from settings and implements both methods with the same shape.
- **Correct cache invalidation** where it matters: `openai_agents`, `google_adk`, and `deep_agents` clear cached tool lists; `deep_agents` correctly clears both `_custom_tools` and `_mcp_tools`.
- **`tool_bridge.py` stays backward-compatible** — new `policy` param is optional.
- **MCP policy construction correctly moved** in `google_adk._build_mcp_toolsets` and `deep_agents._build_mcp_tools` to use `self._policy` so live updates take effect for MCP server filtering.

### Issues / risks

1. **External-CLI backends silently no-op on `set_tool_policy`.** `codex_cli`, `copilot_sdk`, and `opencode` store the new policy but never consult `self._policy` anywhere — tools run in the external process. Per-pocket `deny_tools` against these backends will be accepted silently and ignored. Please add a docstring note or `logger.debug` on those setters so this is discoverable.

2. **Fallback backends are not scoped.** `loop.py` only calls `set_tool_policy` on `router._backend`. If `AgentRouter.run()` falls through to a `_fallback_backends` entry (`router.py:125`), that backend runs with its own un-scoped policy. Pre-existing gap, but the fix makes this code path actually live for the first time — worth at least a FIXME or iterating over `_fallback_instances` too.

3. **Race condition on concurrent messages.** The save/scope/restore pattern mutates shared backend state. Two concurrent `_process_message_inner` calls for different pockets will clobber each other's `_saved_policy`. Pre-existing design, but now observable. At minimum add a FIXME.

4. **Protocol methods have no defaults.** Any third-party `AgentBackend` not updated for this protocol will `AttributeError` inside `loop.py`. Consider `hasattr` guards or default implementations on a base mixin.

5. **Tests are shallow.** `TestToolPolicyProtocol` only asserts the methods *exist*. A round-trip test — `set_tool_policy(p)` then assert `get_tool_policy() is p`, plus asserting `_custom_tools is None` after set for the cache-invalidating backends — would catch real regressions. A no-op `def set_tool_policy(self, p): pass` would pass the current test.

### Style
Follows ruff + import conventions. `ToolPolicy` moved out of `TYPE_CHECKING` guard where now needed at runtime — correct.

### Recommendation
Approve with minor changes requested:
- Docstring/log note on the three store-only setters (codex/copilot/opencode).
- Strengthen `TestToolPolicyProtocol` with round-trip + cache-invalidation assertions.
- Optional: FIXME in `loop.py` about fallback backends and the concurrent-pocket race.

Core fix is sound and the bug it addresses is real.

DevRohit06 (approved):
All five items from my earlier review are addressed:

1. Store-only setters (`codex_cli`, `copilot_sdk`, `opencode`) — `logger.debug` + explanatory comment added. ✅
2. Fallback backends — `router.scoped_tool_policy` iterates `[_backend] + _fallback_instances.values()`, scoping primary + fallbacks in one place. ✅ (went beyond the FIXME I'd have accepted)
3. Concurrent-request race — `_policy_lock` + async context manager serializes the save/scope/restore. ✅ (went beyond the FIXME I'd have accepted)
4. Third-party backends without protocol methods — `hasattr` guard in `scoped_tool_policy` prevents `AttributeError`. ✅
5. Shallow tests — `TestToolPolicyProtocol` now has `test_round_trip` and `test_cache_invalidation_on_set` across all 7 backends. ✅

Core fix (`router._registry` → `router._backend` via the new protocol) is correct, cache invalidation matches each backend's caching pattern, and `google_adk` / `deep_agents` now read `self._policy` inside `_build_mcp_*` so there's a single source of truth.

Nit (non-blocking): the new tests use `cls.__new__(cls)` to skip `__init__`, which satisfies the round-trip/invalidation checks but won't catch a backend whose `__init__` forgets to set `self._policy` — exactly the shape of the ADK fixture fix in this PR. A follow-up adding one normal-construction assertion per backend would harden against that regression class.

LGTM.

PR #1032, opened 2026-04-30
feat(extraction): pluggable extraction adapters with local + gemini-flash

prakashUXtech:
## What ships

Adds `ee/cloud/extraction/`, a pluggable package that lets each MIME pick its own extractor. Wraps the existing pypdf, python-docx, and pytesseract logic into `LocalExtractor` and adds a `GeminiFlashExtractor` sibling for image captioning. `agents/knowledge.py:_extract_file` now dispatches through the chain.

This is Stage 1.A of the "Files as Knowledge" plan (paw-workspace PR #9). The `FileReady` listener and indexing wiring are Stage 1.B's PR.

## Why pluggable, not a swap

Captain directive: keep the traditional extractors. Reasons:

- Self-hosted and offline deployments need a no-network path.
- Cost-sensitive uploads (large text-only PDFs) don't need cloud captioning.
- Privacy-classified files may not be allowed to leave the network.

`POCKETPAW_EXTRACTION_CHAIN` controls the order. `LocalExtractor` is always built as the offline fallback regardless of config.

## New surface

```python
class ExtractionAdapter(Protocol):
    name: str
    supports_mimes: set[str]
    requires_network: bool
    async def extract(self, path: Path, mime: str) -> ExtractionResult: ...
```

Concrete adapters:

- `LocalExtractor` — `name="local"`, wildcard MIME, `requires_network=False`. Behaviour-preserving port of the previous `_extract_file`.
- `GeminiFlashExtractor` — `name="gemini-flash"`, image/png + image/jpeg + image/jpg + application/pdf, `requires_network=True`. Uses google-genai (already in pyproject).

`build_chain(settings)` reads:

- `POCKETPAW_EXTRACTION_CHAIN` (list, default `["local"]`)
- `POCKETPAW_EXTRACTION_PER_MIME` (dict, default `{}`)
- `POCKETPAW_EXTRACTION_OFFLINE_FALLBACK` (str, default `"local"`)
- `POCKETPAW_GEMINI_API_KEY` (str, default `None`)

When `POCKETPAW_GEMINI_API_KEY` is unset, `gemini-flash` is silently dropped during chain construction so misconfigured prod doesn't crash startup.

## Selection rules

1. Per-MIME override wins.
2. Otherwise the first chain adapter whose `supports_mimes` matches the MIME (or has `*`).
3. Network-required adapters get skipped if the host is offline; the fallback handles it.
4. On adapter exception the chain logs and falls through to the offline fallback. Fallback exceptions propagate.
5. If no adapter claims the MIME, the offline fallback handles it.

## PDF strategy in GeminiFlashExtractor

Stage 1.A doesn't ship a PDF-to-image renderer (no new dep). The PDF path uses pypdf for text per page; pages with <200 chars get marked `[page N: image-heavy, no caption]` and excluded from captioning. If we want to caption image-heavy pages later, that's a follow-up and the dep choice (pdf2image, pypdfium2, etc.) needs to be picked then.

## Test plan

- [x] LocalExtractor parity tests: text fallback, invalid UTF-8 with errors='replace', blank PDF, missing pypdf raises, DOCX with mocked python-docx, missing python-docx raises, image with mocked pytesseract+PIL, missing pytesseract raises, wildcard MIME / metadata.
- [x] GeminiFlashExtractor: model name + MIME on Part + image bytes go through the SDK; configured model; empty response handled; PDF sparse-page marker; PDF makes no Gemini call when every page is sparse; unsupported MIME raises.
- [x] ExtractionChain: per-MIME override wins; first match wins; wildcard supports any MIME; unmatched MIME falls through to fallback; offline skips network adapter; failing adapter falls through; fallback failure propagates; unknown override name falls through.
- [x] `build_chain`: local-only; skips gemini without key; includes gemini with key; unknown adapter name raises; per-MIME map passed through.
- [x] Full non-e2e suite green: 4327 passed, 1 skipped.
- [x] `ruff check` clean on touched files.
- [ ] `mypy ee/cloud/extraction/`: 2 pre-existing import-not-found errors for `docx` and `pytesseract` (no public type stubs). Same errors existed in the previous `_extract_file`.

## Files

- New: `ee/cloud/extraction/{__init__,adapter,local,gemini_flash,chain}.py`
- New: `tests/cloud/extraction/{test_local,test_gemini_flash,test_chain}.py`
- Modified: `ee/cloud/agents/knowledge.py` (`_extract_file` now calls the chain)
- Modified: `src/pocketpaw/config.py` (4 new settings)

## Out of scope (Stage 1.B)

- FileReady listener subscription
- `KnowledgeService.ingest_text_to_scope`
- Multi-scope context_builder
- Wiring extraction into `uploads/service.py`


prakashUXtech:
Closing in favor of #TBD — same content rebased onto `ee` per updated branch policy. The original branch (`feat/extraction-adapters`) was based off `dev` which has diverged from `ee`; cherry-picked the single feature commit cleanly onto `origin/ee` in `feat/extraction-adapters-rebase`. All 28 extraction tests still pass; full non-e2e suite green.

PR #1039, opened 2026-05-01
feat(cloud): pockets-builder — backend-agnostic native pocket creation

prakashUXtech:
## Summary

Pocket creation only worked on the `claude_agent_sdk` backend. The whole flow lived behind in-process MCP servers that no other backend can register, so Codex / Ollama / OpenAI Agents / Copilot / deep_agents / opencode users had no path to create or update a pocket from chat. This PR pulls creation out into its own builder that runs against whatever provider the active agent already uses.

## What changed

- New `ee/cloud/pockets/builder/` package on the 4-file shape: domain, dto, prompts, providers, service.
- `providers.structured_call` dispatches to Anthropic tool_use, OpenAI `response_format=json_schema`, Ollama's native `/api/chat`, or a plain-text fallback for everything else. No new pip deps.
- Two LLM calls: a tiny classifier (Haiku on Anthropic) and then the spec builder on the agent's configured model. The classifier result hits the SSE stream as `intent.detected` so the UI has something to react to before the spec call comes back.
- `_run_agent_stream` dispatches through the builder before `pool.run`. Confidence is logged. Routing does not gate on it (captain's call).
- `build_context_block` no longer injects the MCP cloud preamble when intent is `pocket_create` — the builder owns that turn.

## Captain's locked defaults

1. Binary intent routing. Confidence is logged, never gated on.
2. No workspace-level provider override. Deferred to Phase 1.5.
3. The update path does not inject the current pocket spec. Also Phase 1.5.
4. `pocket.updated` reuses the existing `_push_replace` mutation shape. No new event surface.
5. OSS bus path `loop._publish_pocket_event` stays untouched in this PR.
6. `PocketSpec` Pydantic `model_validator` rejects both `widgets` and `ripple_spec` if both are set non-empty.

## Phase 2 deferred deletions

These stay live in this PR. The Phase 2 PR will remove them:
- `agents/sdk_mcp_pocket.py` `create_pocket` + `update_pocket` handlers
- `loop.py` `_publish_pocket_event` + `_create_pocket_and_session`
- `_POCKET_CREATION_CONTEXT` and `_POCKET_INTERACTION_CONTEXT` in `pocketpaw/api/v1/pockets.py`
- `agent_context.create_pocket_for_agent`

## Tests

- 11 builder unit tests pass in 0.3s.
- 1569 cloud tests pass (6 skipped, 1 pre-existing deselected).
- 13/13 importlinter contracts pass (9 existing + 4 new).

## Design

Design lives at `ee/docs/architecture/pockets-builder-design.md`.

## Test plan

- [ ] Run the builder end-to-end against an Anthropic-backed agent in staging.
- [ ] Confirm the builder falls through cleanly when the classifier returns `none` (normal chat still works).
- [ ] Confirm the frontend sees `stream_start, intent.detected, spec.building, pocket.created, chunk, stream_end` in that order.
- [ ] Spot check Codex and Ollama paths via the plain-text fallback.

### DIFF
--- a/ee/cloud/chat/agent_router.py
+++ b/ee/cloud/chat/agent_router.py
@@ -374,6 +374,78 @@ async def _run_agent_stream(
 
     full_text = ""
     cancelled = False
+
+    # ----------------------------------------------------------------------
+    # Builder dispatch — pockets-builder owns native pocket creation /
+    # update intent for every backend (Anthropic, OpenAI, Ollama, Codex,
+    # Copilot, deep_agents, opencode).  Runs BEFORE pool.run; if the
+    # builder finishes with success or error it short-circuits the rest of
+    # the stream.  When the classifier returns ``intent=none``, we fall
+    # through to the normal pool.run path.
+    # ----------------------------------------------------------------------
+    pocket_intent_handled = False
+    try:
+        async for builder_event, builder_done in _dispatch_builder(
+            ctx, body, instance, cancel_event, _drain_side_channel
+        ):
+            for ev in _drain_side_channel():
+                yield ev
+            if cancel_event.is_set():
+                cancelled = True
+                break
+            # Internal ``_sentinel`` marker tells the SSE loop the builder
+            # finished without a fresh frame to forward — used after the
+            # trailing ``chunk`` so we don't double-emit.
+            if builder_event[0] != "_sentinel":
+                yield builder_event
+                if builder_event[0] == "chunk":
+                    chunk_content = builder_event[1].get("content", "")
+                    if isinstance(chunk_content, str):
+                        full_text += chunk_content
+            if builder_done:
+                pocket_intent_handled = True
+                break
+    except Exception:
+        logger.exception(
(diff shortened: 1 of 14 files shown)


prakashUXtech:
Deferring this PR — it's currently CONFLICTING (DIRTY) against `ee`. After the RBAC stack landed (#1058 → #1059 → #1060 → #1061), plus #1057 $source resolver and #1055 multi-agent fix, the conflicts likely widened. Please rebase on current ee and re-request review when ready.

prakashUXtech:
Closing in favor of #1156.

This PR predates the OSS-EE namespace split (`ee.*` to `pocketpaw_ee.*`) and the pocket-specialist sprint. Every file targets a path that no longer exists, and `agent_router.py` — the file it patches — was rewritten since. A rebase would mean rewriting all 16 files anyway.

The design still holds value: a structured-output pocket-creation path for non-MCP backends is a real gap. #1156 tracks rebuilding it on the current layout, including the two defects the review surfaced (no audit log on the creation path, no injection scan on the classifier input). The domain models, provider adapters, and prompts here are reusable — #1156 references this branch.

PR #1041, opened 2026-05-01
fix(callback): validate OAuth2 state to prevent CSRF

### LINKED ISSUE #999: security: OAuth state param CSRF gap on /oauth/integrations/authorize
Surfaced during Cluster C Wave 3 (paw-enterprise PR #106 + pocketpaw #995) review. Pre-existing, not introduced by the PR — but the new per-connector Connect UI makes this URL much easier to reach.

## Problem
`GET /api/v1/oauth/integrations/authorize?service=google_drive` initiates the Google OAuth flow. The `state` parameter is not signed or bound to the user session, so an attacker can craft a URL with a known state and the callback will accept it. Classic OAuth CSRF vector.

Cluster C-2 added the UI button that triggers this URL. The underlying endpoint behaviour is unchanged.

## Fix direction
- Generate `state` server-side, sign it with a session-bound HMAC.
- Store the signed state in a short-lived session cookie or in-memory map keyed to the user.
- Reject callbacks whose `state` does not verify + match the stored value.
- Invalidate used state tokens on successful callback.

## Tests
- Valid flow still works (Connect UI → callback → token stored).
- Crafted state in callback without initiating authorize → 403.
- Replayed state after callback consumed → 403.
- State issued to user A reused by user B → 403.

Priority: P1. Cluster F is the direct client; any eventual fleet / customer-facing PawPrint flow that delegates OAuth should also benefit.

safwan027:
# MCP OAuth State Flow 

This document explains the OAuth `state` hardening used in `src/pocketpaw/mcp/manager.py`.

## Related Issue

Fixes #999

## Why this exists

PocketPaw now wraps OAuth callback state in a signed token to prevent:

- forged callback state values,
- cross-flow mixups,
- replay of previously valid callback states.

The implementation preserves SDK compatibility by carrying the original upstream state inside the signed payload.

## Terms

- **Upstream state**: the original `state` generated by the OAuth SDK/client flow.
- **Signed state token**: PocketPaw-generated state sent to provider/browser roundtrip.
- **Pending flow**: server-side one-time record keyed by internal `sid`.

## Top-to-bottom flow

1. OAuth SDK creates auth URL and calls `redirect_handler(auth_url)`.
2. PocketPaw extracts the original upstream `state` from query params.
3. PocketPaw creates a signed token payload containing:
   - `sid` (internal state id)
   - `uid` (user binding)
   - `prv` (provider/server binding)
   - `iat` and `exp` (issue/expiry)
   - `n` (random nonce)
   - `ups` (original upstream state)
4. PocketPaw stores `_PendingOAuthFlow` in `_oauth_pending[sid]`.
5. PocketPaw replaces outbound URL `state` with the signed token and broadcasts URL to UI.
6. Provider redirects callback with `code` + signed `state`.
7. Callback endpoint calls `set_oauth_callback_result(state, code)`.
8. PocketPaw verifies signature + expiry, loads pending flow by `sid`, validates bindings (`uid`, `prv`), and resolves future with `(code, upstream_state)`.
9. Pending state is removed (`pop`) to enforce one-time use.
10. `callback_handler()` resumes and returns `(code, returned_state)` to SDK, which continues token exchange and token storage.

## Why we keep upstream state

We intentionally preserve upstream state (`ups`) because SDK internals may expect the original state value for correlation.

So we use:

- **signed state** for browser/provider security boundary,
- **upstream state** for SDK compatibility after validation.

## Security checks performed

- HMAC signature verification for state token integrity.
- Expiration checks (`exp` and pending TTL checks).
- Binding checks against pending flow metadata.
- One-time consume semantics (`_oauth_pending.pop(...)`).

## Current limitations / future improvements

- `user_id` is currently set to `"local_user"` in this flow and can be upgraded to real authenticated session identity.
- Secret fallback (`dev-oauth-state-secret`) is for development; production should always set `POCKETPAW_OAUTH_STATE_SECRET`.
- Persisted distributed replay protection (e.g., Redis) could be added if multi-process callback handling is required.


ragini-pandey (commented):
## 🔴 Blocking / Must Fix

### 1. Signing secret bypasses `Settings` — violates project security convention

```python
def _state_signing_key() -> bytes:
    return (os.getenv("POCKETPAW_OAUTH_STATE_SECRET") or "dev-oauth-state-secret").encode()
```

Per the project's security rules (`CLAUDE.md` / `copilot-instructions.md`): *"Never hardcode credentials. Use `get_settings()` and store secrets via `credentials.py` (Fernet-encrypted). New secret fields must be added to `SECRET_FIELDS` in `credentials.py`."*

This implementation has two problems:
- It reads the env var with `os.getenv()` at call time instead of going through `get_settings()`.
- `oauth_state_secret` is not declared in `Settings` and not added to `SECRET_FIELDS` in `credentials.py`, so it is never encrypted at rest.

**Required changes:**
1. Add `oauth_state_secret: str = Field(default="", env="POCKETPAW_OAUTH_STATE_SECRET")` to `Settings` in `config.py`.
2. Add `"oauth_state_secret"` to `SECRET_FIELDS` in `credentials.py`.
3. Change `_state_signing_key()` to call `get_settings()` inside the function body:
   ```python
   def _state_signing_key() -> bytes:
       from pocketpaw.config import get_settings
       return get_settings().oauth_state_secret.encode()
   ```

---

### 2. Hardcoded fallback key renders the HMAC useless in the common case

```python
return (os.getenv("POCKETPAW_OAUTH_STATE_SECRET") or "dev-oauth-state-secret").encode()
```

`"dev-oauth-state-secret"` is public. Any attacker who knows it (anyone who reads this PR) can forge valid signed state tokens. If the env var is not set — which is the case for most users who have not read the deployment docs — the entire CSRF protection is bypassed.

The fallback should either:
- **Raise** in non-development environments:
  ```python
  secret = get_settings().oauth_state_secret
  if not secret:
      raise RuntimeError(
          "POCKETPAW_OAUTH_STATE_SECRET must be set. "
          "Run: pocketpaw config set oauth_state_secret $(openssl rand -hex 32)"
      )
  ```
- Or **auto-generate and persist** a random key on first use (similar to how other secret fields are handled).

The PR description itself acknowledges this as a limitation but frames it as "for development". It must not land without a clear path that does not allow silent production degradation.

---

### 3. No tests for the security-critical paths

The test file only updates the two existing happy-path tests. None of the new security checks have test coverage:

| Scenario | Covered? |
|---|---|
| Expired state token (`exp` in the past) | ❌ |
| Tampered signature (bit-flip in sig) | ❌ |
| `uid` binding mismatch | ❌ |
| `prv` binding mismatch | ❌ |
| TTL-expired pending flow (`created_at` stale) | ❌ |
| Replay: second call with same valid token | ❌ |
| Malformed token (missing `.` separator) | ❌ |
| `_verify_state_token` with non-JSON payload | ❌ |

A security fix with no negative test cases is not verifiably correct. At minimum the expiry, signature tampering, and binding mismatch cases need tests.

---

### 4. `uid` binding provides no security since it is always `"local_user"`

```python
user_id = "local_user"
signed_state, state_id = _create_state_token(
    user_id=user_id, ...
)
```

And in the callback:
```python
if payload.get("uid") != entry.user_id or payload.get("prv") != entry.provider:
    ...
    return False
```

Both sides of the `uid` comparison are always `"local_user"`. An attacker who forges a valid token (e.g., using the fallback key from issue #2) could set `uid = "local_user"` and pass the binding check. This means the `uid` binding check only works correctly once the secret fallback issue (#2) is fixed **and** real user identity is wired in.

This alone does not block merge if issues #1 and #2 are fixed, but the binding check must be documented as intentionally deferred (with a TODO tied to a follow-up issue), not silently ineffective.

---

## 🟡 Logic / Correctness

### 5. Memory leak: abandoned OAuth flows are never cleaned up

```python
_oauth_pending: dict[str, _PendingOAuthFlow] = {}
```

If a user opens an OAuth flow and never completes it (closes the browser tab, network error, etc.), the `_PendingOAuthFlow` entry stays in `_oauth_pending` indefinitely. The TTL check in `set_oauth_callback_result` only runs when a callback arrives — it cannot clean up flows that never get a callback.

A periodic cleanup task or a weak reference + finalizer is needed. Simplest approach:

```python
async def _cleanup_expired_oauth_flows() -> None:
    now = time.time()
    stale = [k for k, v in _oauth_pending.items()
             if (now - v.created_at) > _OAUTH_STATE_TTL_SECONDS]
    for k in stale:
        _oauth_pending.pop(k, None)
```

This can be called at the start of `redirect_handler` (which is already async) with negligible overhead.

---

### 6. `_flow_state["state"]` stores `state_id` but `callback_handler` checks it against `_oauth_pending`

```python
# redirect_handler (new code):
_flow_state["state"] = state_id          # stores the internal sid

# callback_handler (new code):
state = _flow_state.get("state")
if not state or state not in _oauth_pending:   # checks by sid — OK
    raise RuntimeError(...)
pending = _oauth_pending[state]
if pending.server_name != config.name:         # second guard
    ...
```

This is correct — `state_id` is the dict key — but it is subtle. Before this PR, `_flow_state["state"]` stored the raw upstream state. Now it stores the internal `sid`. The `pending.server_name != config.name` check in `callback_handler` replicates the same binding that `set_oauth_callback_result` already enforces, which is fine as defence-in-depth, but a comment explaining the double-check would prevent future contributors from removing it as "redundant".

---

## 🟠 Code Quality / Nits

### 7. Unrelated whitespace changes — `models.py` and `server.py`

```diff
# models.py: trailing whitespace on blank line inside a method
+            
# server.py: two extra blank lines before the singleton comment
+
+
```

Neither of these is in a file touched by the security fix. Ruff's `W291` (trailing whitespace) would flag the `models.py` change. Please revert these hunks; they add diff noise and may fail CI lint.

---

### 8. `_verify_state_token` catches `Exception` too broadly

```python
def _verify_state_token(state_token: str) -> dict[str, Any] | None:
    try:
        ...
    except Exception:
        return None
```

A bare `except Exception` here swallows real programming errors (e.g., `AttributeError` from a typo in the payload field name, `TypeError` from wrong key type) and makes debugging very hard. Scope the catch to the expected parsing failures:

```python
except (ValueError, KeyError, UnicodeDecodeError):
    return None
```

`binascii.Error` (from bad base64) is a subclass of `ValueError`, so it is covered.

safwan027:
I will check on this, and create a PR.

PR #1057, opened 2026-05-05
feat(ripple): pocket $source resolver (v1)

DevRohit06:
## Summary

Adds a server-side resolver so pocket `rippleSpec.state` can declare `{"$source": "<name>", ...args}` markers that get replaced with live workspace data on read, instead of the LLM fabricating literal arrays. Mongo stores markers verbatim; resolution happens in `pockets.service.get` after the access check.

- **New modules** — `ee/cloud/ripple_resolver.py` (walker + registry, no cloud-domain imports) and `ee/cloud/ripple_sources.py` (concrete sources, single point of contact with entity modules).
- **v1 sources** — `workspace.pockets` (id, name, type, icon, color) and `workspace.members` (id-only; richer fields v2). Both tenancy-scoped per ee/cloud rule 7.
- **Failure model** — `_resolve_marker` swallows source-level exceptions and returns `None`; `service.get` wraps the resolver call in try/except and falls back to the raw spec. Resolution can never raise out to the router.
- **Prompt update** — both creation and interaction agents now learn about `$source` via `_STATE_SOURCES_BLOCK`. The block names the v1 allowlist and tells the agent to use literal arrays only for canvas-local UI state (drafts, counters), never for fabricated business data.
- **Out of scope** — `pocket.tool_call` source, client write-back, `action: tool` button handlers, realtime invalidation, caching. Tracked for v2.

## Test plan

- [x] `uv run pytest tests/cloud/test_ripple_resolver.py tests/cloud/test_pocket_service_resolver.py tests/cloud/test_pocket_prompt_state_sources.py tests/cloud/test_pocket_prompts_single_source.py` → 27 passed
- [x] Broader regression: `uv run pytest tests/cloud/ -k "pocket or ripple" --ignore=tests/cloud/extraction` → previously-green tests still green
- [x] `uv run ruff check` clean on touched files
- [x] `uv run mypy ee/cloud/ripple_resolver.py ee/cloud/ripple_sources.py` clean
- [ ] Manual: create a pocket via the cloud chat agent with the prompt "make me a pocket that lists all my other pockets" — verify the agent emits `{"$source": "workspace.pockets"}` and the rendered pocket shows real workspace pockets

## Notes for review

- `e105687e` introduced an encoding corruption (em-dashes mojibakeed during a polish commit). Reverted by `cbb6a8f8` and re-applied cleanly in `7d64291e`. The branch tip has correct UTF-8; the bad SHA remains in history but nothing reads from historical commits.
- `agent_view` (the agent-facing pocket read) intentionally does NOT resolve markers — agents must see raw `$source` values to preserve them on edit, otherwise edits would bake snapshots of live data into the spec. Documented in the function's docstring.
- `access_via_share_link` also does not resolve; share-link viewers have no auth context to build a `ResolveCtx`. Marked with a `# no-resolve` comment for v2 follow-up.


DevRohit06:
Added one more commit to this branch — `65a7f87d feat(ripple): teach prompts the new composite layouts + add no-invented-widgets rule`. Scope is the system prompts only, no new backend code:

- Adds the new ripple composite layouts (`comparison-layout`, `entity-detail`, `form-layout`, `wizard-layout`, `checklist-layout`, `report-layout`, `invoice-layout`, `order-status`, `map`, `location-picker`) to WIDGET_CATALOG, USE-THE-WIDGET, FULL-PANE, and COMPOSITION_COOKBOOK rules. Dashboard variants are intentionally deferred.
- New `NO_INVENTED_WIDGETS_RULE` block — the registry is closed; the rule lists common invention modes (pluralizing, compounding like `metric-card` / `kpi-tile`) and the right typed widget for each rebuild antipattern. Spliced between WIDGET_CATALOG and WIDGET_SPEC_TOOL_RULE.
- Fixes inlined `table` examples in CANONICAL_SHAPES + the Todos creation example to use `rows:` (manifest's documented prop) instead of the `data:` runtime alias, so prompts and manifest agree.
- One catalog typo fix (`entity-details` → `entity-detail`).

This complements the new ripple PR (https://github.com/qbtrix/ripple-iui/pull/34) that ships the corresponding widgets and manifest entries — agents won't see them unless both ship together.

prakashUXtech:
Pushed two cross-workspace tenancy tests per review feedback. They prove the find query is exact-keyed by ctx.workspace_id (not substring-present), and that swapping ctx.workspace_id from w1 to w2 causes the query to track — proves no spec-level value can override server-built ctx. Test file note: `test_pocket_prompts_single_source.py` is present in the diff (landed via #1056 merge into this branch).

PR #1060, opened 2026-05-06
feat(rbac): add cloud-native require_plan_feature dependency

prakashUXtech:
Routes that require business or enterprise features can now declare that in the dependency list instead of hand-rolling the check inside each handler.

Three files changed:

`workspace/service.py` gets a new `get_workspace_plan(workspace_id)` helper. It loads the Workspace doc and returns the plan string. If the workspace is not found (e.g. bad ID, soft-deleted), it returns "team" rather than crashing, which means the gate fails closed on restricted features without a 500.

`_core/deps.py` gets `require_plan_feature(feature)`. It grabs the workspace ID from `current_workspace_id`, calls the service helper (lazy import, Beanie stays out of deps.py), and compares against `PLAN_FEATURES` from `pocketpaw.ee.guards.abac`. On deny it raises `Forbidden("plan.feature_denied", ...)` with the minimum required plan in the message.

```python
@router.get(
    "/fabric/types",
    dependencies=[
        Depends(require_plan_feature("fabric")),
        Depends(require_action_any_workspace("fabric.read")),
    ],
)
```

`shared/deps.py` re-exports it so import paths that already exist keep working.

10 new tests in `tests/cloud/test_plan_feature_gate.py`. Covers fabric on team/business/enterprise, instinct on business/enterprise, workspace-not-found fallback, and the error message content. All pass. 2 pre-existing failures on the ee branch are unrelated.

### DIFF
--- a/ee/cloud/_core/deps.py
+++ b/ee/cloud/_core/deps.py
@@ -11,6 +11,8 @@ delegates to ``pocketpaw.ee.guards`` (the platform-wide RBAC package) for
 the actual policy lookup. We translate platform ``GuardForbidden``
 exceptions to cloud-native ``Forbidden`` so the standard error envelope
 applies.
+
+Changes: added require_plan_feature dependency for plan-tier feature gating.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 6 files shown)


prakashUXtech:
Pushed plan-feature wiring per stack-level review concern (commit c7bba04d). Fabric and Instinct routers now have `Depends(require_plan_feature("fabric"))` and `Depends(require_plan_feature("instinct"))` at the constructor level alongside the existing `require_license`.

Closes the plan-tier bypass: previously a team-plan member who passed the #1059 workspace RBAC check still hit business+ Fabric/Instinct for free.

Note: 35 pre-existing test failures in `tests/cloud/test_ee_instinct.py` and `tests/cloud/test_ee_fabric_list_endpoints.py` exist on `ee` post-#1059 (fixtures don't seed auth context for the new workspace RBAC guards). Independent of this PR — present before my changes too. Needs a follow-up to update the test fixtures.

Issue #1064, opened 2026-05-07
fix(tests): seed auth context in Fabric/Instinct fixtures (35 broken tests post-#1059)


prakashUXtech:
## Background

PR #1059 added workspace RBAC guards to Fabric, Instinct, and agent knowledge endpoints. The route-test fixtures in `tests/cloud/test_ee_instinct.py` and `tests/cloud/test_ee_fabric_list_endpoints.py` build a FastAPI app and include the router, but they don't override `current_active_user` or `current_workspace_id`. Every route hit gets 403 against the new guards.

These tests were green before #1059. The RBAC PR didn't update them.

## Reproduce

```
uv run pytest tests/cloud/test_ee_instinct.py tests/cloud/test_ee_fabric_list_endpoints.py
# 35 failed, 38 passed
```

The failures are pre-existing as of this issue (independent of PR #1060's plan-feature wiring landed today).

## Fix

Add dep overrides to each `client` / `test_app` fixture so tests run with a seeded user + workspace. The pattern to mirror is in `tests/cloud/test_rbac_routes.py` (added by #1061): `app.dependency_overrides[current_active_user]` plus `app.dependency_overrides[require_license]`.

## Files to update

- `tests/cloud/test_ee_instinct.py` — `test_app` fixture (around line 30)
- `tests/cloud/test_ee_fabric_list_endpoints.py` — `client` fixture (around line 22)
- `tests/cloud/test_e2e_connector_to_fabric.py` — likely shares the issue, verify

## Related

- Companion issue: pytest is not in the PR Quality Gate workflow, which is why these failures shipped silently — see the CI follow-up issue.

prakashUXtech:
Shipped in #1067, which merged to dev. The PR targeted dev rather than the default branch, so this never auto-closed. Closing it now.

PR #1068, opened 2026-05-07
ci(tests): install enterprise extras so ee/cloud tests can import beanie

prakashUXtech:
## What

Follow-up to #1066. The newly-landed `tests.yaml` workflow ran on its own PR and errored at pytest collection:

```
ee/cloud/auth/core.py:23: in <module>
    from beanie import PydanticObjectId
E   ModuleNotFoundError: No module named 'beanie'
```

`beanie`, `motor`, and `fastapi-users[beanie,oauth]` live in `[project.optional-dependencies].enterprise`. `uv sync --dev` skips them. Switching to `uv sync --dev --all-extras` pulls everything ee/cloud tests need.

## Trade-off

`--all-extras` is heavier than the minimum (`--extra enterprise --extra soul`). I picked it for full coverage so future ee additions don't re-trigger this same gap. uv's cache makes the second run fast.

## Closes

Unblocks the gate added in #1066. Without this, every future PR shows red.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,6 +1,9 @@
 # Tests — runs the pytest suite on every PR and on pushes to ee/main/dev.
 # Created: 2026-05-07 — Added pytest job to PR Quality Gate so broken tests
 # can no longer slip through CI on PRs targeting ee. Closes #1065.
+# Updated: 2026-05-07 — install enterprise + soul extras so ee/cloud tests
+# can import beanie/motor/fastapi-users. `uv sync --dev` alone misses them
+# (they live under [project.optional-dependencies].enterprise).
 name: Tests
 
 on:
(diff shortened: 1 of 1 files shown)


prakashUXtech:
Update: `--all-extras` got past the beanie ImportError. Now the suite collects but errors on a different pre-existing gap — `AttributeError: module 'pocketpaw.security.guardian' does not have the attribute 'get_settings'` across guardian + LLM-classification tests. Same shape as #1064: tests monkeypatch a module attribute that doesn't exist (or got renamed). Worth a follow-up issue mirroring the #1064/#1067 pattern. Recommend merging this PR anyway since the beanie fix is correct + necessary; the next gap is independent.

prakashUXtech:
Closed by #1074 (merged 2026-05-08). `uv sync --dev --all-extras` now installs the enterprise extras so ee/cloud tests can collect cleanly.

PR #1078, opened 2026-05-08
Notifications feat and workspace channels with permissions 

### LINKED ISSUE #1076: Backend notification service missing event creation and room-aware notification support
# Description

The backend notification system does not consistently generate notifications for supported user actions and lacks proper room-aware navigation metadata. This causes missing notifications, broken navigation targets, and inconsistent realtime behavior.

# Problems

- Notifications are not consistently created for:
  - messages
  - mentions
  - reactions
  - group invites
  - workspace invites
- Group notification creation flow can crash due to variable ordering
- Notification payloads lack proper room_id support
- Realtime notification events are not fully synchronized with read-state changes

# Required Changes

- Integrate notifications_service.create() into:
  - message_service.py
  - group_service.py
  - workspace/service.py
- Add room_id to NotificationSource
- Fix notification creation ordering in group flows
- Emit proper WebSocket events for:
  - new notifications
  - read updates
  - clear-all updates
- Ensure unread count queries remain accurate

### LINKED ISSUE #1077: Implement backend support for private/public channels and membership-based visibility
# Description

The backend chat system currently lacks complete support for channel visibility controls and private channel access restrictions. Private channels should only be visible to members and should require invitation-based access.

# Problems

- Channel documents do not fully support visibility configuration
- Private channels are visible to non-members
- Users can directly join private channels
- Channel visibility cannot be updated after creation
- Private channels appear in mention suggestions

# Required Changes

- Add visibility field across:
  - models
  - DTOs
  - schemas
  - domain objects
- Restrict private channel visibility to members only
- Block direct joining of private channels
- Support updating visibility through update APIs
- Exclude private channels from public channel suggestions
- Ensure backward compatibility for existing channels

AMRITESH240304:
## What does this PR do?

Adds a complete notification system for chat and workspace events, channel visibility (public/private) with access control, and granular member
  permission roles with runtime enforcement.

## Related Issue

Fixes #1076 
Fixes #1077 

## Changes Made

 - `ee/cloud/notifications/service.py`: Create, list, mark-read, mark-all-read, unread-count operations with WebSocket emits                         
  - `ee/cloud/notifications/router.py`: REST endpoints — GET /notifications, GET /notifications/unread-count, PATCH /notifications/{id}/read, PATCH 
  /notifications/read-all                                                                                                                           
  - `ee/cloud/notifications/domain.py`: Notification + NotificationSource value objects with room_id for navigation
  - `ee/cloud/notifications/dto.py`: NotificationOut response schema, CreateNotificationRequest input schema                                          
  - `ee/cloud/chat/message_service.py`: Wire notifications_service.create() on send for DM/group messages, mentions, and reactions; enforce           
  post_no_media role — block file attachments for restricted users                                                                                  
  - `ee/cloud/chat/group_service.py`: Wire notification on add_members(); add visibility field with private-channel access filtering; accept and store
   post_no_media member role; map restriction roles to GroupRole.MEMBER for basic post access                                                       
  - `ee/cloud/chat/schemas.py`: Add visibility to CreateGroupRequest/UpdateGroupRequest; extend AddGroupMembersRequest/UpdateMemberRoleRequest role
  literal to accept post_no_media                                                                                                                   
  - `ee/cloud/chat/domain.py`: Add MemberRole literal values for post_no_media
  - `ee/cloud/chat/dto.py`: Expose visibility in group wire dict                                                                                      
  - `ee/cloud/models/group.py`: Add visibility field with pattern validation; extend MemberRole literal                                               
  - `ee/cloud/workspace/service.py`: Wire notification on workspace invite

## How to Test

 1. Notifications: Send a DM to another user → verify notification appears in their bell dropdown with correct room_id navigation. Send a group    
  2. Channel visibility: Create a private channel → verify non-members cannot see it in the workspace listing. Create a public channel → verify all 
  members can see and join it.                                                                                                                      
  3. post_no_media role: Add a member with "Chat (no attachments)" role → verify they can send text messages but file attachments return 403 
  group.attachments_disabled.                                                                                                                       
  4. Test suite: Run pytest tests/cloud/chat/ -x -k "not test_send_message_user_and_broadcast_mention_dedupes" — all 69 tests pass (1 pre-existing
  dedup failure excluded).  

## Checklist

- [x] PR targets `ee` branch (not `main`)

### DIFF
--- a/ee/cloud/_core/realtime/audience.py
+++ b/ee/cloud/_core/realtime/audience.py
@@ -117,6 +117,8 @@ class AudienceResolver:
             return await self._group(d["group_id"])
         if t == "message.sent":
             return [d["sender_id"]]
+        if t == "thread.reply":
+            return await self._group(d.get("group_id", ""))
         if t == "message.ui_state.updated":
             # Group-context messages fan out to every group member so a peer
             # viewing the same room sees the kanban update live. Pocket /
--- a/ee/cloud/chat/domain.py
+++ b/ee/cloud/chat/domain.py
@@ -20,7 +20,7 @@ from typing import Literal
 # Mirror persistence-layer literals
 ContextType = Literal["pocket", "group", "session"]
 PocketRole = Literal["user", "assistant", "system"]
-MemberRole = Literal["view", "edit", "admin"]
+MemberRole = Literal["view", "edit", "post_no_media", "admin"]
 GroupType = Literal["public", "private", "dm", "channel"]
 
 
--- a/ee/cloud/chat/domain.py
+++ b/ee/cloud/chat/domain.py
@@ -82,6 +82,7 @@ class Group:
     message_count: int
     created_at: datetime
     updated_at: datetime
+    visibility: str = "public"  # "public" | "private" — for channels
 
 
 @dataclass(frozen=True)
--- a/ee/cloud/chat/domain.py
+++ b/ee/cloud/chat/domain.py
@@ -83,6 +83,7 @@ class Group:
     created_at: datetime
     updated_at: datetime
     visibility: str = "public"  # "public" | "private" — for channels
+    active_threads: tuple[str, ...] = ()
 
 
 @dataclass(frozen=True)
@@ -106,6 +107,8 @@ class Message:
     mentions: tuple[Mention, ...] = field(default_factory=tuple)
     reply_to: str | None = None
     thread_count: int = 0
+    thread_id: str | None = None
+    is_thread_parent: bool = False
     attachments: tuple[Attachment, ...] = field(default_factory=tuple)
     reactions: tuple[Reaction, ...] = field(default_factory=tuple)
     edited: bool = False
--- a/ee/cloud/chat/dto.py
+++ b/ee/cloud/chat/dto.py
@@ -24,6 +24,7 @@ from ee.cloud.chat.schemas import (  # noqa: F401
     AddGroupAgentRequest,
     AddGroupMembersRequest,
     CreateGroupRequest,
+    CreateThreadRequest,
     CursorPage,
     EditMessageRequest,
     GroupResponse,
@@ -63,6 +64,8 @@ def message_to_wire_dict(m: Message, *, parent: Message | None = None) -> dict[s
         "replyTo": m.reply_to,
         "replyPreview": _reply_preview(parent) if m.reply_to else None,
         "threadCount": m.thread_count,
+        "threadId": m.thread_id,
+        "isThreadParent": m.is_thread_parent,
         "attachments": [
             {"type": a.type, "url": a.url, "name": a.name, "meta": dict(a.meta)}
             for a in m.attachments
@@ -161,6 +164,7 @@ def group_to_wire_dict(
         "memberRoles": dict(group.member_roles),
         "agents": populated_agents,
         "pinnedMessages": list(group.pinned_messages),
+        "activeThreads": list(group.active_threads),
         "archived": group.archived,
         "lastMessageAt": iso_utc(group.last_message_at),
         "messageCount": group.message_count,
--- a/ee/cloud/chat/dto.py
+++ b/ee/cloud/chat/dto.py
@@ -153,6 +153,7 @@ def group_to_wire_dict(
         "slug": group.slug,
         "description": group.description,
         "type": group.type,
+        "visibility": group.visibility,
         "icon": group.icon,
         "color": group.color,
         "owner": group.owner,
--- a/ee/cloud/chat/group_service.py
+++ b/ee/cloud/chat/group_service.py
@@ -67,6 +67,7 @@ def _group_doc_to_domain(doc: _GroupDoc) -> _GroupDomain:
         icon=doc.icon,
         color=doc.color,
         type=doc.type,
+        visibility=getattr(doc, "visibility", "public"),
         members=tuple(doc.members),
         member_roles=tuple(doc.member_roles.items()),
         agents=tuple(_group_agent_to_domain(a) for a in doc.agents),
@@ -294,6 +295,7 @@ async def _group_response(group: _GroupDoc) -> dict:
         "slug": group.slug,
         "description": group.description,
         "type": group.type,
+        "visibility": getattr(group, "visibility", "public"),
         "icon": group.icon,
         "color": group.color,
         "owner": group.owner,
@@ -371,14 +373,19 @@ async def _populate_lookups_for_domain_groups(
 
 
 async def _list_visible_in_workspace(workspace_id: str, user_id: str) -> list[_GroupDomain]:
-    """Public/channel groups in the workspace + private/DM groups
-    the user is a member of. Excludes archived."""
+    """Channels (public) and public groups in the workspace + private/DM groups
+    the user is a member of. Excludes archived. Private channels are only
+    surfaced to members who have been granted access."""
     docs = await _GroupDoc.find(
         {
             "workspace": workspace_id,
             "archived": False,
             "$or": [
-                {"type": {"$in": ["public", "channel"]}},
+                # Public groups and channels visible to all workspace members
+                {"type": "public"},
+                # Public channels (visibility not set, or set to "public")
+                {"type": "channel", "visibility": {"$ne": "private"}},
+                # Private groups, DMs, and private channels: membership required
                 {"members": user_id},
             ],
         }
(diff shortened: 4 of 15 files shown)


ragini-pandey (commented):
## 🔴 Blocking / Must Fix

### 1. Notification flood on every `send_message` — `ee/cloud/chat/message_service.py`

```python
notif_tasks = [
    notifications_service.create(
        workspace_id=..., recipient=member, ...
    )
    for member in notif_recipients   # every member except the sender
]
await asyncio.gather(*notif_tasks)
```

One notification document is inserted per group member on **every** message send. In a workspace channel with 200 members this fires 200 individual writes on each message. This is O(N) DB inserts with no bulk path and no user preference check. At meaningful scale this will make send latency unacceptable.

**Required changes:**
- Add a `create_bulk(recipients: list[str], ...)` method in `notifications_service` that does a single `insert_many`.
- Add a notification preference model or at minimum a per-user "mute group" flag so members can opt out.
- Consider skipping broadcast notifications for `@channel`/`@everyone` messages — those already have the unread-bump path.

---

### 2. `unread_count` fetches all documents to count — `ee/cloud/notifications/router.py`

```python
notes = await notifications_service.list_for_user(
    ctx.user_id, unread=True, limit=9999   # ← anti-pattern
)
return {"count": len(notes)}
```

`limit=9999` is a hack. If a user has >9999 unread notifications the count is silently wrong. Replace with a proper `count()` query in `notifications_service`:

```python
# In service.py
async def count_unread(user_id: str) -> int:
    return await _NotificationDoc.find(
        {"recipient": user_id, "read": False}
    ).count()
```

---

### 3. N+1 query per thread in `get_active_threads` — `ee/cloud/chat/message_service.py`

```python
for tid in group.active_threads:
    reply_count = await _MessageDoc.find(..., thread_id=tid, ...).count()
    last_reply  = await _MessageDoc.find(..., thread_id=tid, ...).sort(...).limit(1).to_list()
```

For a group with 50 active threads this issues 100 serial queries. Use an aggregation pipeline instead:

```python
pipeline = [
    {"$match": {"context_type": "group", "thread_id": {"$in": active_thread_ids}, "deleted": False}},
    {"$group": {
        "_id": "$thread_id",
        "reply_count": {"$sum": 1},
        "last_reply_at": {"$max": "$createdAt"},
    }},
]
```

---

### 4. `group_id` URL param is never validated in `get_thread_messages` — `ee/cloud/chat/router.py` + `message_service.py`

```python
@_licensed.get("/groups/{group_id}/threads/{thread_id}/messages")
async def get_thread_messages(group_id: str, thread_id: str, ...):
    return await message_service.get_thread_messages(thread_id, user_id, cursor, limit)
```

The service ignores `group_id` entirely and derives the group from the parent message's stored `group` field. A caller can pass any valid `group_id` in the URL (including a group they are a member of) and access threads belonging to a different group they are NOT a member of, as long as they know `thread_id`. At minimum, assert `parent.group == group_id` in the service before the membership check.

---

### 5. Bare `except Exception` swallows real errors — `message_service.py`

```python
try:
    sender_doc = await _UserDoc.get(PydanticObjectId(user_id))
    sender_name = sender_doc.full_name or sender_doc.email or "Someone"
except Exception:       # ← too broad
    sender_name = "Someone"
```

This silently swallows database connection failures, network timeouts, and auth errors in the hot send path. Scope the catch to `(ValueError, AttributeError)` or the specific `NotFound` exception; let unexpected errors propagate.

---

## 🟡 Logic Bugs / Correctness

### 6. `close_thread` permission check passes silently for deleted/missing messages

```python
if group.owner != user_id and group.member_roles.get(user_id) != "admin":
    msg = await _MessageDoc.get(PydanticObjectId(thread_id))
    if msg and msg.deleted or msg.group != group_id:   # msg is None → no raise
        raise Forbidden(...)
```

If the thread parent document is deleted or not found (`msg is None`), the entire inner `if` is skipped and a non-admin non-owner successfully passes the permission gate. The correct logic:

```python
if not msg or msg.deleted or str(msg.group) != group_id:
    raise NotFound("thread", thread_id)
if msg.sender != user_id:
    raise Forbidden("thread.not_authorized", "...")
```

---

### 7. Duplicate `/notifications/read-all` and `/notifications/clear` endpoints — `notifications/router.py`

Both `POST /notifications/read-all` (new) and `POST /notifications/clear` (existing, unchanged) call `notifications_service.clear_all(ctx.user_id)`. One of them must go. The PR description specifies `PATCH /notifications/read-all` but the implementation uses `POST`. Decide on one route and method; remove the other.

---

### 8. Breaking change: `workspace/service.py` invite source `id` changed from ObjectId to token

```python
# Before
source=NotificationSource(type="invite", id=invite.id)
# After (this PR)
source=NotificationSource(type="invite", id=invite.token, room_id=invite.group_id)
```

`invite.id` is the MongoDB document ObjectId used elsewhere to look up the invite. Changing it to `invite.token` without a migration or client contract update is a silent breaking change. If the intent is to expose the token for deep-link navigation, add `token` as a separate field on `NotificationSource` rather than overwriting `id`.

---

### 9. `post_no_mention` listed in comment but missing from `MemberRole` — `ee/cloud/models/group.py`

```python
#   "post_no_mention" — can post but @mentions are blocked
#   "post_no_media"   — can post but file attachments are blocked
MemberRole = Literal["view", "edit", "post_no_media", "admin"]  # post_no_mention is absent
```

Remove the `post_no_mention` comment line, or add the role and its enforcement. Leaving an undeclared role in a doc comment creates confusion for future contributors.

---

### 10. `is_thread_parent` set without `thread_id` on the parent message — `message_service.py`

```python
if not msg.is_thread_parent:
    msg.is_thread_parent = True
    await msg.save()
    # msg.thread_id is never set
```

The parent message's own `thread_id` field is `None` after `create_thread`. If any query relies on `thread_id` being set for parent messages (e.g. to navigate to the thread root from a reply), it will fail. Set `msg.thread_id = str(msg.id)` when marking a message as a thread parent.

---

## 🟠 Type Safety / Code Quality

### 11. `visibility: str` in domain object loses type safety — `ee/cloud/chat/domain.py`

```python
visibility: str = "public"  # "public" | "private" — for channels
```

The comment describes a `Literal`. Use it:

```python
from typing import Literal
visibility: Literal["public", "private"] = "public"
```

Consistent with how `MemberRole`, `GroupType`, and `ContextType` are declared in the same file.

---

### 12. `_role_for` return type annotation updated but callers not fully audited

`_role_for` now returns `Literal["owner", "admin", "edit", "post_no_media", "view", "none"]`, and `resolve_group_role` maps `"post_no_media"` → `GroupRole.MEMBER`. However `_add_members_doc` still checks:

```python
if role in ("admin", "view", "post_no_media"):
    doc.member_roles[mid] = role
elif role == "edit" and mid in doc.member_roles:
    doc.member_roles.pop(mid, None)
```

The `"edit"` branch only fires when the incoming role is literally `"edit"`. If a member is being changed **from** `"post_no_media"` **to** `"edit"`, the `elif` fires and removes the entry (correct). But if changing from `"edit"` to `"post_no_media"`, the `if` branch fires and stores it (correct). This logic is sound but the asymmetry between stored roles and "absence means edit" needs a comment explaining the invariant.

---

### 13. Pre-existing test failure claim in checklist warrants investigation

> Run `pytest tests/cloud/chat/ -x -k "not test_send_message_user_and_broadcast_mention_dedupes"` — all 69 tests pass (1 pre-existing dedup failure excluded).

The excluded test name contains "dedupes". The notification blast added in this PR creates a new notification for every member on every send. It is plausible that this PR itself is causing the deduplication test to fail (by adding extra notification records that break an assertion). This must be verified — do not ship with a known failing test labelled as pre-existing without confirmation it was failing on `ee` before this branch.

---

## ℹ️ Nits / Minor

- **`audience.py` — `thread.reply` handler uses `.get("group_id", "")` with empty-string fallback.** An empty group_id will produce an empty audience rather than an error. Use `.get("group_id")` and guard with `if not group_id: return []`.
- **`suggest_channels` now silently excludes private channels** — the docstring should say so to avoid surprising callers using this for admin tooling.
- **`CreateThreadRequest` has no `min_length` guard on `message_id`** — any non-empty string passes schema validation. Add `Field(..., min_length=1)` at least.
- **Thread endpoints in the router are undocumented in the `__all__` export list in `message_service.py`** — `close_thread` and `get_thread_messages` are added but `get_active_threads` is still missing from `__all__`.

Wait — checking the diff: `get_active_threads` IS in `__all__`. Disregard this nit.

PR #1069, opened 2026-05-07
feat(ripple/agents): pocket specialist subagent + slim chat prompt + cache-friendly ordering

DevRohit06:
## Summary

Two architectural shifts that work together to cut chat-mode prompt cost from ~12k to ~1.7k tokens per turn while improving correctness on pocket-mode interactions.

**1. Pocket specialist subagent** — heavy `POCKET_*_PROMPT_MCP` text moves off the main chat agent onto a dedicated subagent invoked via Claude Agent SDK's built-in `Agent` tool. Mutation tools (`create_pocket`, `update_pocket`, `add_widget`, `update_widget`, `remove_widget`) are filtered off the main agent's allowlist and live only on the specialist — architecture is enforced structurally, not just by instruction. Read-only `get_pocket` / `list_pockets` stay on the main agent for conversational queries.

**2. Slim inline ripple prompt + on-demand catalog** — `INLINE_RIPPLE_SYSTEM_PROMPT` cut from ~9k to ~1.7k tokens. Six core widgets (`text`, `heading`, `stat`, `button`, `table`, `flex`) named in the slim prompt; everything else (charts, kanbans, sparklines, gauges, timelines) deferred behind a new `get_inline_widget_help` MCP tool that returns the widget catalog payload on demand.

**3. Cache-enable system prompt** — `build_context_block` reorders so static prompt content leads, dynamic `<scope>` / `<participants>` / `<current-pocket>` tags trail. The static prefix now hits Anthropic's prompt cache.

Also bundled (earlier on the branch):
- Ripple resolver gets `.where` / `.whereIn` / `.sortBy` / `.limit` / `.reverse` array methods + bracket indexing — fixes the dead-binding pattern where filter/sort controls couldn't actually filter/sort.
- `ripple_normalizer` strips `entity-detail.props.actions[]` items lacking handlers, lifts `on_click → actions` for items using the wrong field — fixes the dead-button pattern.
- `_design.py` prompt teaches the new resolver methods + an explicit "interactive elements need handlers" rule.

Backend gating: the specialist mechanism is Claude-only (uses `ClaudeAgentOptions.agents` + the built-in `Agent` tool). Other backends (`openai_agents`, `google_adk`, `codex_cli`, `opencode`, `copilot_sdk`, `deep_agents`) fall back to the pre-split path with the heavy pocket prompt inlined — they still work, just without the token savings. Universal backend-agnostic version is the planned next step.

[...]

PR #1084, opened 2026-05-09
feat(deep_agents): fix Responses-API regression + skills/memory plumbing

DevRohit06:
## Summary

Stacked on #1083. Three changes against `agents/deep_agents.py` that the bump unlocks:

| | What | Why |
|---|---|---|
| 1 | Add `langchain-openai>=1.2.0,<2.0.0` to `deep-agents` extra | Broken pre-existing path: `deepagents` 0.5.x only pulls `langchain-anthropic` + `langchain-google-genai`, so the `openai`/`openai_compatible`/`openrouter`/`litellm` provider branches in `_build_model()` raise `ImportError: Initializing ChatOpenAI requires the langchain-openai package` at runtime. |
| 2 | Force `use_responses_api=False` for non-OpenAI OpenAI-compat endpoints | In deepagents 0.5.x, `init_chat_model(\"openai:...\")` defaults to the OpenAI **Responses API**. DeepSeek, OpenRouter, LiteLLM proxy, and vLLM speak chat-completions but not Responses → every call 404s. Plain `openai` without a custom base_url is unaffected. |
| 3 | Plumb `Settings.deep_agents_skills` + `deep_agents_memory` (`list[str]`) | Forward to deepagents' `SkillsMiddleware` (progressive AGENTS.md-style loading) and `MemoryMiddleware` (cross-thread recall). Forwarded only when populated. Compiled-graph cache key now includes both lists so changing them invalidates cleanly. |

## Files

- `pyproject.toml` — adds `langchain-openai>=1.2.0,<2.0.0` to `deep-agents` extra and the `all` / `dev` / `[dependency-groups].dev` mirrors.
- `src/pocketpaw/config.py` — adds `deep_agents_skills` and `deep_agents_memory` fields.
- `src/pocketpaw/agents/deep_agents.py` — `_build_model()` flags `is_openai_compat_endpoint` for `openrouter`/`openai_compatible`/`litellm` and sets `use_responses_api=False`; `_get_or_create_agent()` forwards `skills`/`memory` and includes them in the cache key.
- `tests/test_deep_agents_backend.py` — 9 new tests across two new classes (`TestDeepAgentsResponsesApiKwarg`, `TestDeepAgentsSkillsMemoryPlumbing`).
- `uv.lock` — sweeps up the deepagents 0.5.1→0.5.8 transitive bump from #1083 plus other resolver updates triggered by adding `langchain-openai`.

## What this PR does NOT do

[...]

PR #1083, opened 2026-05-09
chore(deps): bump deepagents to 0.5.8

DevRohit06:
## Summary

- Bumps `deepagents` from `>=0.1.0` to `>=0.5.8,<0.6.0` (latest stable on PyPI).
- Bumps `langchain-mcp-adapters` from `>=0.1.0` to `>=0.2.2,<0.3.0`.
- Applied across all four pin sites: `deep-agents` extra, `all` composite extra, `dev` optional-dep, and `[dependency-groups].dev`.

> **Note:** the branch name says `0.4.12` because the initial pin was based on stale PyPI metadata; the corrected pin (`0.5.8`) is in the second commit on this branch. Branch was not renamed to keep the PR thread intact.

## Why

The `>=0.1.0` floor was set when the backend first integrated deepagents and is now several minor versions stale. The 0.5.x signature surface is what we'll actually code against in follow-up work:

| 0.5.x parameter | What it unlocks |
|---|---|
| `cache=langgraph.cache.base.BaseCache` | LLM response cache — big cost/latency win with DeepSeek prompt-cache pricing |
| `response_format=ToolStrategy\|ProviderStrategy\|AutoStrategy` | Pydantic-validated structured output (e.g. pocket-spec JSON) |
| `middleware=Sequence[AgentMiddleware]` | `FilesystemMiddleware`, `MemoryMiddleware`, `SubAgentMiddleware` (verified exports) |
| `subagents=list[SubAgent\|CompiledSubAgent]` | Declarative sub-agent specs |
| `skills=list[str]`, `memory=list[str]` | Progressive context loading |
| `interrupt_on`, `checkpointer`, `store`, `backend` | HITL gates, durability, custom backends |

## Verified 0.5.1 top-level exports

`CompiledSubAgent`, `FilesystemMiddleware`, `MemoryMiddleware`, `SubAgent`, `SubAgentMiddleware`, `create_deep_agent`, `backends`, `graph`, `middleware`. Note: `SkillsMiddleware`, `SummarizationMiddleware`, and `ProviderProfile` are NOT at the package root in 0.5.x — those moved or were removed since the 0.4.x docs.

Existing implementation in `src/pocketpaw/agents/deep_agents.py` (`init_chat_model` + `astream` v2 + compiled-graph cache) stays API-compatible — this is a pure floor bump, no code changes.

## Test plan

- [ ] `uv sync --dev` resolves cleanly.
- [ ] `uv run pocketpaw --help` boots without import errors.
[...]

Issue #1081, opened 2026-05-08
test(stream_event): two thinking-event integration tests fail on event ordering


prakashUXtech:
Surfaced by the new pytest gate (#1066, fixed in #1074).

## Failing tests

- \`tests/test_stream_event.py::TestLoopThinkingIntegration::test_loop_thinking_publishes_system_event\` — \`AssertionError: assert 'thinking_done' in ['agents_md_loaded', 'agent_start', 'thinking', 'agent_end']\`
- \`tests/test_stream_event.py::TestLoopThinkingIntegration::test_loop_thinking_not_in_memory\` — \`assert 0 == 1\`

## Why it failed

The first test expects a \`thinking_done\` event in the published stream but only sees \`['agents_md_loaded', 'agent_start', 'thinking', 'agent_end']\`. Either the agent loop no longer publishes \`thinking_done\` (renamed/removed), or the test mock no longer triggers the path that fires it.

The second test asserts \`0 == 1\` — thinking content is not being persisted to memory when expected.

CI run: https://github.com/pocketpaw/pocketpaw/actions/runs/25541307753

## Likely cause

Recent agent-loop refactors around the extended-thinking surface may have changed event names or the memory-write predicate. The tests were never re-run because there was no CI gate.

## Acceptance

- Both tests pass against the current agent loop's event surface.
- If events were intentionally renamed, update the tests; if the loop regressed, fix the loop.

prakashUXtech:
Closed by #1088 (merged 2026-05-12). The two TestLoopThinkingIntegration tests now pin `agent_id` to skip the default-loop router-rebuild path that was crashing on the MagicMock router. The two corresponding `--deselect` lines were removed from `.github/workflows/tests.yaml` so the gate exercises these tests again.

PR #1085, opened 2026-05-09
feat(pocket-specialist): single-shot pocket creation + deepagents 0.5.8 + ripple validator

DevRohit06:
## Summary

Squashes the full `feat/pocket-specialist` line of work against `ee`. The early-history bullets (deepagents 0.5.8 bump, Responses-API regression fix, single-shot specialist) are still part of this PR — see the commit list for the originals. Recent work on top:

### Specialist architecture
- **Edit specialist** (`pocket_specialist__edit`) mirrors create. Accepts optional `pocket` + `target_node_ids` handoff so the parent can pre-resolve disambiguation.
- **langchain_react backend** — thin alternative to `deep_agents` (uses `langgraph.prebuilt.create_react_agent` directly, skips the middleware stack that pocket flow doesn't use).
- **Plan handoff fields** on `PocketSpecialistHints` (`purpose / layout / focal_widget / data_shape / key_interactions`). Parent agent decides shape; specialist follows.
- **No placeholder pockets on failure.** Dropped `_force_persist_fallback`. When the specialist can't ship a real pocket, `run_specialist` returns `{ok:false, action:"failed", pocket:null, error}`. Reverses the earlier "always ships" rule — empty canvases captioned "auto-created from a brief" were worse UX than a clean failure message.

### Mutation surface
- **Granular UI/state mutations**: 5 node ops (`add/replace/set_prop/move/remove`) + 4 state ops (`set/append/remove/patch`) replace whole-spec rewrites. New `spec_ops.py` + `state_ops.py` pure helpers. MCP tools + agent_context wrappers emit granular SSE events with only the changed subtree.

### DeepSeek thinking on direct API
- **`reasoning_content` round-trip**: direct DeepSeek API path via `openai_compatible` provider. `_patch_openai_message_serializer` monkey-patches `langchain_openai` to capture + echo `reasoning_content` per the [DeepSeek thinking-mode contract](https://api-docs.deepseek.com/guides/thinking_mode) — without it every tool-using turn 400s. Applied in both `DeepAgentsBackend` AND `LangchainReactBackend`.

### Quality gates
- **Manifest validation retry gate**: `persist_pocket` validates prop names against the live widget manifest BEFORE saving. On invented props (e.g. chart `series`/`xAxis`/`categoryKey` — all Recharts hallucinations) returns `{ok:false, redraft_required:true, warnings, message}` without persisting. Model fixes and retries up to `pocket_specialist_max_validation_retries` (default 3). After cap, persists anyway. Manifest warnings also surface to the agent via tool result through `_format_manifest_warnings_for_agent`.
- **Chart prompt hardening** in `_design.py` — explicit ban list (`series/xAxis/dataKey/categoryKey/legend/axes/margin/stack`) + concrete per-type `{label, value}` examples for bar/line/donut/pie, plus multi-series via `series` field on each data point.

### Chat UX
- **Atomic auto-open + session bind**: `persist_pocket` pushes `pocket_created` SSE + calls `attach_pocket_to_session_doc` directly after `_agent_create` succeeds (shares the parent stream's contextvars). No longer depends on `_maybe_handle_specialist_response` parsing the parent's `tool_result` event.
- **Prompt restructure**: behavioral rules (`INLINE_RIPPLE_SYSTEM_PROMPT` + `POCKET_DELEGATION_RULE`) hoisted out of the `## Your Knowledge Base` wrapper into a new `instructions` channel on `pool.run`. New `build_behavior_instructions()` + `build_dynamic_context()` split the static rules from per-turn reference data. Strengthened the delegation rule with a hard "talk before you call the tool" preface so the parent emits text before kicking off the multi-second specialist call.
- **Real-time side-channel streaming**: `agent_router` races the next agent event against `side_channel_queue.get()` so `push_sse_event` calls from inside in-process tools (specialist status pushes during its multi-second run) flush to the SSE consumer in real time instead of all at once after the tool returns.
- **Sub-stage tool_start events**: specialist pushes synthetic `tool_start` events (`pocket_specialist:build`, `pocket_specialist:save`) so the desktop client's `TOOL_LABELS` lookup updates the loader label as work progresses, instead of leaving "Designing pocket…" frozen.

## Why one PR

The deepagents bump, Responses-API fix, specialist subagent, single-shot refactor, reasoning_content patch, mutation surface, retry gate, and prompt restructure are tightly coupled. The specialist runtime depends on the deepagents API surface; the granular mutations depend on the new MCP tools; the retry gate depends on manifest warnings reaching the agent; auto-open depends on the contextvar/SSE plumbing the specialist also uses. Reviewing piecewise would be strictly worse.

Closes #1069, #1083, #1084.

## Test plan

- [ ] `uv run pytest --ignore=tests/e2e` passes. The known pre-existing failures on `ee` are tracked separately (#1079 concurrency, #1080 context_budget, #1081 stream_event); this PR review surfaced four blockers in the new code (specialist authz on `_agent_load_doc`, `loop.py` rebuild `stop()`, CLI test boot setup, prompt drift) — all fixed in the latest commits.
- [ ] Specialist creates a pocket end-to-end via `pocket_specialist__create` (MCP path) in under 60s with `POCKETPAW_POCKET_SPECIALIST_MODEL=openai_compatible:deepseek-v4-flash` and `POCKETPAW_DEEP_AGENTS_DISABLE_THINKING=false`.
- [ ] No `reasoning_content` 400s on multi-turn specialist runs.
- [ ] Edit flow (`pocket_specialist__edit`) applies granular ops to an existing pocket; canvas updates in place via SSE.
- [ ] Auto-open: new pocket appears in the desktop client immediately after specialist returns; session.pocket is set in Mongo.
- [ ] Existing pockets continue to load and modify; ripple validator catches invalid specs at write time.
- [ ] `deep_agents` / `langchain_react` backends still work against Anthropic / OpenAI / Ollama providers (no regression).


prakashUXtech (changes requested):
## Review summary

Architecture is solid. There's one blocking authorization gap on the agent mutation path plus a handful of PR-introduced test regressions to land before merge. The PR body also overstates how clean CI is and needs a small rewrite.

**Status:** request changes
**Rebase needed:** yes (CONFLICTING vs `ee`, 76 commits since merge-base)

---

## Blockers

### 1. Missing workspace/owner authz on the agent mutation path

`_agent_load_doc` in `ee/cloud/pockets/service.py:692` loads a pocket by `ObjectId` and returns it without checking workspace or owner. All 9 new granular mutation tools (5 node ops + 4 state ops) in `src/pocketpaw/agents/sdk_mcp_pocket.py:180-285` route through this loader, and `pocket_specialist__create / __edit` in `ee/agent/pocket_specialist/mcp_tool.py` only verify that the workspace/user ContextVars are present (non-None), never that they match the pocket.

The REST update path at `service.py:338` does the right thing via `_check_domain_edit_access`. The agent mutation path at `service.py:833-866` skips it. An agent with a valid session in workspace A can call `set_state(pocket_id=<pocket_in_workspace_B>)` and mutate it if it knows or guesses the ObjectId.

Fix: add `doc.workspace == workspace_id` plus an owner / shared-with check inside `_agent_load_doc` (or a new `_agent_load_doc_authed` helper). Both values are already in ContextVars.

### 2. `loop.py` regression breaks 5 tests

`src/pocketpaw/agents/loop.py:449` does `asyncio.create_task(old.stop())`. Tests that pass a `MagicMock` (rather than `AsyncMock`) fail with `TypeError: a coroutine was expected, got <MagicMock name='mock.stop()'>`. Affected:

- `tests/test_concurrency.py::test_session_lock_serialises_same_session`
- `tests/test_concurrency.py::test_cross_session_runs_in_parallel`
- `tests/test_concurrency.py::test_global_semaphore_caps_concurrency`
- `tests/test_stream_event.py::TestLoopThinkingIntegration::test_loop_thinking_publishes_system_event`
- `tests/test_stream_event.py::TestLoopThinkingIntegration::test_loop_thinking_not_in_memory`

Fix: switch the test fixtures to `AsyncMock`, or guard the call site with `inspect.iscoroutine(maybe_coro)` before scheduling.

### 3. `_run_cloud_handler` test failures

`src/pocketpaw/tools/cli.py:282` calls `_ensure_cloud_runtime_initialized` unconditionally before dispatching the handler. The two new tests added in this PR pass a bare async function that doesn't need Mongo, but the init guard runs first and returns `{ok: False, error: "POCKETPAW_MONGO_URI ... not set"}`:

- `tests/test_tools_cli_cloud.py::test_run_cloud_handler_serializes_to_json_line`
- `tests/test_tools_cli_cloud.py::test_run_cloud_handler_catches_exceptions`

Fix: patch `pocketpaw.tools.cli._ensure_cloud_runtime_initialized` in the test setup.

### 4. PR body misrepresents CI state

Two factual issues in the body that should match the code before merge:

- "6 pre-existing failures: `test_runtime` patches a renamed helper + `test_settings` env-leak from `.env` defaults". `ee` base actually has 2 collection errors (`test_fleet_router`, `test_pocket_journal_stream`), not 6. CI's `--deselect` list already covers the 6 real pre-existing failures (concurrency #1079, context_budget #1080, stream_event #1081). The actual red on this PR is from new failures introduced here, not pre-existing ones in `test_runtime` / `test_settings`.
- "MAX_VALIDATION_RETRIES (6, was 3)". `src/pocketpaw/config.py:338` defaults `pocket_specialist_max_validation_retries` to **3**. The code matches "was 3", but the bump to 6 didn't happen.

---

## High-priority (please land in this PR if quick)

### 5. `run_edit_specialist` silent failure

`ee/agent/pocket_specialist/runtime.py:566` returns `ok=True` for every edit run regardless of whether any ops applied or the backend errored mid-stream. An empty run (model refuses, network drops, exception swallowed in the `finally` block of `backend.run`) returns `PocketSpecialistEditOutput(ok=True, ops=[])` and the caller has no way to tell "no work needed" from "the specialist crashed."

Suggest a `success` flag that starts `False` and flips `True` only after the loop completes without exception. Mirrors how `run_specialist` uses `persist_capture.get("pocket")` as its success signal.

### 6. Monkey-patch needs `hasattr` guard

`src/pocketpaw/agents/deep_agents.py:71-107` reaches into private `langchain_openai` symbols (`_convert_dict_to_message`, `_convert_delta_to_message_chunk`, `_convert_message_to_dict`). The `try/except ImportError` guard at line 65 catches a missing module but not a missing attribute. If a future langchain-openai 1.x bump renames or moves one of these, the patch raises `AttributeError` on the first DeepSeek call with no early warning.

Wrap each of the three assignments in `if hasattr(_oa, "_convert_dict_to_message"):` etc., and `logger.error` when the symbol is absent so a langchain upgrade surfaces loudly in CI instead of crashing in production.

### 7. Three new cloud test files don't collect without `--all-extras`

`tests/cloud/test_spec_ops.py`, `tests/cloud/test_state_ops.py`, `tests/cloud/test_pocket_granular_ops.py`, all added by this PR, import via `ee.cloud.pockets`, which pulls `router` → `service` → `from beanie import PydanticObjectId`. CI sees them because of `uv sync --dev --all-extras`, but any local run without the cloud extras hits `ModuleNotFoundError: beanie` and the tests vanish from the suite silently.

Either direct-import `ee.cloud.pockets.spec_ops` / `state_ops` without crossing `__init__.py`, or mark them with a `pytest.mark.cloud` gate.

---

## Non-blocking, track as follow-ups

- No `AuditEvent` on granular mutation tools or the force-persist-after-retries path. Operators watching audit logs see neither the 9 new agent mutation surfaces nor the case where an invalid spec lands in Mongo after cap.
- `attach_pocket_to_session_doc` is read-then-write with no optimistic lock. Narrow race when two concurrent specialist calls touch the same session_doc.
- `side_channel_queue` at `ee/cloud/chat/agent_router.py:497` is unbounded. Degenerate edit-specialist runs could grow it without limit. A `maxsize` cap would prevent that.
- `replace_node` in `ee/cloud/pockets/spec_ops.py:185-186` silently discards a replacement's id when the target had one. Downstream SSE events referencing the old id will target a ghost node.
- `ee/agent/pocket_specialist/tools.py:19` has a top-level `from langchain_core.tools import StructuredTool`, which violates the lazy-import convention for optional dependencies.
- New MCP tools register via `@tool` rather than the PocketPaw `ToolProtocol`. `tools_deny` / `tool_profile=minimal` policy filtering won't suppress them.
- Granular mutations do full-document Mongo saves rather than field-level `$set`. Last-write-wins under concurrent agent ops.

---

## What's right

Worth calling out, since the diff is big:

- `_force_persist_fallback` is cleanly and completely removed. The `{ok: false, action: "failed", pocket: null}` path in `run_specialist` is wired correctly and no dead code remains in live modules.
- `spec_ops.py` / `state_ops.py` are textbook pure-function design. Side-effect free, comprehensive error raises, clear `__all__`, solid docstrings. The `move_node` cycle guard at `spec_ops.py:307` is correct and tested.
- The side-channel race in `agent_router.py:555-584` is correctly structured. No leaked tasks, proper drain in the `finally` block, cancellation propagation works as intended.
- Multi-tenancy enforcement in the tool factories holds up: `workspace_id` and `user_id` are closed over, never exposed as LLM-visible arguments. The LLM can't cross workspaces via hallucinated argument names.
- `LangchainReactBackend` subclassing `DeepAgentsBackend` is the right move. Reuses model build, streaming shape, stop semantics, and tool bridging without duplicating any of it.
- Test coverage on the monkey-patch and the new pocket_specialist subsystem is real: 5 reasoning_content tests, 8 streaming-event tests, 105 tests under `tests/ee/agent/test_pocket_specialist/` once collection succeeds.

---

## Suggested order

Blockers 1-4 first. 5-7 in the same PR if scope allows. The non-blocking items can become follow-up issues. Rebase against current `ee` happens at the same time.

prakashUXtech (approved):
All 7 review items addressed across 6 fix commits. CI green (pytest 3m19s, check-quality, security-scan). The two cleanup test fixes I pushed (test_credentials env isolation + tool-count contract update for the function-tool vs shell-CLI specialist split) clear the last failures. Convention-tightening items folded into a follow-up issue. Merging.

PR #1091, opened 2026-05-12
fix(callback): validate OAuth2 state to prevent CSRF Revised

### LINKED ISSUE #999: security: OAuth state param CSRF gap on /oauth/integrations/authorize

safwan027:
# MCP OAuth State Flow 

This document explains the OAuth `state` hardening used in `src/pocketpaw/mcp/manager.py`.

## Related Issue

Fixes #999, superseeds #1041 PR

## Why this exists

PocketPaw now wraps OAuth callback state in a signed token to prevent:

- forged callback state values,
- cross-flow mixups,
- replay of previously valid callback states.

The implementation preserves SDK compatibility by carrying the original upstream state inside the signed payload.

## Terms

- **Upstream state**: the original `state` generated by the OAuth SDK/client flow.
- **Signed state token**: PocketPaw-generated state sent to provider/browser roundtrip.
- **Pending flow**: server-side one-time record keyed by internal `sid`.

## Top-to-bottom flow

1. OAuth SDK creates auth URL and calls `redirect_handler(auth_url)`.
2. PocketPaw extracts the original upstream `state` from query params.
3. PocketPaw creates a signed token payload containing:
   - `sid` (internal state id)
   - `uid` (user binding)
   - `prv` (provider/server binding)
   - `iat` and `exp` (issue/expiry)
   - `n` (random nonce)
   - `ups` (original upstream state)
4. PocketPaw stores `_PendingOAuthFlow` in `_oauth_pending[sid]`.
5. PocketPaw replaces outbound URL `state` with the signed token and broadcasts URL to UI.
6. Provider redirects callback with `code` + signed `state`.
7. Callback endpoint calls `set_oauth_callback_result(state, code)`.
8. PocketPaw verifies signature + expiry, loads pending flow by `sid`, validates bindings (`uid`, `prv`), and resolves future with `(code, upstream_state)`.
9. Pending state is removed (`pop`) to enforce one-time use.
10. `callback_handler()` resumes and returns `(code, returned_state)` to SDK, which continues token exchange and token storage.

## Why we keep upstream state

We intentionally preserve upstream state (`ups`) because SDK internals may expect the original state value for correlation.

So we use:

- **signed state** for browser/provider security boundary,
- **upstream state** for SDK compatibility after validation.

## Security checks performed

- HMAC signature verification for state token integrity.
- Expiration checks (`exp` and pending TTL checks).
- Binding checks against pending flow metadata.
- One-time consume semantics (`_oauth_pending.pop(...)`).

## Recent Changes: Real Session Identity & Tests

The codebase has been updated to replace the hardcoded `"local_user"` identity with the real authenticated session identity.

### Codebase Analysis

**1. ContextVar for Session Identity (`manager.py`)**
- A `ContextVar` named `_oauth_subject_id` is now used to store the current OAuth subject binding for the active request/task context.
- Functions `set_oauth_subject_id` and `reset_oauth_subject_id` manage this token context.
- During OAuth initialization (`redirect_handler`), `user_id = _oauth_subject_id.get()` is retrieved dynamically instead of hardcoding `"local_user"`.

**2. Test Coverage (`test_mcp_oauth.py`)**
- Tests previously using `"local_user"` have been updated to use generic user identities (e.g., `"test_user"`).
- Tests properly cover scenarios including binding mismatch rejections (`test_set_oauth_callback_result_rejects_uid_binding_mismatch`), validating that the binding checks enforce strict user identity mapping.


safwan027:
Is there update on this ?

PR #1102, opened 2026-05-13
Fix OAuth token lifecycle: auto-refresh, accurate status, recovery on restart

### LINKED ISSUE #1101: OAuth connectors show stale "Connected" state, no recovery on restart, dead tokens poll forever
## Summary

The OAuth token lifecycle for connectors has several broken flows that cause the UI to show incorrect connection state and fail to recover after restarts.

Example: A user connects Gmail via OAuth, uses it for a few hours, then comes back the next day. The access token expired overnight but the UI still shows green "Connected". The "Reauthorise" button never appears because `cred_state` is cached statically at connect-time and never dynamically checked. If the user refreshes the page or the server restarts, all connectors reset to grey "Not connected" even though valid tokens exist on disk — the in-memory adapter instances are lost with no recovery mechanism.

---

## Repro

1. Sign into `paw-enterprise` as any user
2. Go to `Pocket Data Panel → Data Sources → Connect Gmail` (or any OAuth connector)
3. Wait for access token to expire (~1 hour) or revoke the app in Google account settings
4. The `ConnectorCard` still shows green `"Connected"` — no `"Reauthorise"` button appears
5. Refresh the page → all connectors show grey `"Not connected"` despite tokens on disk
6. Open browser DevTools → `/api/v1/connectors/gmail/status` returns:

```json
{
  "connected": false,
  "cred_state": "missing"
}
```

7. Check server logs → `GET /api/v1/tools` triggers `"Token refresh failed"` warnings for every dead token on every poll, forever

---

## Suggested Fix

Multiple changes across backend and frontend:

#### `api/v1/connectors.py`

- Add name aliasing (`google_drive → drive`) so `404` doesn't block `ConnectorCard`
- Add lazy-connect that detects tokens on disk and auto-creates the adapter (handles server restart)
- Call `adapter.health()` for real-time `cred_state` instead of caching it at connect-time
- Delete OAuth tokens on disconnect so it stays disconnected

#### `api/v1/oauth_integrations.py`

- Add bidirectional service ↔ connector name mapping so:
  - `?service=drive`
  - `?service=google_drive`

  both work

#### `api/v1/tools.py`

- Use `OAuthManager.get_valid_token()` with auto-refresh instead of just checking `tokens.access_token` existence
- Return `"connected"` only when a valid token is actually obtainable

#### `clients/oauth.py`

- Delete dead tokens on `4xx` refresh errors so subsequent polls don't keep retrying and logging warnings

---

#### `ConnectorCard.svelte`

- Check `cred_state === "expired"` before `connected` in `deriveLocalState()`
- Expired tokens should show `"Reauthorise"` amber badge instead of green `"Connected"`

#### `PocketDataPanel.svelte`

- Remove duplicate `/api/v1` from `startOAuth()` URL
- `BASE_URL` already contains it, currently producing:

```txt
localhost:8888/api/v1/api/v1/oauth/... → 404
```

AMRITESH240304:
## What does this PR do?

Fixes the OAuth token lifecycle for connectors — expired tokens now properly show "Reauthorise" instead of stale green "Connected", connectors
  recover their state after server restart via lazy-connect from disk tokens, and dead tokens are cleaned up instead of logging warnings on every
  poll. Also fixes a duplicate /api/v1 in the Tools OAuth URL that caused 404s.

## Related Issue

Fixes #1101 

## Changes Made

  - `src/pocketpaw/api/v1/connectors.py`: Added name aliasing (google_drive→drive) so ConnectorCard 404s are fixed. Added lazy-connect that
  auto-creates adapters from disk tokens on restart. Status endpoint now calls adapter.health() for real-time cred_state. Disconnect deletes OAuth
  tokens so it stays disconnected. Execute records cred_state="valid" on success.
  - `src/pocketpaw/api/v1/oauth_integrations.py`: Added SERVICE_TO_CONNECTOR/CONNECTOR_TO_SERVICE mappings so authorize accepts both ?service=drive
  and ?service=google_drive.
  - `src/pocketpaw/api/v1/tools.py`: list_tools() now uses OAuthManager.get_valid_token() with auto-refresh instead of just checking file existence.
  Returns "connected" only when a valid token is obtainable.
  - `src/pocketpaw/clients/oauth.py`: refresh_token() deletes permanently dead tokens on 4xx errors so subsequent polls stop retrying.
  - `tests/v1/test_api_v1_connector_status.py`: Updated tests for name aliasing and added async mock for health checks.

## Checklist

- [x] PR targets `ee` branch (not `main`)


prakashUXtech:
Heads up — this PR targets `ee`, which is now deprecated. PR #1157 synced `ee` into `dev` on 2026-05-21, so `dev` is the single canonical branch from here on.

This PR needs to be retargeted before it can merge — otherwise its base branch no longer exists as a merge target:

```
gh pr edit 1102 --base dev
```

Then rebase your branch onto `dev` and resolve any conflicts. `dev` already carries everything `ee` had, so the rebase is mostly your own commits replaying on top.

PR #1106, opened 2026-05-14
feat(pockets): Tier-2 array-item ops + slim edit-specialist prompts

DevRohit06:
Follow-up to #1085. Adds granular (Tier-2) array-item operations on pocket widget props, slims the ripple/pocket-specialist prompts, and tightens the one-shot edit handoff.

## Changes

**Tier-2 array-item ops** (`ee/cloud/pockets/`)
- `prop_arrays.py` — allowlist of editable prop-array paths
- `spec_ops.py` — `match_array_item` helpers (id / index / predicate)
- `service.py` — `agent_set_prop_array_item`, `agent_append_prop_array_item`, `agent_remove_prop_array_item` + `*_for_agent` wrappers
- `agent_context.py` — wires the new ops into agent-side context

**Pocket specialist** (`ee/agent/pocket_specialist/`)
- `tools.py` — Tier-2 LangChain tool factories surface the new service handlers
- `runtime.py` / `mcp_tool.py` — refreshed wiring; one-shot edit handoff

**Ripple prompts** (`ee/ripple/`)
- `_design.py`, `_inline_core.py`, `_pockets.py` — slim prompts; document Tier-2 ops; drop dead branches

**Tests**
- New: `test_prop_arrays.py`, `test_spec_ops.py`, `test_pocket_granular_ops.py`
- Refreshed: edit / edit-handoff / widget-diversity suites; prompt-cache and prompt-single-source assertions

## Test plan

- [x] `uv run pytest tests/cloud/test_prop_arrays.py tests/cloud/test_spec_ops.py tests/cloud/test_pocket_granular_ops.py`
- [x] `uv run pytest tests/ee/agent/test_pocket_specialist/`
- [x] `uv run ruff check .`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ Thumbs.db
 htmlcov/
 .tox/
 .nox/
+.tmp/
 
 # Internal drafts
 idocs/
--- a/ee/agent/pocket_specialist/mcp_tool.py
+++ b/ee/agent/pocket_specialist/mcp_tool.py
@@ -253,12 +253,14 @@ def build_pocket_specialist_server() -> Any:
     @tool(
         "edit",
         (
-            "Edit an existing pocket from a natural-language intent. The "
-            "specialist reads the current pocket, picks the smallest set "
-            "of granular ops (set_state for data, set_node_prop for "
-            "widget appearance, add/move/remove_node for structure), and "
-            "applies them. Each op persists and pushes its own SSE event "
-            "so the canvas updates in place. Returns "
+            "Edit an existing pocket in ONE LLM round-trip. Call "
+            "get_pocket first, then pass the result here as `pocket` "
+            "(required). The specialist has no read tool — without "
+            "`pocket` the edit fails. It picks the smallest set of "
+            "granular ops (set_state for data, set_node_prop for "
+            "widget appearance, add/move/remove_node for structure) "
+            "and applies them. Each op persists and pushes its own "
+            "SSE event so the canvas updates in place. Returns "
             "{ok, pocket_id, ops, duration_ms}."
         ),
         {
@@ -282,27 +284,25 @@ def build_pocket_specialist_server() -> Any:
                 "pocket": {
                     "type": "object",
                     "description": (
-                        "OPTIONAL handoff. The current pocket view "
-                        "(rippleSpec + metadata) you already fetched. "
-                        "When passed, the specialist skips its own "
-                        "get_pocket call. Pass this when you read the "
-                        "pocket to disambiguate or confirm the edit."
+                        "REQUIRED. The current pocket view (rippleSpec "
+                        "+ metadata) you fetched with get_pocket. The "
+                        "specialist reads this directly and has no way "
+                        "to fetch it itself — pass it on EVERY edit "
+                        "call, no exceptions."
                     ),
                 },
                 "target_node_ids": {
                     "type": "array",
                     "items": {"type": "string"},
                     "description": (
-                        "OPTIONAL handoff. Node ids you identified as "
-                        "edit targets after reading the pocket. When "
-                        "set, the specialist works ONLY on these nodes "
-                        "and does not search. Best practice for any "
-                        "edit that needs disambiguation (the user said "
-                        "'the chart' and there are three)."
+                        "OPTIONAL but encouraged. Node ids you "
+                        "identified as edit targets after reading the "
+                        "pocket. When set, the specialist works ONLY "
+                        "on these nodes and does not search."
                     ),
                 },
             },
-            "required": ["pocket_id", "intent"],
+            "required": ["pocket_id", "intent", "pocket"],
             "additionalProperties": False,
         },
     )
--- a/ee/agent/pocket_specialist/runtime.py
+++ b/ee/agent/pocket_specialist/runtime.py
@@ -354,10 +354,9 @@ def _build_user_message(input: PocketSpecialistCreateInput) -> str:
 def _build_edit_user_message(input: PocketSpecialistEditInput) -> str:
     """Build the edit specialist's first user message.
 
-    When the parent already read the pocket and/or identified the
-    target nodes, surface that in the body of the message. The
-    specialist's system prompt has matching rules for skipping its
-    own ``get_pocket`` call and working only on the targeted nodes.
+    The parent agent is contractually required to send the current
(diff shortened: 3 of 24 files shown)


DevRohit06:
## Code Review

**Status:** CI green (pytest / quality / security-scan all PASS) · 2101 additions / 988 deletions · 23 files

### Overview

Adds three Tier-2 array-element ops (`set_prop_array_item`, `append_prop_array_item`, `remove_prop_array_item`) so the LLM can surgically edit one chart row / table column / kanban column without re-shipping the whole array. Layered cleanly: `prop_arrays.py` (allowlist) → `spec_ops.py` (matching) → `service.py` (mutations) → `agent_context.py` (agent wrappers) → `tools.py` (LangChain tool factories). Also slims the ripple/pocket-specialist prompts and tightens the edit-specialist into a strict one-shot handoff.

---

### Strengths

- **Closed allowlist** in `prop_arrays.py` prevents silent prop mangling on typos. Co-located with design doc + prompt cheatsheet, with a comment telling future contributors to keep all three in sync.
- **Fail-fast contract** in `run_edit_specialist` (runtime.py:132–150): missing `pocket` returns a deterministic error before any LLM round-trip. Saves cost and surfaces parent-side bugs immediately.
- **Ambiguity is explicit** — `match_array_item_candidates` returns the full candidate list so the service can render `"ambiguous: N items matched; candidates=[...]"`. Much better than first-match silent wins.
- **Append asymmetry is intentional and documented** (service.py:674): set/remove bail on missing array (nothing to address), append creates the array (its whole job).
- **Backward-compat on prompt restructure**: `CANONICAL_SHAPES = preamble + "\n".join(WIDGET_SHAPES.values())` keeps existing callers working while new code can fetch one widget's shape (~2k chars) instead of the full ~10k blob.
- **Strong test coverage** — 3 new files totalling 550 lines, plus refreshed edit/handoff/diversity suites. CI green.

---

### Concerns

#### 1. Breaking change in MCP tool schema (`mcp_tool.py:72`) — Medium
`pocket` moved from optional to required:
```python
"required": ["pocket_id", "intent", "pocket"],
```
Combined with the runtime fail-fast, ANY caller that hadn't pre-fetched will now hard-error instead of degrading. Schema description hammers "no exceptions" — but if `pocket_specialist__edit` is exposed externally, downstream callers get a breaking change without a major-version bump. **Action:** confirm all internal + downstream callers prefetch, or add a version flag / deprecation window.

#### 2. `partial` parameter name is misleading for non-dict items (`service.py:617–619`) — Low–Medium
```python
if isinstance(arr[idx], dict):
    arr[idx] = {**arr[idx], **partial}
else:
    arr[idx] = partial   # non-dict element: replace wholesale
```
Arg is documented as a shallow-merge "partial", but for non-dict items it's wholesale replacement. The LLM reading the tool description sees "fields to merge" and won't expect replacement semantics. **Suggestion:** either reject non-dict items explicitly with a clear error, or rename to `update` and document both branches.

#### 3. Shallow merge will silently clobber nested data — Low–Medium
Documented, but worth a worked example in the edit-specialist cheatsheet: setting `{style: {color: "red"}}` on `{label:"X", style:{color:"blue", weight:"bold"}}` drops `weight`. The LLM will get bitten by this on chart series objects with nested config. **Suggestion:** add a "watch out: shallow merge" example in `_pockets.py`, or expose a deep-merge variant for known-nested props.

#### 4. Lock held across `emit` (`service.py:627`, repeated in append/remove) — Low
```python
async with _pocket_lock(pocket_id):
    ...
    await doc.save()
    await emit(PocketUpdated(...))
```
If `emit` awaits slow SSE subscribers, the per-pocket lock is held longer than necessary and subsequent ops queue up. **Suggestion:** verify `emit` is fire-and-forget; if not, hoist it out of the lock.

#### 5. `deepcopy(old_item)` on every set call (`service.py:615`) — Low
Acceptable for small rows but expensive if a row contains embedded richtext or large nested payloads. Only used for the return value. **Suggestion:** opt-in via `return_old=False` default, or document the size guardrail.

#### 6. `_item_matches` silently skips non-dict items (`spec_ops.py:828`) — Low
A malformed pocket with a stray primitive in an allowed array will silently skip past matches instead of erroring. **Suggestion:** debug log or raise on first non-dict so corruption surfaces early.

#### 7. Prompt slim dropped `<parent-handoff>` from `_pockets.py` — Low
Deletion is intentional (merge commit notes it), runtime still injects `TARGET NODE IDS` / `CURRENT POCKET` in the user message — but the inner specialist no longer has prompt-level guidance on how to use them. The bound-vs-literal decision added in commit 3 is supposed to compensate. **Action:** verify the test suite has a case where the specialist receives `target_node_ids` and must work only on them without explicit prompt instructions.

#### 8. Test gaps I'd want to see filled
`test_pocket_granular_ops.py` is solid, but I'd want explicit coverage for:
- `prop` in allowlist but node has no `props` dict at all (set should error; append should create — the documented asymmetry).
- `by_field` matching where the candidate field value is literally `None` (`item.get(field) == None` matches a missing key too — possible false-positive).
- Concurrent two-writer scenario on the same pocket (does `_pocket_lock` actually serialize?).
- Index out-of-range error message shape (callers may parse `candidates=[...]`).

---

### Risk summary

| Risk | Severity | Mitigation |
|---|---|---|
| Breaking external callers of `edit` MCP tool | **Medium** | Version flag or grace period |
| Shallow-merge gotcha on nested chart config | Low–Medium | Cheatsheet warning + worked example |
| `partial` semantics differ for non-dict items | Low | Tighten error or rename arg |
| Lock held across emit | Low | Verify emit is async-non-blocking |
| Prompt slim regressed handoff guidance | Low | Add test or restore minimal handoff note |

---

### Verdict

Substantive, well-layered work with strong tests. The architectural shape (allowlist + matcher + service + agent wrappers + tool factories) is right and will scale to future widget types.

Primary blockers before merge: **resolve the breaking-change angle on the `edit` MCP tool** (item 1) and **clarify `partial` semantics for non-dict items** (item 2). Other items are follow-up-worthy but don't have to land in this PR.

The `needs-work` label is likely tracking exactly these kinds of items — recommend pinging existing reviewer comments to confirm scope before merging.

prakashUXtech:
Closing — superseded. The work here was reworked onto the current `dev` layout (post OSS-EE split) and merged as #1159 (`feat(pockets): Tier-2 array-item edit ops + slim design prompts`). This PR also targeted the now-deprecated `ee` branch.

Issue #192, opened 2026-02-17
Replace pipe-to-execution pattern in Windows uv installation with safer download-execute flow


DS-Kushagra:
The Windows installer currently uses a pipe-to-execution pattern:

irm <url> | iex

This approach has risks:
- Partial execution if download fails
- Security concerns in restricted environments
- Harder debugging of installer failures

### Proposed Improvement
Switch to:
1. Download script via Invoke-RestMethod -OutFile
2. Execute locally after verification

This ensures safer and more reliable installation.

DevRohit06:
Fixed by PR #194

PR #1114, opened 2026-05-16
feat(cloud): Projects entity + snapshot scheduler for Mission Control

prakashUXtech:
## Summary

Backend work to round out Mission Control on the enterprise cloud:

- **Projects entity** under `ee/cloud/projects/`. A Workspace can contain Projects; a Project optionally groups Pockets, Tasks, and Cycles. Every child reference is optional (default `None`), so existing rows keep working as "unassigned" with no migration. Soft-archive is idempotent. Hard-delete soft-unassigns children rather than cascading, so the underlying work survives even if the container goes away.
- **In-process daily-snapshot scheduler** for the cycles burnup chart, opt-in via `POCKETPAW_CLOUD_SCHEDULER_ENABLED=true`. Off by default so pytest runs don't spawn a background loop. The loop sleeps until the next UTC midnight, then calls `snapshot_all_active()` for every workspace that has at least one active cycle. There's also a new `POST /cycles/{id}/snapshot` endpoint that forces today's snapshot for tests and intern onboarding (idempotent within a UTC day).
- **Mission Control (Cloud) docs page** at `/advanced/mission-control-cloud`, separate from the existing local Mission Control page. Opens with a callout flagging the distinction so search traffic doesn't confuse the two surfaces.

## Test plan

- [x] 19 new tests in `test_projects_service.py` and `test_projects_router.py` covering CRUD, tenant isolation, and cascade unassign on delete.
- [x] 5 new tests in `test_cycles_snapshot_scheduler.py` covering manual trigger, same-day idempotence, scheduler start/stop wiring, and the workspace discovery helper.
- [x] 3 new tests in `test_mission_control_project_filter.py` confirming `project_id` narrows the feed via the visible-pocket set.
- [x] Adjacent suites all green: 78 tests across cycles, mission_control, tasks, projects.
- [x] `uv run lint-imports`: 13 contracts kept, 0 broken. (New Projects contract added; nothing else changed.)
- [x] `uv run ruff check` and `ruff format` both clean on the touched files.
- [x] Pre-existing failures in unrelated modules (uploads, connectors, decision_traces, ee_correction) confirmed present on `ee` without my changes.

## Notes

- Base branch is `ee`, not `main`.
- Scheduler is opt-in. Multi-instance deployments that prefer external cron, Kubernetes CronJob, or Celery beat leave the flag unset and dispatch the same `snapshot_all_active()` callable from their platform scheduler — same idempotency, no double-fire.
- Touch-time migration rule satisfied: every entity I edited (Pocket, Task, Cycle) was already on the 4-file shape. The new Projects entity matches the canonical pockets layout.
- Cascade on project delete is soft-unassign across all three child entities. Children keep their data and emit their own `*.updated` events per row, so downstream listeners (search index, Pawprints projection) stay consistent.

### DIFF
--- a/docs/advanced/mission-control-cloud.mdx
+++ b/docs/advanced/mission-control-cloud.mdx
@@ -1,0 +1,167 @@
+---
+title: "Mission Control (Cloud): operator console for the SaaS surface"
+description: "Mission Control (Cloud) is the operator console for the PocketPaw enterprise cloud — Tray, Pawprints, Snags, Projects, Cycles. Workspace-scoped REST API + SSE event stream, backed by MongoDB."
+section: Advanced
+ogType: article
+keywords: ["mission control", "cloud", "saas", "operator console", "tray", "pawprints", "cycles", "projects"]
+tags: ["advanced", "cloud", "saas", "operator"]
+---
+
+# Mission Control (Cloud)
+
+<Callout type="info">
+**Different surface from the local Mission Control.** The page at [/advanced/mission-control](/advanced/mission-control) describes the *local multi-agent orchestration framework* — file-based JSON storage at `~/.pocketpaw/mission_control/`, single-process, used by Deep Work and CLI tooling.
+
+This page describes the **cloud SaaS surface**: workspace-scoped REST API + MongoDB-backed entities, served by the enterprise dashboard at `pocketpaw/ee/cloud/`. The two surfaces share a name (and a mental model) but no code paths.
+</Callout>
+
+Mission Control (Cloud) is the operator console for the PocketPaw enterprise cloud. It projects every kind of work the operator can see — Nudges waiting for approval, Tasks in flight, Cycles in progress, Pawprints already shipped — into a unified work-item feed, scoped to the active workspace.
+
+## Vocabulary
+
+| Term | Meaning |
+|------|---------|
+| **Tray** | Items awaiting human approval (Nudges from Instinct). |
+| **Pawprints** | Items already approved, rejected, or executed (audit projection). |
+| **Snags** | Items that failed or are blocked — recoverable. |
+| **Agents** | Tasks currently in flight, claimed by an agent runtime. |
+| **Cycle** | A time-boxed work window — Linear-style. Carries a burnup chart fed by the daily snapshot job. |
+| **Project** | A workspace-scoped grouping of pockets, tasks, and cycles. Optional reference everywhere. |
+| **Pocket** | The canvas / workspace surface. Pockets live inside a workspace and optionally inside a Project. |
+| **Nudge** | A pending action proposed by Instinct, surfaced in the Tray for human approval. |
+
+## Hierarchy
+
+```
+Workspace
+  └── Project (optional)
+        └── Pocket
+              └── Cycle (optional, pocket-scoped)
+              └── Task (optional, can attach to Project, Pocket, Cycle, or none)
+```
+
+Every level except the workspace is optional. A task created without a `project_id`, `pocket_id`, or `cycle_id` lands in the workspace-wide "unassigned" bucket and still shows up in Mission Control. The frontend rolls items up into project / cycle / pocket buckets as the operator drills down.
+
+## WorkItem shape
+
+`GET /api/v1/mission-control/items` returns a list of `WorkItem` records — the unified projection over Instinct nudges + the Tasks entity. Every entry carries:
+
+| Field | Type | Notes |
+|-------|------|-------|
+| `id` | string | Prefixed: `nudge:<action_id>` or `task:<task_id>`. |
+| `workspace_id` | string | Always the caller's active workspace. |
+| `section` | enum | `tray` / `pawprints` / `snags` / `agents`. |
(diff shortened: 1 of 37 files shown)


prakashUXtech:
## Review summary

**Verdict:** REQUEST CHANGES — one blocking, one high-priority. Architecture is solid; fixes are narrow.

### Blocking

**Silent swallow on cascade unassign makes delete non-atomic in a dangerous direction.** `ee/cloud/projects/service.py` lines 1718-1751: `_unassign_project` catches all exceptions per-child and logs a warning, then proceeds to delete the project row. If any unassign step fails (transient Mongo error, version mismatch), the project is gone but children keep dangling `project_id` values that resolve to nothing. The unassign steps are simple bulk-update-many calls with no external dependencies — they should not be swallowed. Raise on failure (or wrap in a CloudError subclass) so the caller can retry.

### High-priority

**`@app.on_event("startup"/"shutdown")` is deprecated.** `ee/cloud/__init__.py` lines 403-412: the file's own comment at line 364-365 notes the host app uses a `lifespan` context manager which supersedes `on_event`. FastAPI 0.134+ still processes `on_event` for back-compat but it's a removal candidate. Convert the scheduler to the lifespan pattern, or document explicitly why `on_event` is acceptable here.

### Non-blocking

- `router.py:1416` — `list_projects` uses `# type: ignore[arg-type]` because of the `status` query param. A `Query(pattern="^(active|archived)$")` annotation drops the ignore and rejects bad values at the HTTP boundary.
- `cycles/router.py:394` — comment says "Re-fetch only to surface a clearer 404" but the re-fetch never happens. Trim the comment or add the re-fetch.
- No end-to-end test of cross-workspace `project_id` injection on Pocket/Task/Cycle creation. `_ensure_project_in_workspace` covers it in isolation but no integration test.

### What's right

- 4-file shape applied cleanly; textbook match to `pockets/`.
- Cascade-unassign order correct (children cleared before project row deleted).
- Each unassign helper lives in its own service file — 4-file rule preserved.
- `_fetch_project` returns uniform NotFound for missing + cross-tenant ids — prevents workspace enumeration.
- Scheduler genuinely idempotent (start/stop double-call safe, cancel-safe via `contextlib.suppress`).
- `_seconds_until_next_utc_midnight` clamps to 1.0s minimum against clock skew.
- import-linter contract entry added — touch-time migration satisfied.
- New docs page clearly distinguishes cloud vs local surface.

### Suggested merge order

Merge this PR first. paw-enterprise#191 depends on the REST surface this ships. paw-workspace#37 (playbook) is independent.

prakashUXtech:
Pushed `cda59281` — blocker addressed.

`_unassign_project` now narrows the except to `ImportError` only (preserving the lazy-import degrade for forks without a child entity). Real failures from the unassign calls propagate, which aborts `agent_delete` before the project row is touched. Children stay attached, caller can retry safely.

New test `test_delete_aborts_if_cascade_unassign_fails` monkeypatches the tasks unassign helper to raise, asserts `agent_delete` raises `RuntimeError`, verifies the project survives. All 28 project/cycle/MC tests pass; ruff clean.

The high-priority `@app.on_event` → `lifespan` conversion and the three non-blocking items are deferred to a follow-up unless you want them squashed in.

prakashUXtech:
Pushed a fix for the **"new task disappears"** symptom the captain caught via smoke test on PR #192.

### Root cause

`agent_list_work_items` in `ee/cloud/mission_control/service.py` only queried Instinct (Nudges). The Tasks entity shipped in this same PR was wired for create/read on its own router, but **never composed into the unified Mission Control feed**. So:

1. Modal POSTs to `/api/v1/tasks` → backend creates the task → returns `{id, status: 'in_progress', ...}`
2. Frontend refreshes via `GET /mission-control/items`
3. Façade reads from Instinct only → returns 0 items
4. Task is invisible despite being in MongoDB

Smoke trace from the browser console confirmed it:
```
[NewWorkItemModal] created OK { id: 6a08..., status: in_progress }
[MissionControl] onCreated → refreshing feed
[WorkFeed] listWorkItems → 0 items {}
```

### Fix

- `agent_list_work_items` now also queries Tasks via `tasks_service.agent_list_tasks` (lazy-imported, matches the projects/cascade pattern for forks without the Tasks entity).
- New `_task_to_work_item` helper projects `Task` → `WorkItem` with sensible status + section mapping. ID prefix `task:<id>` matches what the bulk-reassign/snooze endpoints already expect.
- Dropped the early `if not visible: return []` — that gated the whole feed on pocket visibility, which is correct for Instinct Nudges but wrong for workspace-scoped Tasks with null/empty pocket_id.

### Test coverage

- New `test_includes_tasks_alongside_nudges` is the regression: proves Tasks surface AND keep surfacing even when the workspace has zero visible pockets (which is what was happening — captain's task had `pocket_id: ''`).
- Three autouse fixtures got `agent_list_tasks` stubbed to `[]` so Instinct-only tests don't need a Beanie DB.

57 tests pass, ruff clean. Captain — hard-refresh the dev frontend after this lands on `ee` and the task should land in The Tray on create.

PR #1116, opened 2026-05-17
test(mission-control): end-to-end smoke for the full primitive chain

prakashUXtech:
## Summary

Adds a regression net for the 2026-05-16 \"task disappears\" bug + cross-tenant isolation + the cycle-snapshot HTTP endpoint. Drives the realistic operator flow against the merged \`ee\` branch in-process via TestClient + mongomock Beanie.

## What's covered

- **TestFullChain.test_project_then_task_surfaces_in_mission_control** — create project → create task scoped to project → confirm task surfaces in \`/mission-control/items\` → filter by project → delete project → confirm task survives with \`project_id=None\`. This is the regression for the bug where the façade only queried Instinct and silently dropped Tasks; the assertion has an inline diagnostic message so the next failure points straight at the regression.
- **TestTenancy.test_alpha_task_invisible_to_beta** — mounts two FastAPI apps with different workspace contexts; confirms a task created in \`w_alpha\` doesn't leak into \`w_beta\`'s MC feed.
- **TestCycleSnapshotEndpoint.test_manual_snapshot_returns_point_then_null_on_re_run** — exercises \`POST /cycles/{id}/snapshot\` (the HTTP form documented in Shift 12 of the intern playbook). Asserts the idempotent same-day behavior so interns who run it twice and see \`null\` know that's expected.

## Test plan

- [x] \`uv run pytest tests/cloud/test_mission_control_e2e_smoke.py\` — 3/3 pass
- [x] Full MC suite still green — 68/68 pass
- [x] \`uv run ruff check\` clean


prakashUXtech:
Heads up — this PR targets `ee`, which is now deprecated. PR #1157 synced `ee` into `dev` on 2026-05-21, so `dev` is the single canonical branch from here on.

This PR needs to be retargeted before it can merge — otherwise its base branch no longer exists as a merge target:

```
gh pr edit 1116 --base dev
```

Then rebase your branch onto `dev` and resolve any conflicts. `dev` already carries everything `ee` had, so the rebase is mostly your own commits replaying on top.

Issue #1117, opened 2026-05-17
security: harden JWT storage — move off localStorage to httpOnly cookie + Tauri keychain


prakashUXtech:
## Background

JWT (\`paw_token\`) is currently stored in browser \`localStorage\` and sent as \`Authorization: Bearer <token>\` on every request. The backend only validates the Bearer header — the same-named cookie is sent (\`credentials: 'include'\`) but never read server-side.

Discovered during the 2026-05-17 Mission Control live smoke. The captain spotted the token in DevTools localStorage and asked whether that's safe.

## Risk surface

| Threat | localStorage | httpOnly cookie |
|---|---|---|
| XSS \`fetch(...token)\` | steals instantly | unreachable from JS |
| Browser extension w/ broad \`host_permissions\` | full read access | unreachable |
| Malicious npm dep in build chain | runs in our origin, steals | unreachable |
| CSRF | safe (no auto-send) | needs SameSite + CSRF token |

For Tauri desktop the practical attack surface is narrow (sandboxed webview, no third-party extensions). For the web build the risk is real and grows with deployment surface.

### P1 — Backend cookie validation
- Wire FastAPI-Users (or whichever auth middleware honors the Bearer chain today) to also validate the \`paw_token\` cookie value. Today the cookie is set on login but never read server-side.
- Login response sets the cookie with \`HttpOnly; Secure; SameSite=Lax; Path=/; Max-Age=<jwt-ttl>\`.
- Add a CSRF token endpoint (\`GET /auth/csrf\`) + middleware that checks the \`X-CSRF-Token\` header against the value on state-changing verbs (POST / PATCH / PUT / DELETE).

### P1 — Frontend web build
- \`src/lib/core/shared/http.ts\`: stop reading the Bearer token from localStorage on web builds. The browser sends the cookie automatically with \`credentials: 'include'\`.
- For state-changing requests, fetch + cache the CSRF token; attach \`X-CSRF-Token\` header.
- Keep Bearer auth as an explicit code path for service-account / webhook clients.

### P2 — Tauri desktop build
- Migrate token storage from \`localStorage\` to \`tauri-plugin-stronghold\` (encrypted at rest in the OS keychain) or \`tauri-plugin-store\` with the same.
- Multi-window sync (\`token-updated\` Tauri event) continues to work — Rust side reads from the keychain, emits the event, frontend listens.
- \`src-tauri/src/oauth.rs\` already handles token persistence; swap the storage backend there.

[...]

PR #1119, opened 2026-05-17
feat(auth): cookie + CSRF chain alongside Bearer (security #1117 P1 backend)

prakashUXtech:
P1 of #1117. The web build can now authenticate via the HttpOnly `paw_auth` cookie that fastapi-users was already minting, with a double-submit CSRF token on state-changing verbs. Bearer auth stays live so the Tauri client and MCP / script callers keep working — P2 migrates Tauri to the OS keychain, then a follow-up audits the rest before Bearer drops.

## What ships

- `ee/cloud/auth/core.py`: pin `cookie_httponly=True` explicitly; make `cookie_secure` env-driven via `POCKETPAW_AUTH_COOKIE_SECURE` (defaults false for local HTTP dev, production must set true).
- `ee/cloud/_core/csrf.py`: new module. `CSRFMiddleware` checks `X-CSRF-Token` vs `paw_csrf` cookie on POST / PUT / PATCH / DELETE for cookie-authenticated callers. Bearer callers bypass (browsers never auto-attach `Authorization`). Bootstrap endpoints (login, logout, register, csrf, health) are exempt by path. `GET /auth/csrf` mints the token and sets the (non-HttpOnly) `paw_csrf` cookie so the web client can read it back as a header.
- `ee/cloud/__init__.py`: wire `CSRFMiddleware` after `TimingMiddleware`; mount `csrf_router` under `/api/v1/auth/csrf`.
- `ee/cloud/auth/router.py`: deprecation note on the bearer sub-router.

## Tests

- `tests/cloud/test_auth_cookie_chain.py` (6): login sets HttpOnly cookie, cookie-only authenticates `/auth/me`, bearer back-compat still works, logout clears the cookie, both backends stay registered, cookie transport attribute spec pinned.
- `tests/cloud/test_csrf_middleware.py` (9): token mint + idempotence, valid happy path, missing / mismatched header rejections, Bearer bypass, no-auth pass-through, GET skip, login exempt.

Run targeted: `uv run pytest tests/cloud/test_auth_cookie_chain.py tests/cloud/test_csrf_middleware.py tests/cloud/test_mission_control_router.py tests/cloud/test_projects_router.py -v` — 37 passed.

Wider auth-adjacent sweep (auth/, _core/, rbac, knowledge) — 140 passed.

## Auth-chain surprise

The job ticket assumed the cookie was named `paw_token`. The actual fastapi-users registration uses `paw_auth` — kept as-is because renaming would expire every live session in dev environments without a migration path. The cookie name is exported as `AUTH_COOKIE_NAME` from `ee/cloud/_core/csrf.py` so the frontend can import it from a single source if the build ever shares constants.

## Out of scope

- Tauri keychain migration (P2, separate PR).
- Dropping Bearer support (deferred until P2 + internal audit).
- Token rotation (separate issue).

## Frontend PR

Wired to https://github.com/qbtrix/paw-enterprise/pull/195 (URL added once frontend PR exists). The web `http.ts` chokepoint branches on `isTauri()` and now uses cookie + CSRF; Tauri stays on Bearer. 16 other paw-enterprise modules that read `localStorage.paw_token` directly migrate at touch-time in follow-ups (tracked in #1117).

### DIFF
--- a/ee/cloud/__init__.py
+++ b/ee/cloud/__init__.py
@@ -93,13 +93,18 @@ def mount_cloud(app: FastAPI) -> None:
     from ee.cloud._core.http import add_error_handler
     from ee.cloud._core.timing import TimingMiddleware
 
-    # Request-timing middleware first so it wraps every subsequent route
-    # (including CSRF rejections, which we want to see in perf data).
+    # Starlette's add_middleware is a stack — LAST registered runs OUTERMOST
+    # on inbound. Effective order here: CSRF → Timing → route handler.
+    # A CSRF 403 short-circuits before Timing observes the request, so perf
+    # data won't include rejected POSTs. That's a deliberate tradeoff: the
+    # CSRF gate exists to be fast and predictable, not measured. Reorder
+    # ONLY if you want Timing to wrap CSRF rejections (swap the two add_
+    # middleware calls — TimingMiddleware would then run outermost).
     app.add_middleware(TimingMiddleware)
 
-    # CSRF middleware — runs after Timing, before any route. Cookie-auth
-    # callers must echo X-CSRF-Token; Bearer-auth callers (Tauri, MCP,
-    # scripts) bypass entirely. See ``ee/cloud/_core/csrf.py``.
+    # CSRF middleware — outermost on inbound, runs before any route.
+    # Cookie-auth callers must echo X-CSRF-Token; Bearer-auth callers
+    # (Tauri, MCP, scripts) bypass entirely. See ``ee/cloud/_core/csrf.py``.
     app.add_middleware(CSRFMiddleware)
 
     # Global error handler — extracted to ee.cloud._core.http
--- a/ee/cloud/__init__.py
+++ b/ee/cloud/__init__.py
@@ -1,5 +1,11 @@
 """PocketPaw Enterprise Cloud — domain-driven architecture.
 
+Updated: 2026-05-17 (security #1117 P1) — Mounts ``CSRFMiddleware`` and
+    the ``/auth/csrf`` token endpoint so the web build can use the
+    HttpOnly auth cookie without exposing CSRF surface. Bearer-auth
+    clients (Tauri, MCP, scripts) skip the middleware entirely so
+    nothing breaks for existing callers; see ``ee/cloud/_core/csrf.py``
+    for the decision tree.
 Updated: 2026-05-16 — Mission Control backend completion. Mounts the
     Projects entity (workspace > project > pocket/task/cycle hierarchy)
     and wires the in-process daily-snapshot scheduler, gated on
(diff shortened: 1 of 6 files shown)


prakashUXtech:
## Review summary

**Verdict:** APPROVE WITH CHANGES — two required fixes before merge.

### Blocking

**1. Middleware registration comment contradicts Starlette stack semantics.** \`ee/cloud/__init__.py:96-98\` — last-registered middleware runs outermost on inbound. Actual order is \`CSRF → Timing → handler\`. CSRF 403 short-circuits BEFORE Timing captures it. Comment says the opposite ("Timing wraps CSRF rejections"). Functional behavior is correct; the comment will confuse anyone debugging a hung 403 and will tempt a "fix" that breaks the stack.

**2. \`paw_csrf\` cookie 7-day lifetime + no rotation on logout.** \`paw_auth\` is cleared on logout; \`paw_csrf\` isn't. A valid \`paw_csrf\` lives a week even post-logout — JS-readable, could be submitted on next login. With \`SameSite=Lax\` the threat is narrow but for a security-scoped PR this should be an explicit decision, not silence. Either document the deliberate choice or add logout-clears-csrf + test.

### High-priority

- **\`_COOKIE_SECURE\` imported from \`auth.core\` with leading underscore** — private name; \`auth.core\` is free to refactor it and break csrf.py at runtime. Export as a proper constant or move the env-read into csrf.py.
- **No wire-level HttpOnly assertion on \`paw_auth\`.** \`test_cookie_transport_pins_httponly_and_name\` checks the transport attribute, not the actual \`Set-Cookie\` header from \`/auth/login\`. Add one test that parses the response and asserts \`HttpOnly\` is present.

### Non-blocking

- \`_path_is_exempt\` uses \`startswith\` — \`/api/v1/auth/login-extra\` would match the \`/api/v1/auth/login\` exempt. No such route today but worth exact-segment matching.
- \`test_post_without_any_auth_passes_csrf_layer\` — pin \`assert res.status_code == 200\` instead of \`!= 403\`.

### What's right

- Decision tree in \`dispatch\` correctly ordered: safe methods → exempt paths → Bearer → no-cookie → finally validate
- \`secrets.compare_digest\` in the right place
- Token reuse on idempotent \`/auth/csrf\` calls (rotating per-request breaks concurrent tabs)
- \`_EXEMPT_PATH_PREFIXES\` covers both cookie + bearer login/logout variants
- Bearer-skips-CSRF design is correct; threat model doesn't apply when \`Authorization\` is present
- \`AUTH_COOKIE_NAME\` export gives FE a single source of truth
- Test structure: one file per concern, isolates real fastapi-users from middleware unit tests

### Suggested merge order

Fix the comment, the private-import, and the logout-clears-csrf decision. Then merge — independent of the planner chain.

PR #1120, opened 2026-05-17
feat(planner): plan_project tool wires deep_work into cloud Projects (#1118 P1)

prakashUXtech:
## Summary

New \`ee/cloud/planner/\` 4-file module that lets cloud Mission Control invoke the OSS Deep Work planner against a workspace Project, with all outputs materialized into existing cloud primitives. No edits to \`src/pocketpaw/deep_work/\` — the OSS public surface stays untouched per the captain's constraint.

## What lands

- **\`ee/cloud/planner/\`** — domain / dto / service / router. \`agent_plan_project(ctx, body)\` calls \`deep_work.start_deep_work()\`, then materializes:
  - PRD markdown → \`ee/cloud/uploads\` at \`/projects/{project_id}/prd.md\`
  - goal.md + plan.json → same folder
  - TaskSpec[] → \`ee/cloud/tasks\` rows with \`project_id\` set
  - AgentSpec[] → matched against \`ee/cloud/agents\`, misses returned as \`agent_gaps\` for the operator to act on
- **\`src/pocketpaw/agents/sdk_mcp_planner.py\`** — in-process MCP server wrapping the planner as a \`plan_project\` tool any Claude SDK agent in cloud chat can call
- **\`ee/cloud/uploads/service.py\`** — new \`write_text_file()\` helper for programmatic byte writes (the planner uses it; existing HTTP upload path unchanged)
- **\`ee/cloud/_core/realtime/events.py\`** — new \`PlanGenerated\` event so the Mission Control Plan tab can refresh without polling

## Endpoints

\`\`\`
POST /api/v1/planner/run            { project_id, goal, deep_research? }
GET  /api/v1/planner/by-project/{project_id}
\`\`\`

## Test plan

- [x] 14 new tests (9 service + 5 router), all pass
- [x] \`uv run ruff check\` clean
- [ ] Drive an end-to-end run against a real workspace Project from cloud chat once the companion frontend PR lands
- [ ] Verify the materialized PRD opens correctly in the Files panel

## Companion PR

paw-enterprise (next) — adds the **Plan** tab in Mission Control + the GeneratePlanModal that calls \`/planner/run\`.

## Tracks

- Closes part of #1118 (P1)
- P2 (UI), P3-P5 (agent gaps as Snags, task dependencies, replay) are follow-ups

Issue #1118, opened 2026-05-17
feat(ee): integrate deep_work as a plan_project tool for cloud Projects


prakashUXtech:
## Why

\`src/pocketpaw/deep_work/\` (OSS) and \`ee/cloud/projects/\` (EE) both use the word "Project" for different shapes. Deep Work is the planning intelligence layer (goal → PRD → TaskSpecs → AgentSpecs); Cloud Project is the multi-tenant container. They're at different layers, not duplicates, but the missed integration is real: EE clients can't drive Deep Work from their cloud workspace today.

Constraint that shapes the design: **deep_work is OSS and we can't change its public surface.** No renames, no breaking changes to \`Project\`, \`TaskSpec\`, \`AgentSpec\`, \`start_deep_work()\`.

The integration is EE-only and additive.

## What ships

A thin EE layer that lets a Cloud Project trigger Deep Work and absorb its outputs as first-class cloud artifacts.

### 1. New tool: \`plan_project\`

Registered with the cloud agent runtime (\`ee/cloud/chat/agent_service.py\` or wherever cloud tools live).

\`\`\`python
@tool
async def plan_project(
    ctx: RequestContext,
    project_id: str,
    goal: str,
    deep_research: bool = False,
) -> PlanProjectResult:
    """Decompose a goal into a PRD + tasks for a cloud Project.

    Invokes the local Deep Work runtime (start_deep_work) with the
    project metadata, then materializes the planner outputs to cloud
    primitives:
      - PRD markdown        → ee/cloud/files/  path=/projects/{id}/prd.md
      - TaskSpec[]          → ee/cloud/tasks/  with project_id set
      - AgentSpec[]         → matched against ee/cloud/agents/, gaps surfaced
    """
\`\`\`

### 2. File path convention

Deep Work outputs land in the existing files primitive under a project-scoped folder:

\`\`\`
/projects/{project_id}/
  ├─ prd.md               ← the main deliverable
  ├─ goal.md              ← the original parsed goal
  ├─ research/            ← if deep_research=true
  │   ├─ source-1.md
  │   └─ ...
  └─ plan.json            ← the raw PlannerResult for replay
\`\`\`

No schema change on \`ee/cloud/files\` — the path convention is enough. If multi-tenant querying later needs a first-class \`project_id\` field, that's a touch-time migration we can do then.

### 3. Task materialization

Each TaskSpec becomes a row in \`ee/cloud/tasks/\` with:
[...]

Issue #197, opened 2026-02-17
Docs use outdated dashboard port in getting-started pages


shivakarnati2004:
Summary: Some getting-started docs reference a dashboard URL/port that is inconsistent with current runtime defaults.
Impact: New users open wrong URL and assume install/startup failed.
Affected files:
quick-start.mdx
installation.mdx
Expected behavior: All docs point to the current default dashboard port.
Acceptance criteria:
Port references are consistent across quick-start + installation docs.
Verified against current app default config/runtime behavior.

DevRohit06:
This was fixed in PR #202 which updated the port references in the getting-started docs. Closing.

PR #1121, opened 2026-05-17
feat(planner): agent-gap resolution + task dependencies (#1118 P3 + P4)

prakashUXtech:
Two stacked shifts. Both build on #1120 (parent branch — will auto-retarget to \`ee\` when #1120 merges).

## P3 — agent-gap → create-agent flow

Plan sessions now persist as a \`PlanSession\` Beanie doc (\`ee.cloud.models.planner\`) so we can find the session again after the operator creates the missing agent. \`POST /api/v1/planner/resolve-gap\` takes \`{plan_session_id, spec_name, new_agent_id}\`, locates the human-fallback tasks for that spec, reassigns them to the new agent, strips the resolved spec from the persisted gap list, and emits \`PlanGapResolved\`. Fallback tasks now carry the wanted spec name on \`assignee.name\` and on \`source.metadata.wanted_agent_spec_name\` so the resolve flow can find the rows without parsing plan.json. The frontend creates the agent itself via \`POST /api/v1/agents\` — no new agent-creation route here.

## P4 — task dependencies

Added \`blocked_by: list[str]\` to the Task domain, DTO, and the Beanie doc. Update is tri-state — \`None\` leaves stored deps alone, \`[]\` clears them, a list replaces them outright. \`_materialize_tasks\` is now two passes: pass 1 inserts every task with empty \`blocked_by\` and builds a \`spec_key → task_id\` map, pass 2 patches deps via \`agent_update_task\` so forward references resolve correctly. Unresolved \`blocked_by_keys\` surface as \`PlanProjectResult.dependency_warnings\` instead of failing the run. The WorkItem projection threads \`Task.blocked_by\` through with the \`task:\` prefix so the frontend can dereference dependency edges without translating ids.

## Other touched bits

- \`PlanGapResolved\` registered in \`_core/realtime/events.py\`
- \`PlanSession\` added to \`ALL_DOCUMENTS\`
- New import-linter contract: \"Planner — Beanie writes only from service.py\"

## Tests

- \`test_planner_resolve_gap.py\` — 5 (happy path, multi-gap, three 404 cases)
- \`test_planner_task_dependencies.py\` — 3 (two-pass, forward refs, unknown dep with warning)
- \`test_tasks_blocked_by.py\` — 5 (create round-trip + tri-state update semantics)
- \`test_mission_control_service.py\` — extended for the prefixed \`blocked_by\` on the projected WorkItem

42 touched-area tests pass. Ruff + import-linter clean.

## Companion PR

paw-enterprise**#197** — wires the inline create-agent flow on the Plan tab + renders the blocked-by state in WorkFeed + WorkItemDetail.

## Tracks

- Closes part of #1118 (P3 + P4)
- P5 (replay + edit history) is the remaining follow-up

### DIFF
--- a/ee/cloud/_core/realtime/events.py
+++ b/ee/cloud/_core/realtime/events.py
@@ -398,3 +398,13 @@ class ProjectDeleted(Event):
 @dataclass
 class PlanGenerated(Event):
     EVENT_TYPE: ClassVar[str] = "plan.generated"
+
+
+# Planner — fires after ``ee.cloud.planner.service.agent_resolve_gap``
+# reassigns the human-fallback tasks for a previously-missing agent spec
+# to a newly-created cloud Agent. Listeners refresh the Plan tab's
+# agent-gap card (removing the resolved spec) and the Mission Control
+# feed (rows whose assignee changed).
+@dataclass
+class PlanGapResolved(Event):
+    EVENT_TYPE: ClassVar[str] = "plan.gap_resolved"
--- a/ee/cloud/mission_control/domain.py
+++ b/ee/cloud/mission_control/domain.py
@@ -4,6 +4,11 @@
 # into the unified frontend representation. Frozen dataclass per the ee/cloud
 # Code Rules (CLAUDE.md §3): workspace + assignee tenancy fields are required,
 # so constructing a WorkItem without them is a type error.
+# Updated: 2026-05-17 (feat/planner-gaps-and-deps) — pocketpaw#1118 P4
+# added ``blocked_by: tuple[str, ...]`` carrying ``task:<id>``-prefixed
+# WorkItem ids the row depends on. The projector in mission_control.service
+# threads ``Task.blocked_by`` through with the prefix so the frontend can
+# render dependency edges in the unified feed.
 """Mission Control domain value objects.
 
 Pure-Python frozen dataclasses. No Beanie, no Pydantic, no FastAPI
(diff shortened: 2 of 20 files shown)


prakashUXtech:
## Review summary

**Verdict:** REQUEST CHANGES — one data-loss bug, otherwise polish.

### Blocking

**\`dependency_warnings\` are generated but never persisted.** \`PlanSession\` Beanie doc has no \`dependency_warnings\` field; \`_persist_plan_session\` doesn't accept or write it. \`get_plan_for_project\`'s P3 fast-path constructs PlanSession without the field, so \`_session_to_dto\` always emits \`[]\` on cold hydrations. The warnings appear in the single \`agent_plan_project\` response then disappear — Plan tab refresh silently drops the signal the operator was supposed to act on.

Fix: add \`dependency_warnings: list[str] = Field(default_factory=list)\` to the Beanie doc, thread through \`_persist_plan_session\`, populate in the hydration path. Domain dataclass + \`_session_to_dto\` already handle it.

### High-priority

- **O(n²) in resolve-gap** (\`service.py\` ~line 214): \`r.id in session_doc.task_ids\` over a list. One-line fix: \`set(session_doc.task_ids)\` before the comprehension.
- **Redundant \`model_validate\`** (\`service.py:186\`): FastAPI already validated; the re-call is a no-op signaling boundary confusion. Drop it.

### What's right

- Tenant isolation on resolve-gap is correct — uniform NotFound (no 403 existence leak), cross-workspace tested end-to-end
- Tri-state \`blocked_by\` semantics (None / [] / list) — Beanie default \`[]\` means old rows read clean; all three states independently tested
- Two-pass materialization handles forward references; ordering in spec list is irrelevant
- \`PlanGapResolved\` event payload is complete — FE has everything for incremental patch (no refetch)
- \`task:\` prefix on WorkItem.blocked_by matches the existing convention — frontend resolves edges via string prefix check
- Import-linter contract enforces "Planner — Beanie writes only from service.py"

### Suggested merge order

Merge #1120 first (parent), fix the persistence bug + set conversion on this branch, then merge. paw-enterprise#197 can go after.

PR #1132, opened 2026-05-19
feat(calendar): native Calendar module (ee/calendar/)

prakashUXtech:
## Summary

New native `ee/calendar/` module — workspace-level calendar primitive that gives Pocketpaw events, recurrence, free/busy, and conflict detection without depending on an external calendar product. Closes the gap surfaced in the 2026-05-19 architecture discussion.

## What this PR adds

- `ee/calendar/` — 13 files
  - `domain.py` (frozen value objects: Event, Calendar, Attendee, Recurrence, FreeBusy + enums; `workspace_id` required at construction)
  - `dto.py` (distinct request/response DTOs with cross-field validators)
  - `service.py` (module-level async functions; validate-at-entry, tenant filter, bus emit on every write)
  - `router.py` (FastAPI routes mounted at `/api/v1/calendar`; service is called via aliased imports so it can be patched per-test)
  - `models.py` (Beanie `_EventDoc` + `_CalendarDoc` with the three required indexes; underscore-prefixed and never imported outside the module)
  - `events.py` (Pydantic bus event payloads + topic constants)
  - `recurrence.py` (pure dateutil RRULE expansion with exceptions + terminators + a 5000-instance safety cap)
  - `freebusy.py` (single-query availability compute with per-attendee windowing)
  - `conflicts.py` (overlap + shared-attendee detection)
  - `policy.py` (read/write/visibility checks; raise Forbidden on deny)
  - `sync.py` (gcalendar pull/push via `pocketpaw.integrations.gcalendar.CalendarClient`; outlook/icloud raise `NotImplementedError`)
  - `_context.py` (tiny `RequestContext` value object — the cloud module doesn't expose one yet, so this PR mints a private one rather than touching `ee/cloud/`)
  - `__init__.py` (re-exports the router + public domain types; Beanie docs stay private)
- `tests/ee/calendar/` — 5 test files, 34 passing tests covering service, recurrence, freebusy, and router. Service tests stand up an in-memory fake of the Beanie API so the suite stays Mongo-free.

## What this PR does NOT do

- Does not mount the router into the cloud app (`ee/cloud/api/`). Separate PR will wire it in once the design is validated.
- Does not implement Mission Control UI. Separate PR.
- Does not implement Outlook or iCal sync. Placeholder TODOs only; gcalendar adapter is the only working sync.
- Does not add an integration suite under `tests/cloud/`. Pure unit tests for now — the real-Mongo coverage lands when the cloud-app mount happens.

## Convention compliance

Mirrors the canonical ee/cloud shape used by `ee/cloud/pockets/`:

- [x] Domain stays at the boundary (frozen value objects); Beanie docs are private to `models.py`
- [x] Distinct request/response DTOs per operation
- [x] Service functions validate input at entry via `model_validate`
- [x] Every read carries `workspace=ctx.workspace_id` as the tenant filter
- [x] Mapping via Pydantic, not hand-rolled helpers
- [x] Bus emit on every state-changing write (`TOPIC_EVENT_CREATED/UPDATED/DELETED/CONFLICT_DETECTED`)
- [x] Errors via `CloudError` subclasses; router never raises `HTTPException`
- [x] No transactions — events as the cross-domain coupling

The 11-rule "exact mirror" spec from the ticket diverges from the actual ee/cloud codebase in a few places (no `RequestContext`, services are classes with statics, errors live in `shared/` not `_core/`). Where they disagreed, this PR follows what's actually shipping in `ee/cloud/pockets/` and adapts the ticket's file layout. The most consequential divergence: services are module-level `async def`s (per the ticket) rather than the `PocketService` static-method class that pockets uses. Happy to refactor either way once the team picks a canonical shape.

## Test plan

- [x] `uv run ruff check ee/calendar/ tests/ee/calendar/` passes
- [x] `uv run ruff format ee/calendar/ tests/ee/calendar/` clean
- [x] `uv run mypy ee/calendar/` clean (not in pre-commit but ran for safety)
- [x] `uv run pytest tests/ee/calendar/ -v` — 34 tests pass
- [x] `uv run pytest tests/ee/` — 142 tests pass (108 existing + 34 new)
- [ ] Captain + maintainer review

### DIFF
--- a/ee/calendar/__init__.py
+++ b/ee/calendar/__init__.py
@@ -1,0 +1,29 @@
+# Calendar module — public surface.
+# Created: 2026-05-19 (feat/calendar-module).
+#
+# Re-exports the router and public domain types. Beanie documents
+# (_EventDoc, _CalendarDoc) stay private — never re-exported.
+
+from ee.calendar.domain import (
+    Attendee,
+    AttendeeResponse,
+    Calendar,
+    CalendarVisibility,
+    ConflictSeverity,
+    Event,
+    FreeBusy,
+    Recurrence,
+)
+from ee.calendar.router import router
+
+__all__ = [
+    "Attendee",
+    "AttendeeResponse",
+    "Calendar",
+    "CalendarVisibility",
+    "ConflictSeverity",
+    "Event",
+    "FreeBusy",
+    "Recurrence",
+    "router",
+]
--- a/ee/calendar/_context.py
+++ b/ee/calendar/_context.py
@@ -1,0 +1,26 @@
+# Calendar module — RequestContext value object.
+# Created: 2026-05-19 (feat/calendar-module).
+#
+# Tiny carrier struct so service functions can take a single ctx parameter
+# (per ee/cloud convention rule 5). The cloud module doesn't expose a shared
+# RequestContext type yet — when it does, swap this for the canonical one.
+# Kept private (underscore prefix) so the router stays the single mounting
+# point for callers outside this module.
+
+from __future__ import annotations
+
+from pydantic import BaseModel, ConfigDict
+
+
+class RequestContext(BaseModel):
+    """Per-request tenancy + actor envelope.
+
+    Carries the workspace_id (multi-tenant filter) and user_id (actor for
+    audit + access checks) into service functions. Frozen so handlers can
+    pass it around without worrying about mutation.
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+    workspace_id: str
+    user_id: str
--- a/ee/calendar/conflicts.py
+++ b/ee/calendar/conflicts.py
@@ -1,0 +1,67 @@
+# Calendar module — conflict detection.
+# Created: 2026-05-19 (feat/calendar-module).
+#
+# Finds other events in the same workspace that overlap a given event's
+# window and share at least one attendee. Excludes the event itself.
+# Returns mapped Event domain objects (DB → domain via _doc_to_event).
+
+from __future__ import annotations
+
+from ee.calendar.domain import Attendee, Event, Recurrence
+from ee.calendar.models import _EventDoc
+
+
+def _doc_to_event(doc: _EventDoc) -> Event:
+    """Map a Beanie document into the frozen Event domain value object."""
+    attendees = [Attendee.model_validate(a) for a in (doc.attendees or [])]
+    recurrence = Recurrence.model_validate(doc.recurrence) if doc.recurrence else None
+    return Event(
+        id=str(doc.id),
+        workspace_id=doc.workspace,
+        calendar_id=doc.calendar_id,
+        title=doc.title,
+        description=doc.description or "",
+        starts_at=doc.starts_at,
+        ends_at=doc.ends_at,
+        timezone=doc.timezone,
+        location=doc.location,
+        attendees=attendees,
+        recurrence=recurrence,
+        fabric_object_id=doc.fabric_object_id,
+        source_connector=doc.source_connector,
+        source_external_id=doc.source_external_id,
+        created_at=doc.created_at,
+        updated_at=doc.updated_at,
+    )
+
+
+async def find_conflicts(workspace_id: str, event: Event) -> list[Event]:
+    """Find other events overlapping `event`'s window with shared attendees.
+
+    Two events overlap when start_a < end_b AND start_b < end_a. We push that
+    filter into Mongo so we don't pull the whole workspace into memory. The
+    shared-attendee filter runs in Python because attendees are stored as an
+    embedded list of dicts.
+    """
+    if not event.attendees:
+        return []
+
+    overlapping_docs = await _EventDoc.find(
+        {
+            "workspace": workspace_id,
+            "starts_at": {"$lt": event.ends_at},
+            "ends_at": {"$gt": event.starts_at},
+        }
+    ).to_list()
+
+    target_emails = {a.email.lower() for a in event.attendees}
+    conflicts: list[Event] = []
+    for doc in overlapping_docs:
+        # Skip the event itself.
+        if str(doc.id) == event.id:
+            continue
+        other_emails = {(a.get("email") or "").lower() for a in (doc.attendees or [])}
+        if target_emails & other_emails:
+            conflicts.append(_doc_to_event(doc))
+
+    return conflicts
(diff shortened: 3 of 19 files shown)


prakashUXtech:
## Convention follow-up (from #1133 work)

After verifying `ee/cloud/pockets/` directly, the Calendar module here is using a slightly different shape than mainline:

- Calendar service: `async def op(ctx: RequestContext, body) -> EventResponse`
- Pockets canonical: `async def op(workspace_id: str, user_id: str, body) -> dict`

The doc in `pocketpaw/CLAUDE.md` previously described the Calendar shape as the target (which is why your agent followed it); the doc has now been updated to match the actual pockets pattern via #1138.

Two options for resolution:

1. **Refactor Calendar to match pockets** before merge — change service signatures to `(workspace_id, user_id, body)`, return wire dicts, drop the private `_context.py:RequestContext`. ~1-2 agent-hours.
2. **Accept the deviation** as an experiment toward a future RequestContext migration. The doc now explicitly mentions Calendar as a forward-looking example. New modules would still follow pockets; pockets would migrate to RequestContext on a future touch-time.

Captain's call. No blockers from my side on either path.

prakashUXtech:
## Code review — `ee/calendar/` (PR #1132)

Scope: `ee/calendar/` and `tests/ee/calendar/` only. The 57 divergence files outside that tree (agent backends, analytics, frontend templates, etc.) are dev-vs-ee drift and were ignored per the review brief.

---

### Blockers

**policy.py is fully wired up but never called — `ee/calendar/service.py` has no import of it**

`policy.py` ships three enforcement functions: `check_calendar_read`, `check_calendar_write`, and `check_event_visibility`. `service.py` never imports or calls any of them. The result is that any authenticated user in a workspace can read, update, or delete any event in any calendar in that workspace, regardless of the calendar's visibility setting or `shared_with_user_ids`. A `PRIVATE` calendar is as readable as a `PUBLIC_TO_WORKSPACE` one.

This is a functional gap, not a theoretical one. The domain model, the policy module, and the `Calendar` doc are all in place — the wiring just didn't land. The fix is straightforward: `create_event` needs to fetch the parent calendar and call `check_calendar_write`; `get_event` and `list_events` need `check_event_visibility` or `check_calendar_read` depending on the desired granularity.

Note: this is distinct from the *workspace* tenant filter, which is enforced correctly everywhere via `workspace=ctx.workspace_id`. The gap is intra-workspace calendar-level visibility, not cross-tenant leakage.

---

### High-priority (should fix before merge)

**`list_events` router parses ISO strings with bare `datetime.fromisoformat` — a malformed value returns a 500**

`router.py` accepts `starts_after` and `starts_before` as raw `str` parameters and parses them inside the handler:

```python
starts_after=datetime.fromisoformat(starts_after),
```

A malformed string raises `ValueError` which FastAPI wraps as an unhandled 500, bypassing the `CloudError` envelope. FastAPI can parse `datetime` query params natively — declare them as `datetime` in the function signature and it handles both parsing and the 422 response automatically. Alternatively, wrap in a try/except and raise a `ValidationError`.

**`push_to_gcalendar` silently loses the external ID after push**

The PR description acknowledges this (`"the next pull will reconcile"`), but the code has a practical problem: after `push_to_gcalendar` returns the external ID, nothing in the call stack writes it back to `_EventDoc.source_external_id`. The next pull will create a *duplicate* event rather than reconcile, because the existing doc has no `source_external_id` to match on. The function already returns the ID — the caller just needs to save it. Worth a follow-up issue at minimum, or a comment that explicitly calls out "this creates duplicates until source_external_id is written back" so the next person knows it isn't just a TODO.

---

### Non-blocking nits

**`recurrence.py`: `_normalize` has an asymmetric path for naive datetimes**

When `dt.tzinfo is None`, `_normalize` returns `dt` unchanged. When it is tz-aware, it converts to UTC-naive. If an exception in `rec.exceptions` is naive and the expansion generates tz-aware occurrences (which happens when `DTSTART` carries a tzinfo), the set-membership check silently never fires — exceptions are skipped. This is a data-integrity edge case, not a crash, but the comment at line 922 notes the problem without resolving it. A single guard — always return naive UTC — would make both paths consistent.

**`test_service.py`: the `_FieldProxy("id_str")` / `doc.id_str = oid` bridge is fragile**

The fake store maps `_EventDoc.id` to a proxy on `"id_str"`, then manually attaches `doc.id_str = oid` in `_new_doc`. This works today but breaks if `_get_event_doc_or_404` is ever refactored to use a different comparison form. Worth a comment explaining the indirection so the next person maintaining the fake doesn't break it silently.

**Cross-workspace test coverage is partial**

`test_list_events_tenant_filter` and `test_get_event_cross_workspace_returns_404` exist, which is good. Missing: a cross-workspace test for `update_event` and `delete_event`. Both are protected by `_get_event_doc_or_404` (same code path as `get_event`), so this is low risk, but the pattern in the test file implies coverage for each operation and the gap is noticeable.

**`list_events` Mongo filter uses `ends_at > starts_after`**

This is correct for overlap semantics, but the field name `starts_after` in the DTO means "events starting after this time" to most readers, while the actual filter returns any event whose window overlaps the query window (including events that started before `starts_after` but end inside it). The field name is the inconsistency — either rename to `window_start` / `window_end` in the DTO, or add a one-line comment explaining overlap semantics. Minor but has caused support confusion in similar systems.

---

### What's right

The multi-tenancy enforcement on every DB call is clean and consistent. Every `_EventDoc.find` and `_EventDoc.find_one` in `service.py`, `freebusy.py`, and `conflicts.py` carries `"workspace": workspace_id` as a first-class filter, not an afterthought. `_get_event_doc_or_404` returns 404 (not 403) on cross-workspace lookups — the correct choice that doesn't leak existence. The 404-not-403 pattern is applied uniformly.

The `Recurrence.count` / `RRULE COUNT` interaction is handled correctly: the more restrictive terminator wins in both directions without double-applying.

`sync.py` is an honest implementation. `pull_from_gcalendar` is real, callable code with proper reconciliation logic; `push_to_gcalendar` is real code with a documented gap; `pull_from_outlook` and `pull_from_icloud` are explicit `NotImplementedError` stubs that will fail loudly if invoked, which is exactly right. No silent no-ops.

The `bus_spy` fixture in `conftest.py` is well-designed — it patches `event_bus.emit` cleanly, captures in emission order, and restores on teardown.

`RequestContext` (`_context.py`) is minimal and correct: `workspace_id` and `user_id` both required with no defaults, frozen, private to the module. The comment explaining the swap path when `ee/cloud` exposes a canonical one is the right documentation for a known deviation.

---

### Net read

The workspace-level tenant boundary is solid, and the structural work (domain, DTOs, bus events, mapping, tests) is well-executed. The policy module being fully written but not connected to the service layer is the one real issue before merge — it ships working calendar CRUD but with calendar-level visibility rules that have no effect.

Issue #1142, opened 2026-05-19
Calendar module — security follow-ups from PR #1132 review (H + M + L findings)


prakashUXtech:
## Context

Security audit of #1132 surfaced 3 High, 5 Medium, and 3 Low findings. The PR is being merged to `ee` to unblock downstream work (router mount #1139, paw-enterprise /calendar route); this issue tracks the post-merge cleanup.

Net read from the audit: **pass-with-followups** — architecturally sound, multi-tenant boundary consistently applied, but the policy layer needs wiring and free/busy needs scope hardening.

### H1 — Wire `ee/calendar/policy.py` into the service layer
`policy.py` defines `check_calendar_read`, `check_calendar_write`, and `check_event_visibility`, but `service.py` never imports or calls any of them. Tenant boundary (`workspace=ctx.workspace_id`) prevents cross-workspace reads but does nothing within a workspace — any authenticated workspace member can read, mutate, or delete any other member's events.

**Fix**: call the appropriate `policy.check_*` function at the top of each CRUD service operation. Raise `Forbidden` via `CloudError` on denial.

`ee/calendar/service.py` lines 87–271. `ee/calendar/policy.py` (defined but never called).

### H2 — Restrict free/busy to requester-accessible attendees
`compute_freebusy` accepts an arbitrary list of attendee emails and returns busy windows across the entire workspace. A workspace member can enumerate every email address in the org and reconstruct the full schedule — a covert availability oracle.

**Fix**: restrict the attendee list to emails the requester has explicit access to (own calendar + calendars shared with them + workspace members where the workspace policy allows freebusy queries). Reject unknown emails with `ValidationError`.

`ee/calendar/freebusy.py:39`. `ee/calendar/service.py:225-234`.

### H3 — Handle malformed `starts_after` / `starts_before` query params
`list_events` calls `datetime.fromisoformat()` on raw query strings without try/except. Malformed input (`?starts_after=notadate`) raises unhandled `ValueError` → HTTP 500. `_core.http` only maps `CloudError`, not `ValueError`.

**Fix**: either use FastAPI's native `datetime` query parameter type or wrap the parse with try/except → `ValidationError`.

`ee/calendar/router.py:77-91`.

## Medium

- **M1** — `Recurrence.rrule` has no `max_length`. `dateutil.rrulestr` parses synchronously — a multi-megabyte RRULE is a DoS vector before the expansion cap helps. Add `Field(max_length=2048)` to the domain model.
[...]

PR #1143, opened 2026-05-19
fix(calendar): wire policy + restrict freebusy + handle date parsing (#1142)

prakashUXtech:
## Summary

Addresses the three High security findings from #1132's security audit (per issue #1142). Plus four Mediums that were tractable in scope.

## Findings addressed

| Finding | File:line | Fix |
|---|---|---|
| H1 - policy.py dead code | service.py multiple | Wired check_calendar_read/write into every CRUD service fn |
| H2 - freebusy oracle | freebusy.py / service.py | Restrict attendee emails to requester-accessible set |
| H3 - datetime 500 | router.py list_events | Use FastAPI native datetime type -> 422 on bad input |
| M1 - RRULE no length cap | domain.py | Field(max_length=2048) |
| M2 - exceptions unbounded | domain.py | Field(max_length=500) |
| M3 - email not validated | domain.py | EmailStr |
| M4 - bus event leaks raw content | service.py update_event | Changed dict to log field names only, not values |

## Deferred to follow-up

- M5 audit log emission (separate concern - needs AuditLogger.log wiring decision across the cloud)
- L1 push_to_gcalendar credentials
- L2 sync.py content caps
- L3 sync.py raw-event logging

## Tests added

- tests/ee/calendar/test_policy.py (new) - authz scenarios
- Expanded test_freebusy.py - attendee-access restriction
- Expanded test_router.py - malformed datetime returns 422

## Test plan
- [x] ruff check / format pass
- [x] pytest tests/ee/calendar/ all green (69 passed)
- [x] Targets ee (NOT dev - ee-separation directive)
- [ ] Captain review

### DIFF
--- a/ee/calendar/conflicts.py
+++ b/ee/calendar/conflicts.py
@@ -1,5 +1,10 @@
 # Calendar module — conflict detection.
-# Created: 2026-05-19 (feat/calendar-module).
+# Updated: 2026-05-19 (fix/calendar-security-hardening, #1142 H-NEW-1).
+#
+# Changes:
+# - _doc_to_event now propagates created_by_user_id from the Beanie doc
+#   into the domain Event. Required for policy.check_event_modify on
+#   update/delete paths.
 #
 # Finds other events in the same workspace that overlap a given event's
 # window and share at least one attendee. Excludes the event itself.
@@ -24,6 +29,10 @@ def _doc_to_event(doc: _EventDoc) -> Event:
         starts_at=doc.starts_at,
         ends_at=doc.ends_at,
         timezone=doc.timezone,
+        # H-NEW-1: hydrate the creator field so check_event_modify works
+        # on Events returned by conflict scans (and any other domain
+        # transit through this helper).
+        created_by_user_id=doc.created_by_user_id,
         location=doc.location,
         attendees=attendees,
         recurrence=recurrence,
--- a/ee/calendar/domain.py
+++ b/ee/calendar/domain.py
@@ -1,5 +1,5 @@
 # Calendar module — domain value objects.
-# Updated: 2026-05-19 (fix/calendar-security-hardening, #1142 M1-M3).
+# Updated: 2026-05-19 (fix/calendar-security-hardening, #1142 M1-M3 + H-NEW-1).
 #
 # Changes:
 # - Attendee.email now uses EmailStr (M3) — Pydantic's email-validator
@@ -9,6 +9,10 @@
 #   pathological RRULE strings that could choke recurrence expansion.
 # - Recurrence.exceptions capped at 500 entries (M2) — guards against
 #   unbounded growth that would slow down conflict detection.
+# - H-NEW-1: Event now requires created_by_user_id (no default). The
+#   service sets it from ctx.user_id on create_event. policy.
+#   check_event_modify gates update/delete on creator-equality so a
+#   synthetic-default Calendar doesn't grant cross-user modify access.
 #
 # Frozen Pydantic models representing the public calendar domain.
 # workspace_id is required on Event and Calendar — enforced at construction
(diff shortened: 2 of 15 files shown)


prakashUXtech (commented):
## Code review — fix/calendar-security-hardening (#1142)

**Net read: merge with two follow-ups tracked.** No blockers. Two items worth noting before or shortly after merge.

---

### What's right

**H1 wiring is complete.** Every CRUD path — `create_event`, `get_event`, `update_event`, `delete_event`, `list_events`, `detect_conflicts`, and `get_freebusy` — gates on `policy.check_calendar_read` or `check_calendar_write` before any data access. No gaps.

**`_load_calendar` synthetic default is well-designed.** The fallback enforces the tenant boundary (the `find_one` query includes `workspace: ctx.workspace_id`), sets owner to the caller's `user_id`, and uses `PUBLIC_TO_WORKSPACE` — which replicates the pre-#1142 permissive behaviour rather than silently widening it. The comment in service.py makes the bridge intent clear. Importantly, a bad `calendar_id` from a different workspace does not get a synthesized default — it simply returns None from the workspace-filtered `find_one` and then synthesizes with `workspace_id=ctx.workspace_id`, so the tenant boundary cannot be slipped via a foreign `calendar_id`. Well thought through.

**H2 attendee filter has consistent error messaging.** All paths that reject unknown emails converge on a single `ValidationError("calendar.unknown_attendee", ...)` and the comment explicitly explains why the message stays generic (don't reveal whether an email exists on a private calendar). The `emails_lower` normalisation also prevents case-variant probing.

**H3 is the correct fix.** Switching `starts_after`/`starts_before` to FastAPI-native `datetime` types is the right approach — Pydantic owns the validation boundary, the router doesn't need a try/except, and the 422 envelope is structurally identical to every other validation failure. The test covers both params plus the missing-param case.

**M4 bus payload is correct.** `changed_fields` carries only names. The `attendees_count` addition is fine — a count is not PII and gives downstream subscribers enough signal to decide whether to re-render without knowing content. All seven mutable fields (`title`, `description`, `starts_at`, `ends_at`, `location`, `attendees`, `recurrence`) appear in the changed_fields list, so coverage is complete.

**test_policy.py coverage is strong.** Cross-workspace denial, private-calendar denial for non-owner, workspace-public write denial (the easy-to-miss case), shared-user read and write grants, the synthetic-default fallback on all service functions, and the freebusy oracle both denied (private calendar) and allowed (public calendar) are all exercised. 572 lines but tightly structured with clear test names.

---

### High priority (should fix before merge)

**CI: `tests/test_mcp_claude_sdk.py` is failing.** Seven tests fail with a `pocketpaw_planner` server appearing in MCP configs when it shouldn't. These failures are *not* in calendar code and the failures appear in the run history before this PR's push, so this is a pre-existing regression on `ee`, not introduced here. Confirm that and get those tests green before merge — merging on a red CI gate, even for an unrelated failure, makes the next author's blame harder.

---

### Non-blocking nits

**`_load_calendar` TODO is implicit.** The comment says "until Calendar CRUD ships" but there is no `# TODO(<issue>)` reference. A single line like `# TODO(#NNN): remove synthetic default once Calendar CRUD surface exists` would make it searchable and prevent the bridge from living forever.

**`sync.py` L3 `fromisoformat` is still unhandled.** `sync.py:58–59` calls `_parse_iso` (a local wrapper) which presumably still uses `fromisoformat` under the hood. This is explicitly deferred as L3 in the PR description, so it is not a problem for this PR — just confirming it is still open so the follow-up issue (#1142) captures it. If `_parse_iso` already catches ValueError, the L3 finding may already be soft-closed.

**`errors` import path.** service.py imports from `ee.cloud.shared.errors` (the shim). The CLAUDE.md rule says new code should import from `ee.cloud._core.errors` directly. Calendar is new code. Not a blocker, but worth aligning on the next touch.

---

### Convention question — captain decision

The Calendar module uses `RequestContext(workspace_id, user_id)` rather than the canonical `async def op(workspace_id: str, user_id: str, body: ...)` pair documented in CLAUDE.md. The CLAUDE.md note explicitly acknowledges Calendar as an example of the `RequestContext` preview pattern and says the mainline stays explicit-parameter until `pockets` migrates. This PR does not make that question worse. Flagging for awareness: if the `pockets` migration lands before Calendar CRUD ships, the Calendar service will need a touch-time alignment pass.

PR #1147, opened 2026-05-20
fix(claude-sdk): only release the client lease when the run owns it

prakashUXtech:
## Symptom

A Claude SDK agent would sometimes cut itself off mid-response. The log showed `Main loop exited without ResultMessage — destroying persistent client`. The agent's stream broke, its loop exhausted with no `ResultMessage`, and the run ended early.

## Root cause

`ClaudeSDKBackend.run()` is one async generator instance shared across every concurrent session of an agent. It dispatches on the bool `self._client_in_use`:

- flag False: persistent path, sets the flag True and streams from the shared persistent client.
- flag True: stateless fallback, streams from a throwaway `query()`.

The bug class is the same on every exit path: a run that never acquired the lease still mutated the shared lease and client.

1. The `finally` block cleared `self._client_in_use = False` unconditionally, even for a run that took the stateless-fallback path and never acquired the lease. A stateless run finishing released a still-streaming sibling persistent run's lease. A later run then saw the flag False, entered the persistent path, and collided on the shared `_client` subprocess.
2. The outer `except Exception` handler did the same on the error path — it cleared `self._client_in_use` and nulled `self._client` unconditionally. A stateless-fallback run that failed hard, or a run that failed before it ever acquired the lease, would steal the sibling's lease and destroy its subprocess.
3. `_persistent_client` was assigned before the persistent path was known to succeed. If the persistent attempt threw and the run fell back to stateless, `_persistent_client` was left as a dangling non-None reference, and the teardown guard could misfire and disconnect the shared `self._client`.

## The fix

`run()` now tracks ownership with a local `acquired_lease` flag, set at the exact point the run takes ownership of the persistent client. The flag is declared above the `try` block so it is always in scope for the `except` handler.

Both exit paths gate on it:

- The `finally` block only clears `self._client_in_use` and runs the persistent-client teardown when `acquired_lease` is true.
- The outer `except` handler only clears the lease and nulls `self._client` when `acquired_lease` is true.

A run that never acquired the lease leaves the flag and the shared subprocess alone, whether it exits cleanly or by exception. The persistent-path fallback handler resets `acquired_lease`, so a dangling `_persistent_client` can no longer misfire the teardown.

The existing check-then-set shape of the bool is unchanged. There is no move to `asyncio.Lock` or per-session backends — that is a larger refactor and out of scope here.

## Verification

The branch carries three reproduction tests in `tests/test_fast_path.py`:

- `test_stateless_fallback_does_not_clear_sibling_lease` — finally path.
- `test_stateless_fallback_does_not_destroy_shared_client` — secondary teardown invariant.
- `test_stateless_fallback_error_path_does_not_clear_sibling_lease` — error path.

Each failed before its fix and passes after. Targeted suites all pass — `test_fast_path.py` (12), `test_stream_event.py`, `test_mcp_claude_sdk.py`, `test_agents.py`, `test_agent_loop.py` (87 tests total).

## Scope

Internal concurrency fix. No public API, CLI, or behavior-surface change, so no doc updates are needed.


prakashUXtech:
Closing — wrong base branch. The cloud backend where this bug lives is on the ee track, not dev. Reopening the same fix against ee.

PR #1154, opened 2026-05-21
refactor(mcp): gate the planner MCP server behind an explicit policy opt-in

### LINKED ISSUE #1150: ee branch CI is red: TestClaudeSDKMCPServers fails after the planner MCP server landed
The `pytest` CI job fails on the `ee` branch. Seven tests in `tests/test_mcp_claude_sdk.py::TestClaudeSDKMCPServers` fail on a clean checkout of `ee` HEAD (`de120627`) with no PR changes applied, so this is branch breakage rather than a PR regression.

## Failing tests
- `test_disabled_server_filtered_out`
- `test_enabled_stdio_server_passes`
- `test_http_server_without_url_skipped`
- `test_multiple_servers_mixed`
- `test_no_mcp_configs`
- `test_policy_denies_group_mcp`
- `test_policy_denies_server`

## Root cause
Commit `7f9191ff` (planner: `plan_project` tool wires deep_work into cloud Projects) added `pocketpaw_planner` as a built-in SDK MCP server to `_get_mcp_servers()`. The test helper `_strip_builtin_servers` in `test_mcp_claude_sdk.py` does not strip `pocketpaw_planner`, so every test that counts external servers is off by one.

Example from `test_no_mcp_configs`:
```
assert len(external) == 2
AssertionError: assert 3 == 2
  where 3 = len({'fs': ..., 'pocketpaw_planner': {'type': 'sdk', ...}, 'web': ...})
```

## Suggested fix
Update `_strip_builtin_servers` to also drop `pocketpaw_planner` (and any other built-in SDK servers), or adjust the affected assertions. The helper should have been updated in `7f9191ff`.

## Impact
`ee` CI stays red for every PR until this is fixed. Unrelated PRs against `ee` (such as #1148) get a red `pytest` check they did not cause.

prakashUXtech:
## What this does

`pocketpaw_planner` is an in-process MCP server exposing one tool, `plan_project`. It was registered unconditionally in `ClaudeSDKBackend._get_mcp_servers()`, so every agent run carried the `plan_project` tool schema, including agents that will never plan a project. It was the only in-process MCP server with no policy gate.

This PR makes the planner opt-in per agent. A cloud agent enables it by listing the token `pocketpaw_planner` in its `tools` field. Every other agent drops the schema.

## Why the first gate alone was not enough

The branch started as a pure gate: register the planner only when the policy says so. But the tool policy is allow-by-default for MCP servers. With the default `full` profile and an empty allow list the resolved allow set is empty, and `is_mcp_server_allowed` returns `True` for everything. Copying the pocket-specialist gate verbatim would have left the planner loaded on every run.

An earlier revision of this branch fixed that by checking the explicit `tools_allow` set instead. That made the planner genuinely off-by-default, but it created a worse problem: nothing anywhere puts a planner entry into `tools_allow`. Not a config file, not an agent definition, not the cloud Agent entity. So the planner would have become unreachable code. A working feature shipped four days ago in #1118 would have turned into dead weight.

This PR closes that gap with the opt-in path.

## The design

A dedicated `mcp_servers_allow: frozenset[str]` field on `ToolPolicy`, kept separate from `tools_allow`.

Reusing `tools_allow` was rejected. Any `mcp:*` entry there makes the resolved `_allowed_set` non-empty, which flips the policy out of allow-by-default into allow-list mode and silently disables every other tool and external MCP server. That is the exact regression we had to avoid. The new field is orthogonal: `tools_allow` keeps its current meaning, and `mcp_servers_allow` is read only by `is_mcp_server_explicitly_allowed`. Deny still wins over it.

`AgentPool._build()` is the translation boundary. It reads the cloud agent's `config.tools`, keeps any entry that names a known built-in MCP server (currently just `pocketpaw_planner`, via the module constant `_BUILTIN_MCP_SERVER_TOKENS`), and builds a per-agent `ToolPolicy` with that frozenset. Users put the bare token `pocketpaw_planner` in the agent's `tools`, not the internal `mcp:...:*` notation. Unknown tokens are dropped without error.

`ClaudeSDKBackend.__init__` takes an optional `policy` argument. When provided it is used as-is; when omitted the policy is built from settings exactly as before, which is what every other caller relies on.

## Multi-backend safety

The reviewer flagged a real bug in the original design: `_build()` selects a backend from `settings.agent_backend`, and the design passed `policy=` to whatever class came back. There are eight registered backends: `ClaudeSDKBackend`, `OpenAIAgentsBackend`, `GoogleADKBackend`, `CodexCLIBackend`, `OpenCodeBackend`, `CopilotSDKBackend`, `DeepAgentsBackend`, and `LangchainReactBackend` (which subclasses `DeepAgentsBackend`). I checked every one. The `AgentBackend` Protocol declares `__init__(self, settings)`, and all seven non-Claude backends take only `settings`. Passing `policy=` to any of them would raise `TypeError` on every agent using that backend.

So `_build()` branches on the resolved backend class: if `backend_cls is ClaudeSDKBackend`, construct with `policy=`; otherwise construct with `settings` alone as today. The branch keys on the class object, not on `settings.agent_backend`, so legacy backend names that remap to `ClaudeSDKBackend` through the registry are handled correctly. Only the Claude SDK backend reads the policy, so this is the right boundary.

I also confirmed nothing passes positional args after `settings` to `ClaudeSDKBackend(...)`, the cloud agent config key really is `"tools"` (`ee/cloud/models/agent.py`), and `is_mcp_server_explicitly_allowed` has no caller other than the planner gate.

## Migration

Every existing agent has an empty `tools` list, so `mcp_servers_allow` is empty and the planner stays off. Nothing else about any agent changes. To enable the planner for an agent:

```
PATCH /agents/{id}   {"tools": ["pocketpaw_planner"]}
```

## Issue #1150

`ee` CI was already red before this branch: seven `TestClaudeSDKMCPServers` tests failed because the test helper `_strip_builtin_servers` never learned that `pocketpaw_planner` is a built-in, so external-server counts were off by one. The gate makes the planner off-by-default in the tests' default policy, so those seven go green on their own. `_strip_builtin_servers` is also updated to drop `pocketpaw_planner` so the helper stays correct for any test that opts the planner in.

`Fixes #1150`

## Tests

TDD throughout: failing tests first, confirmed red, then the implementation.

- `tests/test_tool_policy.py::TestMCPExplicitAllow` drives the opt-in through the new `mcp_servers_allow` argument. Covers the empty case, the populated case, deny-wins (server wildcard and `group:mcp`), the case where a populated `tools_allow` does not leak into the MCP opt-in, and the default-empty argument.
- `tests/cloud/test_agent_pool_planner_opt_in.py` is new. It unit-tests `AgentPool._build()` with a stubbed backend and agent doc. Five cases: `tools=[]` leaves the planner off; the `pocketpaw_planner` token turns it on; the non-regression case where a populated global `tools_allow` stays intact and `Bash` / `read_file` / external MCP servers still resolve; `tools_deny` wins over the token; an unknown token is dropped without a crash.
- `tests/test_mcp_claude_sdk.py::TestPlannerMCPGate` is reworked to inject a `ToolPolicy` whose `mcp_servers_allow` names the planner, and adds a case proving an `mcp:*` entry in `tools_allow` does NOT opt the planner in.

```
uv run pytest tests/test_tool_policy.py tests/test_mcp_claude_sdk.py tests/test_fast_path.py tests/cloud/test_agent_pool_planner_opt_in.py
71 passed
```

Broader sweep across `test_agents.py`, `test_rails.py`, headless and audit suites: 226 passed. One unrelated failure, `test_agents.py::test_auto_resolve_no_key_gives_ollama`, reproduces on a clean `ee` checkout and is about provider auto-resolution, not this change. Three `agent_router` cloud tests also fail on a clean checkout (a `_RecordingPool` test double missing the `instructions` kwarg) and are likewise unrelated.

`ruff check` and `ruff format --check` clean on every changed file. `claude_sdk.py` has one pre-existing format nit at an untouched line near 925, left alone to keep the diff focused.

## Docs

There is no doc surface describing the planner MCP server or the in-process MCP registration policy. The agent-config wiki article documents the `tools` field generically and is auto-generated by a build hook, so it is not hand-edited here. The new behavior is documented in code: the file-top comment blocks of all three changed source files, the `_BUILTIN_MCP_SERVER_TOKENS` docstring, and the gate comment in `claude_sdk.py`.

## Not merging

Draft until the captain reviews.


prakashUXtech:
Marking this draft. The gate itself is correct, but verification found it incomplete: nothing in the codebase maps any per-agent config into `tools_allow`, and the planner was built to be ambient (registered for every agent, no designated planner agent). Merging this alone makes `plan_project` unreachable by every agent.

It needs a companion change that wires an opt-in path before it can land. The issue #1150 test-helper fix included here is sound and can be salvaged separately if the approach changes.

Issue #1163, opened 2026-05-21
pocket_specialist__edit: silent 0-ops on large project-dashboard canvas (6 attempts, same pocket)


prakashUXtech:
## Summary

The `pocket_specialist__edit` MCP tool silently returns `ok=true` with `ops=[]` on every edit attempt against a specific large pocket (`Sprint May-2 · Calendar + Composio`). No error is returned, but no change lands on the canvas either. Same backend (`deep_agents`) succeeded earlier in the session when *creating* this pocket, so the create path works — only the edit path is stuck.

This is also the second session in which I've hit the same 0-ops-no-error pattern (prior session: adding a map widget to a travel planner pocket — same silent rejection).

## Environment

- **Surface:** paw-enterprise chat (PocketPaw chat agent)
- **Pocket ID:** `6a0eb47ac0dd58069139b985`
- **Workspace:** `69e4f93b57ff64b3903868e3`
- **Owner:** `69d0f4d09dfb6ddfd8e2da84` (prakash@snctm.com)
- **Tool:** `mcp__pocketpaw_pocket_specialist__edit`
- **Backend reported:** `deep_agents`
- **Session date:** 2026-05-21
- **Pocket size:** large — root flex with 6 children, includes a `project-dashboard`, a `kanban` with 26 cards, and a 16-row PR `table`. RippleSpec JSON is roughly 12KB+.

## Reproduction

Pocket was created earlier in the same session via `pocket_specialist__create` and successfully extended once (the initial team rebuild that produced `Gaurv (handle TBD)` — that worked). After that point, every subsequent `edit` call returns 0 ops.

## Attempt log (all in one session, same pocket id, same backend)

| # | Intent | Approach | Result | duration_ms |
|---|---|---|---|---|
| 1 | Replace `team[3]` `Gaurv (handle TBD)` → `Gaurav Dewani (dewani12)` | Natural-language intent describing the array-item swap | `ok:true, ops:[]` | 1746 |
| 2 | Set entire `team` array to a 4-element list (only diff = entry 4) | Explicit full-array intent | `ok:true, ops:[]` | 1342 |
| 3 | Same edit, requesting `set_node_prop` explicitly | Named the op in the intent | `ok:true, ops:[]` | 1333 |
| 4 | Add new `section` with a bar `chart` of merged-PRs by author, inserted between Sprint board and Open PRs sections | Positional structural add | `ok:true, ops:[]` | 1352 |
[...]

PR #1177, opened 2026-05-21
Read-only live data sources for pockets (RFC 04 alpha)

prakashUXtech:
## What this does

Pockets render a static `rippleSpec` today — a one-time data snapshot. This PR adds the first piece of RFC 04: a pocket can be bound to one external backend and declare read-only GET "sources" in its spec. A server-side executor runs the GETs and returns the JSON; the frontend applies the results to the pocket's ripple state.

Alpha is read-only — GET bindings only. Write bindings are the next milestone.

## How it works

- A pocket is bound to one backend: a base URL plus an auth credential, set once via `PUT /pockets/{id}/backend`.
- The spec's `rippleSpec.sources` declares GET bindings — a path, a `state` path to write into, and a refresh policy (`pocket_open` / `manual`).
- `POST /pockets/{id}/sources/run` loads the spec, runs the selected sources against the backend, and returns the results.

Three routes added to the existing pockets router:

- `PUT /pockets/{id}/backend` — bind a pocket to a backend
- `GET /pockets/{id}/backend` — read the binding summary (never the token)
- `POST /pockets/{id}/sources/run` — run the read-only sources

## Security model

**The credential lives in its own collection.** `PocketBackendCredential` is a separate Beanie document in `pocket_backend_credentials` — never on the `Pocket` doc, never in `rippleSpec`. The spec stays shareable and secret-free. The token is encrypted at rest with AES-GCM; the key is derived per row via HKDF-SHA256 over `AUTH_SECRET` with a random salt. Reads return only `base_url` / `auth_type` / `configured` — the plaintext token only ever leaves the collection to be handed straight to the executor.

**The executor is an SSRF boundary.** `source_executor.py` is the one pocket-domain module that makes outbound HTTP, and it carries the full set of defenses from the security review:

- base URL re-validated strict at call time (https-only, no internal hosts, no env escape hatch)
- relative paths only — absolute URLs, `..` traversal, and cross-host joins are rejected
- DNS check before each call so a public name that resolves to a private IP is blocked
- redirects disabled — any 3xx is a source error
- tight connect/read/write timeouts and a 512 KB response cap
- error messages sanitized — no raw exception text, query strings stripped from logged URLs
- per-pocket rate limit, 10 runs per minute

## Delivery channel

The hydrated state comes back in the `POST /sources/run` HTTP response body, not via a `pocket_mutation` SSE event. The run endpoint is a standalone REST call made outside any SSE-stream context, so there is no live stream to emit onto — the caller reads the results from the response and patches ripple state client-side. The RFC has been amended to record this.

## One of three PRs

This is 1 of 3 for the RFC 04 alpha:

1. **This PR** — backend binding, the source executor, the run endpoints (pocketpaw)
2. ripple — the `run_source` action type
3. paw-enterprise — the backend-config UI and the wiring that calls `/sources/run` on pocket open and on the refresh button

There is no merge-order dependency for the tests in any one repo, but all three are needed for the feature end to end.

## Tests

New tests under `tests/cloud/`:

- `test_pocket_backend_crypto.py` — encrypt/decrypt round-trip, salt uniqueness, strict URL validation
- `test_pocket_source_executor.py` — SSRF rejections, size cap, refresh-policy selection, bind-path stripping, redirect handling, rate-limit breach, auth headers
- `test_pocket_backend_service.py` — set/get/remove, upsert, token never returned, workspace scoping
- `test_pocket_backend_routes.py` — the three endpoints

52 new tests pass. The full cloud suite shows no new failures attributable to this change — the pre-existing `uploads/` and `test_pocket_prompt_state_sources.py` failures reproduce identically on `dev`.

## Docs

- New endpoints documented in `docs/api-reference.md`
- RFC 04 amended: status moved to "Alpha — in implementation", delivery-channel note added

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -6,6 +6,10 @@ Created: 2026-05-21 (RFC 04 alpha) — documents the per-pocket backend
 binding + read-only source-run endpoints. The rest of the cloud pockets
 API is described in the auto-generated wiki article
 `ee/docs/wiki/pockets-router-*.md`.
+
+Updated: 2026-05-21 (PR #1177 security pass) — documented the new
+DELETE /pockets/{id}/backend endpoint and the edit-access requirement on
+GET /pockets/{id}/backend.
 -->
 
 # Cloud REST API Reference
@@ -46,10 +50,15 @@ Response `200`:
 The token is never echoed back. A non-https or internal `base_url` yields
 a `400`.
 
+For `basic` auth, send `auth_token` as the raw `user:pass` credential —
+the server base64-encodes it into the `Authorization: Basic` header. Do
+not pre-encode it yourself.
+
 ### `GET /pockets/{pocket_id}/backend`
 
-Read the pocket's backend binding summary. Read access mirrors
-`GET /pockets/{pocket_id}`.
+Read the pocket's backend binding summary. Requires pocket **edit** access
+(owner or editor) — backend config metadata is owner/editor-facing,
+consistent with the `PUT` route. Viewers receive a `403`.
 
 Response `200`:
 
@@ -60,10 +69,23 @@ Response `200`:
 Returns `404` when the pocket has no backend configured. The token is
 never included in the response.
 
+### `DELETE /pockets/{pocket_id}/backend`
+
+Revoke the pocket's backend binding — deletes the stored (encrypted)
+credential. Requires pocket **owner** access.
+
+Returns `204 No Content`. Idempotent: deleting when no backend is
+configured still returns `204`. The removal is written to the audit log.
+
 ### `POST /pockets/{pocket_id}/sources/run`
 
 Run the pocket's read-only `rippleSpec.sources` (GET bindings) against its
-configured backend. Read access mirrors `GET /pockets/{pocket_id}`.
+configured backend. Read access mirrors `GET /pockets/{pocket_id}` —
+deliberately **not** gated on edit access. Any pocket reader may run the
+already-authored sources: a viewer of a shared live pocket triggering the
+`pocket_open` refresh is the core shared-dashboard UX. A viewer cannot
+change the backend or the source paths (both are edit-only), so the SSRF
+hardening plus the immutable, edit-authored source list bound the risk.
 
 Request body (all fields optional):
 
@@ -99,4 +121,6 @@ Returns `400` when the pocket has no backend configured.
 the base URL, rejects absolute-URL paths / `..` traversal / cross-host
 joins, runs a DNS check against internal IPs, disables redirects, applies
 tight timeouts, caps response bodies at 512 KB, sanitizes error messages,
-and rate-limits to 10 runs per pocket per minute.
+and rate-limits to 10 runs per `(pocket, user)` pair per minute. Every run
+is written to the audit log (actor, pocket, status, query-stripped base
+URL) — the credential token is never logged.
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -1,0 +1,102 @@
+<!--
+docs/api-reference.md — Hand-maintained reference for cloud REST endpoints
+that are not covered by the per-endpoint Mintlify pages under docs/api/.
+
+Created: 2026-05-21 (RFC 04 alpha) — documents the per-pocket backend
+binding + read-only source-run endpoints. The rest of the cloud pockets
+API is described in the auto-generated wiki article
+`ee/docs/wiki/pockets-router-*.md`.
+-->
+
+# Cloud REST API Reference
+
+This file documents cloud (`pocketpaw-ee`) REST endpoints that do not yet
+have a dedicated page under `docs/api/`. All cloud endpoints require a
+valid enterprise license and an authenticated workspace context.
+
+## Pockets — Backend Binding & Live Data Sources
+
+RFC 04 alpha. A pocket can be bound to **one** external backend (base URL +
+auth credential). Its `rippleSpec.sources` declares read-only `GET`
+bindings; a server-side executor runs them and returns the JSON results.
+
+The credential is stored in a **separate, encrypted collection**
+(`pocket_backend_credentials`) — never inside the `Pocket` document and
+never inside `rippleSpec`, so the spec stays shareable and secret-free.
+
+### `PUT /pockets/{pocket_id}/backend`
+
+Bind a pocket to one backend. Requires pocket **edit** access.
+
+Request body:
+
+| Field | Type | Notes |
+|-------|------|-------|
+| `base_url` | string | Required. Must be `https://` and point to an external host (no loopback / RFC1918 / link-local). |
+| `auth_type` | string | One of `bearer`, `api_key`, `basic`, `none`. |
+| `auth_token` | string | The secret. Encrypted at rest; never returned. Required unless `auth_type` is `none`. |
+| `auth_header` | string \| null | Custom header name for `api_key` auth. Defaults to `X-Api-Key`. |
+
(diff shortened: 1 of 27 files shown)


prakashUXtech:
## Security review pass

Pushed `b93ca91b` to address every item from the audit.

**Blockers**
- `basic` auth was sending the stored token raw after `Basic ` — a malformed header. It now treats the token as a `user:pass` string and base64-encodes it.
- There was no way to revoke a configured credential — `DELETE /pockets/{id}/backend` was missing even though the service function existed. Added the route (owner-only, idempotent, audit-logged).

**Should-fix**
- `GET /{id}/backend` now requires edit access, matching the PUT route. Viewers don't need backend config metadata.
- `POST /{id}/sources/run` keeps read access on purpose — a viewer of a shared live pocket has to be able to trigger the `pocket_open` refresh, and they can't touch the backend or the source paths. Documented that reasoning in the route docstring. What changed: the rate limiter is now keyed per `(pocket, user)` so one member can't drain another's budget.
- The rate limiter's check-and-record was a read-filter-write with no lock — a TOCTOU race under `asyncio.gather`. Wrapped it in an `asyncio.Lock`.
- The executor was importing the private `_host_is_internal`. Added a public `host_is_internal` alias in `url_validators` and switched to it.
- Source runs make authenticated outbound calls but wrote nothing to the audit log. Every run now logs an entry (actor, pocket, status, query-stripped base URL) on the same path the backend-config functions use. The token is never logged.

**Nits**
- Empty-string `source` coerces to `None`; empty `base_url` is now a 422 at the DTO layer.
- `PocketBackendCredential` dropped from `models __all__` (still registered for Beanie init).

**Verification**
- `tests/cloud/test_pocket_backend_*` + `test_pocket_source_executor.py`: 61 passed. New tests cover the base64 encoding, the DELETE route, the 403 for a non-editor, the per-user rate-limit key, async-safety under `gather`, and the audit entries.
- `ruff check` and `mypy` clean on every changed file (one pre-existing `add_agent` mypy warning is unrelated).
- Both import-linter contracts still pass.
- The unrelated `tests/cloud/uploads/` failures in the full sweep are pre-existing on `dev` — confirmed against the base commit with these changes stashed.

prakashUXtech:
Follow-up: closed the edit-path gap on the sources feature.

The first pass let a pocket carry a `rippleSpec.sources` block when it was created, but the edit specialist couldn't add one to a pocket that already exists. Its granular ops only reached `state` and `ui`, and its prompt never mentioned `sources` or the `run_source` action. So when someone asked it to "add a live data source," it stashed a fake descriptor in state and wired up a refresh button that just sent a chat message — wrong on both counts.

What this adds:

- `sources_ops.py` — pure `set_source` / `remove_source` helpers, the Beanie-free sibling of `state_ops.py` for the third top-level rippleSpec key.
- `agent_set_source` / `agent_remove_source` in `pockets/service.py` — they validate the binding through the existing `SourceBinding` model (an invalid or write-verb binding is rejected before anything is written), persist it to the Pocket doc, and emit an event.
- `set_source` / `remove_source` edit tools, registered in `make_edit_pocket_tools`.
- An edit-flavored sources block in the specialist prompt, plus the tool inventory updated so it stays accurate.

The mutation side reuses the full-document `replace` action the frontend already handles, so there's no paw-enterprise change here — this stays a pocketpaw-only fix. `sources_ops.py` is in the import-linter contract so it can't pull in Beanie.

Tests: `tests/cloud/test_pocket_sources_ops.py` covers the pure ops, the service functions (happy path, invalid-binding rejection, persistence, idempotent removal), the agent_context wrappers, and the tool factories. The full pocket suite (296 tests) passes, ruff and mypy are clean on the new code, and all 15 import-linter contracts hold. One stale prompt test was also fixed — it had been checking the wrong prompt constant since the #1163 split.

prakashUXtech:
Follow-up 2: the edit specialist can now see the backend that's already configured.

Testing the last fix surfaced the next gap. A pocket can have a backend connected (the "Configure Backend" modal shows it as "Connected"), but when someone asked the edit specialist to "rewire this to use a real data source," it would ask back: "what's your backend, what endpoint should this point at?" It had no idea a backend was already set, because the credential lives in its own collection and nothing handed the agent even the non-secret part of it.

What this adds:

- `get_pocket` / `agent_view` now return a `backend` field — `{base_url, auth_type, configured}`, the same shape `get_pocket_backend` already produces. The token is never part of it; a fetch failure just degrades to `{configured: false}` so a backend hiccup can't break the pocket view.
- The edit-specialist prompt's `<current-pocket>` block gained a `Backend:` line, e.g. "configured — https://jsonplaceholder.typicode.com (auth: none)" or "not configured". A new `fill_current_pocket` helper fills both the pocket-id and the backend-summary tokens, and every prompt-fill site routes through it so the new token can't leak as literal text.
- The sources guidance now tells the specialist to read that line. If a backend is configured, author the source against it directly — a source `path` is always relative to the configured base URL, so the agent only ever needs the relative path and should never ask for the backend URL it can already see. If it genuinely can't infer the path, it asks only for the relative path.

This stays pocketpaw-only — no frontend change, the token never reaches agent context. Tests cover the backend summary in `agent_view` for configured and not-configured pockets, that the token never appears anywhere in the agent view, workspace scoping, and the token substitution. The full pocket suite (280 tests) plus the broader specialist/ripple sweep pass; ruff and mypy are clean on the touched files (mypy net error count actually dropped two); all import-linter contracts hold. One stale test was aligned to the post-#1163 prompt split.

prakashUXtech:
## Deterministic fix for the agent's hallucinated data-source output

Reproduced twice against MongoDB ground truth: the pocket-authoring agent does not produce the RFC 04 `rippleSpec.sources` shape. Every run it instead writes:

- `rippleSpec.tool_specs: [{id, kind:"rest", method, url, auto_fetch, into}]` — an invented schema. None of `kind`/`url`/`auto_fetch`/`into` exist anywhere in the codebase, and `tool_specs` is not even a real rippleSpec field (the real `tool_specs` is a top-level Pocket field for LLM tools).
- the refresh button as `{action:"run_source", source_id:"..."}` — wrong field name; the runtime reads `handler.source`.

The runtime only reads `rippleSpec.sources` and only `handler.source`, so the agent's output is inert. Prompt instructions alone have not shifted the model off this prior.

### What changed

`normalize_ripple_spec` now runs a repair step (`_lift_rest_sources`) before any other normalization. Because every persist path — create and edit — routes through the normalizer, this covers both.

1. **Lift `tool_specs` REST entries into `sources`.** Heuristic: `kind == "rest"` or (has a `url`/`path` and an `into`/`bind`). Each becomes `{method, path, bind, refresh}` — absolute URLs reduced to relative paths, `auto_fetch` mapped to a refresh policy, non-GET entries skipped (alpha is read-only). A correctly-authored `sources` block is never clobbered. The bogus `tool_specs` key is dropped once empty.
2. **Repair `run_source` buttons.** Walks the UI tree and pane trees, renames `source_id` to `source`, and binds a sourceless handler when there is exactly one source. Defensive against single-object and array handler slots.

Also hardened the create and edit prompt blocks with blunt WRONG-vs-RIGHT guidance so the model has a shot at getting it right unaided.

### Tests

New `tests/cloud/test_pocket_sources_lift.py` — written red first. Feeds the exact hallucinated shape and asserts it normalizes to a working `sources` block plus a repaired button. Covers pass-through of correct specs, `auto_fetch:false`, absolute URLs, non-GET skip, handler arrays, and sole-source binding. 10 new tests pass; the full `test_pocket_*` suite stays green at 290. Ruff, mypy, and import-linter clean on the touched files.

Issue #238, opened 2026-02-18
Files access from within docker container


aptonline:
I'm running PocketPaw on my Mac via a docker container and can successfully access the `pocketpaw_pocketpaw-data` volume from within Docker Desktop and see all the config files but I can't see where the files section (ie any files that PocketPaw creates) is within the volume.

<img width="300" height="259" alt="Image" src="https://github.com/user-attachments/assets/96e50541-97e0-49c4-b440-cb55dc03ca1e" />

How do I access files created via PocketPaw locally on my host?

DevRohit06:
## Fix — PR #254

All issues resolved in #254:

### What was broken
1. Agent-created files landed in the container's ephemeral filesystem — lost on restart, inaccessible from host
2. Claude Code CLI and Codex CLI not installed — `claude_agent_sdk` and `codex_cli` backends failed immediately
3. Health check referenced wrong package name (`claude-code-sdk` instead of `claude-agent-sdk`) — false FAIL on `--doctor`
4. Host-side Ollama unreachable from the container (`localhost` inside Docker ≠ host machine)

### What was fixed
- **Workspace volume**: `./workspace` bind mount for agent-created files
- **CLI backends**: Node.js 22 + Claude Code CLI + Codex CLI installed in image
- **Health check**: Corrected package name to `claude-agent-sdk`
- **Host networking**: `host.docker.internal` mapped via `extra_hosts`
- **Docs**: Updated `docker.mdx` and `README.md`

### Using Ollama with Docker
- **Ollama in Docker**: `POCKETPAW_OLLAMA_HOST=http://ollama:11434`
- **Ollama on host**: `POCKETPAW_OLLAMA_HOST=http://host.docker.internal:11434`

Issue #41, opened 2026-02-12
Surprise us: Build a skill or plugin you'd personally use


prakashUXtech:
## Description

This is an open-ended challenge. PocketPaw has a skill/plugin system — build something **you** would actually use.

## The challenge

1. Look at the existing built-in tools in \`src/pocketclaw/tools/builtin/\` to understand the pattern
2. Build a new tool or skill that does something genuinely useful to you
3. It should follow the \`ToolProtocol\` interface (see \`src/pocketclaw/tools/protocol.py\`)

## Ideas to spark creativity (but don't limit yourself)

- A Hacker News / Product Hunt monitor
- A personal finance tracker (parse bank CSV exports)
- A Pomodoro timer with focus session tracking
- A Git repo analyzer (PR stats, code churn, contributor insights)
- A bookmark manager with AI-powered categorization
- A habit tracker
- A meeting notes summarizer
- A code snippet library

## What we're looking for

- **Originality** — Did you think of something interesting?
- **Quality** — Is it well-implemented? Tests? Error handling?
- **Practicality** — Would someone actually use this?
- **Code style** — Does it follow the project's conventions?

## Difficulty

🔵 Creative — No ceiling. Show us what you can do.

hkgowda2974:
Hi! I’d like to work on this challenge.

I’m planning to build a GitHub Repository Analyzer tool that provides contributor insights, PR statistics, and activity summaries.

I’ll follow the ToolProtocol interface and include proper tests and error handling.

Looking forward to contributing!

prakashUXtech:
Moving to backlog — we're focused on installation reliability and bug fixes for v0.3. See #115 for how to contribute during stabilization. We'll revisit feature work once the install path is solid across all platforms.

PR #1179, opened 2026-05-22
Pocket write actions: increment 1 (RFC 05 M2a)

prakashUXtech:
## What this is

Increment 1 of RFC 05 — the write half of the pocket data layer. RFC 04's alpha made pockets live: read bindings hydrate `state`. This makes them do things — a pocket can submit a form, mark a row done, or trigger a backend operation through a declared `rippleSpec.actions` write binding.

This is 1 of 3 increment-1 PRs. The other two are ripple #41 and paw-enterprise #238.

## The write executor

A new `action_executor.py` runs `POST`/`PUT`/`PATCH`/`DELETE` bindings against a pocket's single configured backend. It is the write counterpart to RFC 04's read-only `source_executor.py`.

A write has blast radius a read does not, so the executor adds three things on top of the SSRF guards:

- **A per-pocket write allowlist.** A `(method, path)` that does not match an `allowed_writes` entry is rejected before any call leaves the server. The allowlist lives on the backend config, outside `rippleSpec`, so the agent that authors a binding cannot widen its own write blast radius. The default is empty — fail-closed, no write fires until a human owner sets a policy.
- **Fail-closed instinct-reject.** M2a has no Instinct approval surface. Rather than accept a `requires_instinct` flag and then quietly ignore it, any action whose declaration carries a truthy `requires_instinct` is rejected with `code: instinct_required` and makes no call. M2b wires the real approval routing.
- **Idempotency.** Every call carries an `Idempotency-Key` header — a client-supplied key when present, so a write retried after a timeout dedupes, otherwise a server-generated one.

The write rate limit (20 writes per `(pocket, user)` per minute) is a separate counter from the read executor's, so a read budget never drains the write budget or vice versa.

## Shared `_http_guard` extraction

The SSRF defenses RFC 04's read executor enforced — path-traversal rejection, absolute-URL rejection, same-host assertion, DNS rebinding check, no redirect following, tight timeouts, the 512 KB response cap, error sanitization — are pulled out verbatim into one canonical `_http_guard.py` module that both executors import. It is a pure refactor; the read-executor regression tests are the proof gate and pass unchanged.

## Two new endpoints

- `POST /pockets/{id}/actions/run` — runs a declared write action. The HTTP method is read server-side from the persisted spec, so a compromised client cannot pick the verb. Access is owner or explicit `shared_with` only — a workspace-visible pocket does not grant run access, because a write has blast radius.
- `PUT /pockets/{id}/backend/write-policy` — sets the write allowlist. Owner-only.

## Authoring path

The normalizer lifts the agent's hallucinated inline-write shape (`{action: "api", method: "POST", url: <relative>}`) into a proper `actions` block plus a `call_binding` handler. An absolute-URL `api` call is left alone — it targets a third party and must never be redirected onto the credentialed backend. The pocket specialist gets `set_action` / `remove_action` granular tools and prompt guidance for both the create and edit paths.

## Tests and gates

- New: `test_pocket_action_executor.py`, `test_pocket_actions_lift.py`, `test_pocket_actions_ops.py` cover the executor, the normalizer lift, and the ops + service layer. The existing backend and agent-backend-view tests are updated to the new `allowed_writes` contract.
- `ruff check` and `ruff format --check` pass.
- `mypy` is clean on every touched file — zero errors over the baseline.
- `pytest tests/cloud/` — all 60 new tests pass. The 54 pre-existing failures sit in unrelated areas (tasks, notifications, connectors, agent_router) and are untouched by this change.
- import-linter — all 15 contracts hold; the three new modules stay Beanie-free.

## Docs

`docs/api-reference.md` documents the two endpoints. RFC 05 is amended — open questions #1 (block naming), #3 (reconcile cost), and #6 (idempotency) are marked resolved.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -10,6 +10,10 @@ API is described in the auto-generated wiki article
 Updated: 2026-05-21 (PR #1177 security pass) — documented the new
 DELETE /pockets/{id}/backend endpoint and the edit-access requirement on
 GET /pockets/{id}/backend.
+
+Updated: 2026-05-22 (RFC 05 M2a) — documented the write-action endpoints
+(POST /pockets/{id}/actions/run, PUT /pockets/{id}/backend/write-policy)
+and the per-pocket write allowlist now carried on the backend summary.
 -->
 
 # Cloud REST API Reference
@@ -44,11 +48,18 @@ Request body:
 Response `200`:
 
 ```json
-{ "base_url": "https://api.example.com", "auth_type": "bearer", "configured": true }
+{
+  "base_url": "https://api.example.com",
+  "auth_type": "bearer",
+  "configured": true,
+  "allowed_writes": []
+}
 ```
 
 The token is never echoed back. A non-https or internal `base_url` yields
-a `400`.
+a `400`. `allowed_writes` is the per-pocket write allowlist (RFC 05 M2a) —
+empty by default, so no write action can fire until an owner sets a policy
+via `PUT /pockets/{id}/backend/write-policy`.
 
 For `basic` auth, send `auth_token` as the raw `user:pass` credential —
 the server base64-encodes it into the `Authorization: Basic` header. Do
@@ -63,11 +74,17 @@ consistent with the `PUT` route. Viewers receive a `403`.
 Response `200`:
 
 ```json
-{ "base_url": "https://api.example.com", "auth_type": "bearer", "configured": true }
+{
+  "base_url": "https://api.example.com",
+  "auth_type": "bearer",
+  "configured": true,
+  "allowed_writes": [{ "method": "POST", "path_pattern": "/leases/*/renew" }]
+}
 ```
 
 Returns `404` when the pocket has no backend configured. The token is
-never included in the response.
+never included in the response. `allowed_writes` carries the current
+write allowlist (RFC 05 M2a).
 
 ### `DELETE /pockets/{pocket_id}/backend`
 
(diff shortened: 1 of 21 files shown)


prakashUXtech:
## Security review hardening (commit 79f45d72)

Four fixes from the M2a security audit. No blockers in the audit; these are the SHOULD-FIX and NIT items on the write executor.

**S1 — allowlist matches the decoded path.** The allowlist check used to glob the human-authored `path_pattern` against a still-encoded request path. It now decodes the path once with `urllib.parse.unquote` before matching, so a client can't flip the verdict by percent-encoding the request. The pattern stays as the human wrote it and is matched as-is.

**S2 — backend HTTP status no longer reaches the client.** A backend `>=400` response used to echo the exact status code back in the error message, which turned the endpoint into a path-probing oracle for the configured backend. The number now goes only to the audit log via a new `_BackendHTTPError`. The client gets a generic "the backend rejected the request" message under the same `http_error` code.

**N1 — empty-params DELETE sends no body.** A DELETE with no params used to still send `json={}`. Some backends and WAFs reject a DELETE that carries a body, so the body is now dropped in that case. A DELETE with params still sends it.

**N3 — workspace_id on write-action audit entries.** `_audit_action_run` logged `pocket_id` but not `workspace_id`. The route now passes `workspace_id` through, so write-action audit entries are tenant-filterable like the backend-config entries already are.

New executor tests cover encoded-path matching (allowed and rejected cases), the generic `http_error` surface, and the empty- vs non-empty-params DELETE body. All 70 tests in the three pocket suites pass; ruff and mypy are clean on the touched files.

PR #1183, opened 2026-05-22
feat(pockets): Instinct routing + outcome emission (RFC 05 M2b)

prakashUXtech:
## What this does

Increment 1 shipped pocket write actions, but its executor fail-closed
rejected any action marked `requires_instinct` — it returned
`instinct_required` and made no call, because there was no approval
surface to route through. This PR builds that surface.

Two pieces, both from RFC 05 M2b:

**M2b.1 — Instinct routing.** A `requires_instinct` write is no longer
rejected. It is parked: the executor runs every cheap validation gate
(rate-limit, base-URL, SSRF, allowlist, DNS) so a write the allowlist
would reject is still rejected up front, then returns an
`instinct_pending` sentinel instead of firing. The router routes the
parked write into an Instinct Action; once a human approves it, the
write runs. The approver defaults to the pocket owner, with an optional
per-pocket route to a named workspace member.

**M2b.2 — outcome emission + a minimal meter.** A write whose binding
declares an `outcome` emits a `pocket.outcome` event after it succeeds.
A new `outcomes` entity appends each event to a per-workspace JSONL
ledger and exposes `GET /api/v1/outcomes` for counts.

`approve_batch` is deferred. `approve_per_row` plus the null policy is
the shipped surface — batch approval is a follow-up.

## Changes

M2b.1:
- `ActionBinding` declares the real fields `requires_instinct`,
  `instinct_policy`, and `outcome`. A validator rejects an
  `instinct_policy` set without `requires_instinct`. The old
  `_instinct_rejected` raw-dict gate is gone.
- `run_action` takes a `from_instinct` flag. A direct run of a
  `requires_instinct` binding validates the write then returns
  `{ok, code: instinct_pending, _park}` and makes no HTTP call. The
  executor stays pure — no Beanie, no Instinct imports.
- New `pockets/instinct_bridge.py`: `propose_pocket_write` builds the
  Instinct Action; `execute_approved_write` re-loads the backend creds,
  re-enters `run_action` with `from_instinct=True`, records the result,
  and emits the outcome. A revoked backend fails closed.
- The action-run route branches on `instinct_pending` and returns
  `proposed_action_id`. The instinct approve hook fires
  `execute_approved_write` best-effort via a lazy import.
- Per-pocket `approval_route` on the backend config, set owner-only via
  `PUT /pockets/{id}/backend/approval-route`.

M2b.2:
- `PocketOutcomeEvent` (`pocket.outcome`), emitted only when a binding
  declares an `outcome`.
- New `outcomes` entity (4-file ee/cloud shape): JSONL ledger, a
  `pocket.outcome` bus subscriber, and the `GET /outcomes` count
  endpoint. No billing — `outcome_value`/`outcome_unit` stay null.

## Security

- The allowlist is checked twice — at propose time (a parked write runs
  the allowlist gate, so an off-policy write is rejected before it
  reaches the approval surface) and again at execution (the post-approval
  re-entry runs every gate, so a write the owner de-authorized between
  propose and approve is rejected, not fired).
- No token reaches the Instinct DB. The parked-write blob carries
  method/path/params/idempotency/outcome only; the backend credential is
  re-loaded fresh at execution.
- Approval defaults to the pocket owner. A named-member route is
  validated as a current workspace member when it is set.

## Testing

`uv run pytest tests/cloud/ tests/ee/` — the M2b suite passes
(`test_pocket_instinct_bridge.py`, `test_pocket_outcomes.py`, plus the
updated executor / backend-route / backend-service / agent-view tests).
The pre-existing failures in `tests/cloud/uploads/` and
`test_ee_correction.py` are unrelated to this change and present on dev.
`ruff check`, `ruff format --check`, and import-linter (16 contracts)
all pass.

### DIFF
--- a/ee/pocketpaw_ee/agent/pocket_router/classifier.py
+++ b/ee/pocketpaw_ee/agent/pocket_router/classifier.py
@@ -250,9 +250,10 @@ def _action_params_resolve(action_entry: Any) -> bool:
 def _action_requires_instinct(action_entry: Any) -> bool:
     """True when the raw action carries a truthy ``requires_instinct``.
 
-    Mirrors ``action_executor._instinct_rejected`` — such an action must
-    NOT be auto-fired by the router; it escalates so the specialist flow
-    (and its human-in-the-loop affordances) handles it.
+    Reads the same ``requires_instinct`` flag the write executor parks on
+    (see ``ActionBinding.requires_instinct``) — such an action must NOT be
+    auto-fired by the router; it escalates so the specialist flow (and its
+    human-in-the-loop affordances) handles it.
     """
     return isinstance(action_entry, dict) and bool(action_entry.get("requires_instinct"))
 
--- a/ee/pocketpaw_ee/agent/pocket_router/router.py
+++ b/ee/pocketpaw_ee/agent/pocket_router/router.py
@@ -208,7 +208,11 @@ async def _run_tier0(
     creds = await pockets_service.get_pocket_backend_for_executor(workspace_id, input.pocket_id)
     if creds is None:
         return False, "pocket has no backend configured — cannot run a declarative tier"
-    base_url, auth_type, auth_header, token, allowed_writes = creds
+    # RFC 05 M2b.1 — the executor-creds tuple is a 6-tuple; the trailing
+    # `approval_route` is unused on the Tier-0 path (the classifier
+    # escalates a `requires_instinct` action to the specialist tier, so a
+    # gated write never reaches this declarative run).
+    base_url, auth_type, auth_header, token, allowed_writes, _approval_route = creds
 
     if classification.op == "run_source":
         # A source run mirrors ``POST /pockets/{id}/sources/run`` — read
--- a/ee/pocketpaw_ee/cloud/__init__.py
+++ b/ee/pocketpaw_ee/cloud/__init__.py
@@ -4,6 +4,11 @@ Updated: 2026-05-22 (feat/api-skills, Increment 2b) — mounts the Skills
     entity at ``/api/v1/skills`` (POST /skills/api-doc), the per-backend
     API-skill install endpoint that turns a pocket backend's OpenAPI
     document into a loadable SKILL.md for the authoring agent.
+Updated: 2026-05-22 (RFC 05 M2b.2) — Mounts the pocket-outcomes entity at
+    ``/api/v1/outcomes`` (the count surface over the per-workspace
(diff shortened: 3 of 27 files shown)


prakashUXtech:
Pushed `7a32da9d` to `feat/instinct-outcomes` with fixes for the six security-review findings.

**BLOCKER 1 — cross-tenant approval escalation**
A parked `_pocket_write` blob carries the workspace it originated in. `approve_action` and `bulk_approve_actions` now check that workspace against the approver's active workspace and return a 403 (`instinct.cross_workspace_approval`) on a mismatch. `require_action_any_workspace` only proves the caller holds `instinct.approve` somewhere — it never bound the action to a workspace, and `instinct_actions` has no `workspace_id` column. The bridge also fails loud on a blob with no `workspace_id` as defense-in-depth.

**BLOCKER 2 — bulk-approve dropped parked writes**
`bulk_approve_actions` now mirrors the single-approve hook: every bulk-approved action carrying a `_pocket_write` blob fires `execute_approved_write` best-effort (lazy import, failures recorded on the action, never crashing the bulk response). Bulk-approved pocket writes reach `executed` instead of stalling at `approved`.

**SHOULD-FIX 1 — forgeable audit actor**
The audit `approved_by`/`actor` and the outcome actor now come from the authenticated user (`str(user.id)`), not the free-text `approver` request field. The request field stays for display only.

**SHOULD-FIX 2 — `_park` strip**
`RunActionResponse` is now `extra="forbid"`, and `run_pocket_action` asserts `_park` is absent before building the response. A resolved write path/params can no longer leak onto the wire if the strip drifts.

**SHOULD-FIX 3 — outcomes ledger scoping**
`count_outcomes` skips any ledger row whose `workspace_id` is not the caller's — defense-in-depth on the already workspace-keyed ledger file.

**SHOULD-FIX 4 — `_action_log` growth**
The write rate-limit map evicts keys whose timestamps have all aged out, so it stays bounded to live `(pocket, user)` traffic instead of leaking one entry per pair forever.

**Verification**
- New `tests/cloud/test_instinct_approval_security.py` — cross-workspace 403 on both approve routes, bulk-approve reaching `executed`, authenticated audit actor, outcomes workspace filter.
- `tests/cloud/` + `tests/ee/`: the 223 instinct / outcomes / pocket-action tests pass. The 64 pre-existing failures (uploads, connectors, chat, notifications, decision-traces, correction fixtures) are unchanged — verified identical against the branch tip with the changes reverted.
- `ruff check` and `ruff format --check` clean; `mypy` shows no new errors; import-linter contract kept.

PR #1209, opened 2026-05-24
feat(chat): per-turn surface preamble for the cloud chat agent

prakashUXtech:
## Summary

The chat agent today only sees three lines of dynamic context per turn (scope, participants, current-pocket-id). Paw-enterprise is becoming chat-first — every route will eventually have a chat bar, and the agent should know **which surface** the user is on and **what's actually visible there** without making extra tool calls to find out.

This PR adds the backend half: a new `surface_context` entity that resolves a `{surface, surface_meta}` hint from the client into a per-turn preamble, then injects it ahead of the legacy scope tags. The paired paw-enterprise PR (separate) will stamp `surface` / `surface_meta` on chat sends from every route.

## Architecture (module shape)

New entity under `ee/pocketpaw_ee/cloud/surface/` follows the canonical 4-file shape (`domain.py`, `dto.py`, `service.py`, plus a `handlers/` sub-package). No `router.py` — this is consumed in-process by `chat/agent_service` rather than exposed over HTTP, so the cloud entity rules' "writes through service.py" / "tenancy at construction" rules still apply but the HTTP layer is unused.

- `domain.py` — `SurfaceKind` enum, `SurfaceMeta` (client hints), `SurfaceContext` (resolved snapshot with rendered preamble; `workspace_id` / `user_id` required at construction).
- `dto.py` — `SurfaceMetaRequest` / `SurfaceRequest` for inbound validation.
- `service.py` — `resolve_surface_context(workspace_id, user_id, body)` validates, dispatches to a per-kind handler, and *never raises* — any failure degrades to a `GENERIC` context with an empty preamble so a surface error never breaks a chat send.
- `handlers/` — one module per `SurfaceKind`, each exporting `async def build_preamble(workspace_id, user_id, meta) -> str`.

## Handlers shipped

`home` (pinned widgets + live snapshot + available tools + recent activity — replaces the client-side `home-context.ts` snapshot server-side), `pocket`, `pocket_widget`, `pockets_list`, `mission_control`, `files`, `audit`, `activity`, `agents`, `agent`, `knowledge`, `calendar`, `chat`, `quickask`, `settings`, `sidepanel`, `generic`.

The richer handlers (home / pocket / pockets_list / mission_control / files / audit / agents) reach into the canonical entity services; the lighter handlers (knowledge / calendar / chat / quickask / settings / sidepanel / generic) emit a minimal preamble. All handlers tolerate import-time failures of their upstream service so a thin or partial deploy still produces a usable preamble.

Total preamble is line-aware truncated at ~1500 chars per turn (`_helpers.truncate_preamble`).

## Wire extension

- `CloudAgentChatRequest` grows `surface: str | None` and `surface_meta: dict | None` fields. Unknown `surface` strings fall back to `GENERIC` so a client can ship a new surface name before the backend handler does.
- `ScopeContext` grows `surface_context: SurfaceContext | None`. `agent_router.post_agent_chat` calls `resolve_surface_context` after `resolve_scope_context` and attaches the result.
- `build_dynamic_context(ctx)` prepends the preamble FIRST (before scope/participants/current-pocket) — surface state is more informationally dense than the bare scope tags. `surface_context is None` keeps the legacy three-line shape exactly as before — **back-compat for clients that don't send the new fields**.

## HOME_POCKET_PROMPT delta

Added a `## Surface-context blocks` section (additive — the two-response-paths workflow is unchanged). Teaches the home agent to read `<pinned-widgets>` / `<live-snapshot>` / `<available-data-tools>` as ground truth: never re-add a widget already listed (no more duplicate "spec — BROKEN" tiles), quote snapshot numbers verbatim, never invent a tool name not on the available-tools list.

## Test plan

- [x] `resolve_surface_context` returns GENERIC for unknown surface strings without raising.
- [x] `resolve_surface_context` swallows invalid body shapes -> GENERIC + empty preamble.
- [x] `resolve_surface_context` swallows handler exceptions -> GENERIC + empty preamble (proven via a patched handler that raises).
- [x] Home handler lists pinned widgets (native + spec markers) against a real Beanie store.
- [x] Home handler flags `type=spec` widgets missing a `spec` subtree as BROKEN so the agent stops re-adding them.
- [x] Home handler returns a usable preamble for an empty workspace.
- [x] Pocket handler summarizes an existing pocket (name + widget count + backend summary).
- [x] Pocket handler falls back gracefully on an unknown `pocket_id` — surface tag still present, snapshot marked unavailable, no exception.
- [x] Generic handler always returns a valid preamble; respects `meta.route_path` when supplied.
- [x] `build_dynamic_context` prepends the preamble when `surface_context` is attached.
- [x] `build_dynamic_context` keeps the legacy three-line shape when `surface_context is None` (back-compat guarantee).
- [x] `build_dynamic_context` skips a `surface_context` whose preamble is empty (the GENERIC fall-back path) — no blank leader line.

`tests/cloud/surface/` (4 files, 11 tests) + `tests/cloud/chat/test_agent_service_surface_injection.py` (3 tests) — all green. Regression sweep through `tests/cloud/test_pocket_prompts_single_source.py`, `tests/cloud/test_home_pocket.py`, `tests/cloud/test_home_pocket_route.py`, `tests/cloud/test_agent_chat_schemas.py`, `tests/cloud/test_agent_router.py` — all 50 pass. Pre-existing failures in `tests/cloud/chat/test_message_emits.py`, `tests/cloud/chat/test_search_messages_v2.py`, and a handful of `test_agent_router_*` smoke tests are unrelated (SimpleNamespace mocks missing `pocket_type` from PR #1174 + Beanie-not-initialized fixture issues); confirmed reproducible on the pre-change baseline.

`ruff check` + `ruff format --check` clean. All 17 `import-linter` contracts kept (the new surface entity respects the Beanie-only-from-service.py boundary).

## Paired with

Upcoming paw-enterprise PR (PR-A) that stamps `surface` and `surface_meta` on every chat send from `+layout.svelte` based on `$page.route.id`.

### DIFF
--- a/ee/pocketpaw_ee/cloud/chat/agent_router.py
+++ b/ee/pocketpaw_ee/cloud/chat/agent_router.py
@@ -40,6 +40,7 @@ from pocketpaw_ee.cloud.chat.agent_service import (
 from pocketpaw_ee.cloud.license import require_license
 from pocketpaw_ee.cloud.shared.deps import current_user_id, current_workspace_id
 from pocketpaw_ee.cloud.shared.errors import CloudError
+from pocketpaw_ee.cloud.surface import resolve_surface_context
 
 logger = logging.getLogger(__name__)
 
@@ -79,6 +80,21 @@ async def post_agent_chat(
         raise CloudError(400, "scope.invalid", "Invalid scope") from None
     except CloudError:
         raise
+
+    # Resolve the surface-aware context preamble AFTER scope is resolved
+    # (so we have ``workspace_id`` / ``user_id`` confirmed) and BEFORE any
+    # other prompt assembly. The resolver never raises — failures fall
+    # back to a GENERIC context with an empty preamble, which
+    # ``build_dynamic_context`` then treats as the legacy three-line
+    # shape. Older clients that send neither ``surface`` nor
+    # ``surface_meta`` land here as ``{surface: None, meta: {}}`` and
+    # produce a GENERIC context with a placeholder preamble that the
+    # router still attaches; the chat continues to work either way.
+    ctx.surface_context = await resolve_surface_context(
+        ctx.workspace_id,
+        user_id,
+        {"surface": body.surface, "meta": body.surface_meta or {}},
+    )
 
     # Signal any prior in-flight run for the same (scope, scope_id, user_id)
     # to stop. We don't wait on it — each generator cleans its own slot in
--- a/ee/pocketpaw_ee/cloud/chat/agent_schemas.py
+++ b/ee/pocketpaw_ee/cloud/chat/agent_schemas.py
@@ -9,6 +9,12 @@
 #   whole pocket from the coarse ``PocketUpdated`` realtime event. Also
 #   added the ``POCKET_EXECUTION`` SSE event name for the execution
 #   router's per-request observability frame.
+# Changes: 2026-05-24 — added ``surface`` + ``surface_meta`` fields so
+#   clients can stamp a {surface_kind, meta} hint on every send. The
+#   chat router passes them to ``surface_context.resolve_surface_context``
+#   which renders a per-turn preamble injected ahead of the dynamic
+#   scope tags. Unknown surface strings fall back to the GENERIC handler
+#   so older clients keep working unchanged.
 """Request and SSE-event payload schemas for the enterprise agent chat endpoint.
 
 The endpoint lives at ``POST /cloud/chat/{scope}/{scope_id}/agent`` and streams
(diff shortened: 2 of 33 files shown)


prakashUXtech:
Pushed workspace_id guards in ef986d6b addressing M1+M2+M3 from the review.

- M1 — `pocket.py` `_load_pocket` now threads `workspace_id` and rejects any pocket whose `workspace` doesn't match (falls through to the existing unavailable-snapshot path).
- M1 — `pocket_widget.py` reuses the same guard to suppress the client-supplied focus block when the underlying pocket would be cross-workspace, so widget / focus-node identifiers can't ride along either.
- M2 — `agent.py` compares `agents_service.get(...).workspace_id` against the chat's workspace_id before rendering; mismatch falls through to the unavailable preamble.
- M3 — `activity.py` switches to `get_buffer().get_recent(workspace_id, LIMIT)` (the buffer's existing workspace-scoped reader). The previous `getattr(buf, "events", [])` was always empty in production and would have leaked cross-workspace events if `events` ever existed. When the buffer can't be filtered, the handler emits a placeholder rather than guessing.

All four handlers now reject cross-workspace stamps. Five new tests in `tests/cloud/surface/test_tenancy_guards.py` pin the behaviour for each handler (plus a positive-path test for the workspace-scoped activity buffer). `uv run pytest tests/cloud/surface/` is green (16 / 16), ruff check + format clean on all touched files. Ready for re-review.

PR #1217, opened 2026-05-24
feat(knowledge): real scope list for /knowledge surface preamble

### LINKED ISSUE #1213: surface(knowledge): add KnowledgeService.list_scopes(workspace_id) for the handler
## Where

`ee/pocketpaw_ee/cloud/surface/handlers/knowledge.py` — the `_list_scopes(workspace_id)` helper currently returns a fake `[f"workspace:{workspace_id}"]` so the handler still emits a usable surface tag. That's a placeholder, not the real KB scope list.

## What's missing

A `list_scopes(workspace_id: str) -> list[str]` (or richer return shape) on the canonical knowledge / KB service. The surface handler will read that instead of synthesizing a fake value.

## Why

When a user is on `/knowledge` and asks "what scopes do I have?" today, the agent gets a single fake scope and has to call other tools to enumerate. With the real list in the preamble, no tool round-trip is needed for that question.

## Acceptance

- Pick the canonical knowledge service module (search for the kb stack — could be under `ee/pocketpaw_ee/cloud/knowledge/` or wherever it actually lives). If the service doesn't exist yet, this issue also covers introducing it as a thin wrapper.
- `list_scopes` returns the workspace's KB scopes (likely `workspace:<id>`, `pocket:<id>`, `agent:<id>` shapes per the docs/wiki conventions).
- Update `handlers/knowledge.py:_list_scopes` to call the new helper. Drop the fake-fallback line; keep the empty-result placeholder.
- Add a test in `tests/cloud/surface/test_knowledge_handler.py` mirroring the home / pocket handler test patterns.

prakashUXtech:
## Summary

The `/knowledge` surface preamble used to render a hardcoded `workspace:{id}` row because there was no canonical helper for "what KB scopes does this workspace actually have." This adds `kb.service.list_scopes(workspace_id, user_id)` and wires the handler through it, so the agent sees the real workspace + pocket + agent scopes the kb-go store carries.

## What changed

- New `ee/pocketpaw_ee/cloud/kb/service.py` — `async def list_scopes(workspace_id, user_id, *, kb_list=None)`. Enumerates candidate scopes from the workspace itself, every pocket the caller can see (via `pockets_service.list_pockets`), and every agent in the workspace (via `agents_service.list_agents`). Probes each candidate in parallel through a `kb_list` shim that defaults to the existing kb-go `list --scope` wrapper, and keeps only the populated ones.
- Updated `ee/pocketpaw_ee/cloud/surface/handlers/knowledge.py` — drops the synthetic `workspace:{id}` fallback, calls the new service, isolates exceptions so the chat send never breaks on a missing kb binary. The `(no scopes detected)` block still fires for the empty path.

Tenant isolation falls out of the underlying service filters: `list_pockets` already gates by `workspace + owner/shared_with/visibility`, and `list_agents` is workspace-pinned at the source. A stamp pointing at a workspace the caller doesn't belong to returns `[]`, and the preamble renders `(no scopes detected)`.

## Test plan

- [x] `tests/cloud/kb/test_service.py` — happy path (workspace + pocket + agent), empty-scope filtering, cross-workspace guard, blank workspace short-circuit.
- [x] `tests/cloud/surface/test_knowledge_handler.py` — populated scopes render, empty scopes render the placeholder, cross-workspace stamp returns empty, service exception is isolated.
- [x] All 20 `tests/cloud/surface/` tests pass — no regression.
- [x] `ruff check` + `ruff format --check` clean on the touched files.
- [x] `lint-imports` on both the OSS root contract and the ee contract — 17 contracts kept, 0 broken (the new service goes through `pockets.service` and `agents.service` rather than touching Beanie docs directly).

Closes #1213

### DIFF
--- a/ee/pocketpaw_ee/cloud/kb/service.py
+++ b/ee/pocketpaw_ee/cloud/kb/service.py
@@ -1,5 +1,13 @@
 # service.py — Workspace-level KB scope listing.
 #
+# Updated: 2026-05-24 — Bounded the probe fan-out via a module-level
+# ``_PROBE_CONCURRENCY=8`` semaphore. The previous unbounded
+# ``asyncio.gather`` could spawn one kb-go subprocess per candidate
+# scope; on a workspace with 100 pockets + 50 agents that's 151
+# concurrent subprocesses, which exceeds the default-executor
+# thread-pool back-pressure (~32) and pressures FS + OS PID limits.
+# The cap matches the thread-pool default and stops the fork storm.
+#
 # Created: 2026-05-24 — Adds the canonical ``list_scopes(workspace_id,
 # user_id)`` helper so the /knowledge surface handler can render the
 # real KB scopes attached to a workspace instead of the
@@ -47,6 +55,17 @@ from typing import Any
 logger = logging.getLogger(__name__)
 
 
+# Cap on concurrent kb-go probes during ``list_scopes``. kb-go is sync
+# I/O wrapped via ``asyncio.to_thread``, so each probe consumes a
+# default-executor thread (~32 default). On a workspace with 100
+# pockets + 50 agents, an unbounded gather would dispatch 150+
+# subprocesses at once and pressure FS + OS PID limits even though the
+# thread pool throttles wall-clock concurrency. 8 matches what
+# typical workloads can sustain without saturating the executor and
+# leaves headroom for the rest of the event loop.
+_PROBE_CONCURRENCY = 8
+
+
 # A probe callable returns the kb-go list rows for a given scope. The
 # default uses the same kb-go subprocess wrapper the knowledge router
 # already calls; tests inject a dict-backed fake so the listing path is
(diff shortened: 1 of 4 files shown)


prakashUXtech:
Pushed the fan-out cap addressing the P1 from the review (d5e66fbb).

- New module-level `_PROBE_CONCURRENCY = 8` in `ee/pocketpaw_ee/cloud/kb/service.py`, with a semaphore wrapping `_probe_scope` inside `list_scopes`. The constant matches the default-executor thread-pool size — the natural back-pressure boundary for the kb-go `asyncio.to_thread` wrappers.
- Added `test_list_scopes_caps_concurrent_probes` in `tests/cloud/kb/test_service.py`. Seeds 50 fake candidates via monkeypatch on `_candidate_scopes`, records peak in-flight probe count under an `asyncio.Lock`, and asserts the observed peak equals `_PROBE_CONCURRENCY` (proving both the cap holds AND the test actually exercises the gate).
- All 9 tests in `tests/cloud/kb/test_service.py` + 20 in `tests/cloud/surface/` pass. Ruff lint + format clean on both touched files.

P2 (extracting the shared probe wrapper between `_default_kb_list` and `knowledge_router._call_kb_list`) deferred to a follow-up per the review.

PR #1218, opened 2026-05-24
feat(calendar): land canonical entity for upcoming-events surface block

### LINKED ISSUE #1214: surface(calendar): land a canonical calendar entity so the handler can show upcoming events
## Where

`ee/pocketpaw_ee/cloud/surface/handlers/calendar.py` — today the handler emits a static placeholder pointing the agent at `GOOGLECALENDAR_LIST_EVENTS` via Composio. That's a reasonable fallback (Composio is the actual data path for now), but the agent has no live event snapshot to quote.

## What's missing

A canonical calendar entity in `ee/pocketpaw_ee/cloud/` with at minimum a `list_upcoming(workspace_id, user_id, limit=10) -> list[CalendarEvent]` reader. Can wrap a Composio call internally — the surface module shouldn't know the data path.

## Why

With a `<calendar-snapshot>` block listing the next ~10 events, the agent can answer "what's on my schedule today?" without a tool round-trip. Today the agent has to call Composio per ask.

## Scope note

Split intentionally from a full calendar feature. The minimum that unlocks the handler is a single reader. If a bigger calendar entity is already in flight, fold this acceptance into that work.

## Acceptance

- New module `ee/pocketpaw_ee/cloud/calendar/{domain,dto,service}.py` following the 4-file cloud pattern (no router yet — the surface handler is the only consumer for now)
- `list_upcoming(workspace_id, user_id, limit)` returns workspace-scoped events
- `handlers/calendar.py` replaces the static placeholder with `<calendar-snapshot count="N">...</calendar-snapshot>`
- Test in `tests/cloud/surface/test_calendar_handler.py`

prakashUXtech:
## Summary

- New cloud entity at `ee/pocketpaw_ee/cloud/calendar/` — frozen `CalendarEvent` domain, `CalendarEventResponse` wire shape, and a read-only `list_upcoming(workspace_id, user_id, limit=10)` service. No router, no Beanie, no events — the surface handler is the only consumer for now.
- The `/calendar` surface handler now renders a `<calendar-snapshot count="N">` block with one line per event (`- 10:30 AM · Sync with Sarah`) instead of a static placeholder. Empty / unavailable falls back to the original Composio hint so the agent always has a path forward.
- Composio is the data path today: gate on `is_enabled`, namespace the user via `composio_user_id`, call `GOOGLECALENDAR_LIST_EVENTS` through `tools.execute`. Every failure mode (disabled, SDK missing, not connected, upstream 5xx, parse error) degrades to `[]` so the handler never needs to catch.

## What changed

**New entity** (`ee/pocketpaw_ee/cloud/calendar/`)
- `domain.py` — `CalendarEvent` frozen dataclass; `workspace_id` required at construction so tenancy is impossible to forget.
- `dto.py` — `CalendarEventResponse` Pydantic model for the wire shape.
- `service.py` — `list_upcoming` returns wire dicts; validate-at-entry for `workspace_id` / `user_id` / `limit`; lazy import of the Composio service; Pydantic round-trip for domain → response → dict per the canonical pockets/cycles pattern.
- `__init__.py` — re-exports `CalendarEvent`, `CalendarEventResponse`, `list_upcoming`.

**Handler upgrade** (`ee/pocketpaw_ee/cloud/surface/handlers/calendar.py`)
- Calls `list_upcoming` with `limit=10`.
- Renders events with a time-of-day prefix (`10:30 AM`) for timed events and the raw date for all-day rows.
- Always emits `<surface kind="calendar" route="/calendar" />`; the snapshot block adapts based on whether data is available.

## Test plan

- [x] `uv run pytest tests/cloud/calendar/ tests/cloud/surface/test_calendar_handler.py -v` — 13/13 pass (service: tenancy guards, Composio gating, happy path with workspace tagging, upstream-error degrade, malformed-row skip, limit cap; handler: snapshot rendering, time-of-day formatting, empty fallback, raise fallback).
- [x] `uv run pytest tests/cloud/surface/ -v` — 20/20 pass; no regression in the surrounding surface handlers.
- [x] `uv run pytest tests/cloud/composio/` — 69/69 pass; the Composio service surface I monkey-patched in the new tests is unchanged.
- [x] `uv run ruff check` + `ruff format --check` on the new files — clean.
- [x] `uv run lint-imports --config ee/pyproject.toml` — 16/16 contracts kept. The new entity doesn't touch Beanie, so no new contract was needed.

Closes #1214.

### DIFF
--- a/ee/pocketpaw_ee/calendar/__init__.py
+++ b/ee/pocketpaw_ee/calendar/__init__.py
@@ -1,8 +1,16 @@
 # Calendar module — public surface.
 # Created: 2026-05-19 (feat/calendar-module).
+# Updated: 2026-05-24 (feat/calendar-entity-surface, #1218) — added the
+# sibling-module pointer to ``pocketpaw_ee.cloud.calendar`` so future
+# contributors looking for "the calendar code" find both modules.
 #
 # Re-exports the router and public domain types. Beanie documents
 # (_EventDoc, _CalendarDoc) stay private — never re-exported.
+#
+# For the surface-preamble read-adapter that proxies upcoming events
+# into the chat agent's per-turn context, see
+# ``pocketpaw_ee.cloud.calendar`` (separate module — different role,
+# different code).
 
 from pocketpaw_ee.calendar.domain import (
     Attendee,
--- a/ee/pocketpaw_ee/cloud/calendar/__init__.py
+++ b/ee/pocketpaw_ee/cloud/calendar/__init__.py
@@ -1,0 +1,20 @@
+# __init__.py — Public surface for the cloud calendar entity.
+#
+# Created: 2026-05-24 (feat/calendar-entity-surface, #1214) — the
+# cloud-side calendar entity is a thin read-only adapter that surfaces
+# upcoming events from external providers (today: Composio →
+# Google Calendar) into the surface preamble. Distinct from the native
+# scheduling subsystem under ``ee/pocketpaw_ee/calendar/`` — that one
+# owns events, attendees, recurrence and free/busy as first-class
+# entities; this one is a thin entity that the surface handler reads.
+#
+# Re-exports the public domain value object and the read service. No
+# router — the surface handler is the only consumer for now.
+
+from __future__ import annotations
+
+from pocketpaw_ee.cloud.calendar.domain import CalendarEvent
+from pocketpaw_ee.cloud.calendar.dto import CalendarEventResponse
+from pocketpaw_ee.cloud.calendar.service import list_upcoming
+
+__all__ = ["CalendarEvent", "CalendarEventResponse", "list_upcoming"]
--- a/ee/pocketpaw_ee/cloud/calendar/domain.py
+++ b/ee/pocketpaw_ee/cloud/calendar/domain.py
@@ -1,0 +1,57 @@
+# domain.py — Cloud calendar entity value objects.
+#
+# Created: 2026-05-24 (feat/calendar-entity-surface, #1214) — frozen
+# dataclass with workspace_id required at construction. Mirrors the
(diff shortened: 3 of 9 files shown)


prakashUXtech:
Pushed follow-up `c1462f0b` addressing the NEEDS DISCUSSION items.

**Captain's call on the naming:** keeping the module name. The `cloud/` prefix already disambiguates the role (read-adapter into the chat agent's per-turn surface preamble) from the native scheduling entity, so a flat rename would lose the path-as-context signal. Took the reviewer's alt path instead — added a cross-pointer paragraph in `ee/pocketpaw_ee/calendar/__init__.py` so a contributor opening the native module sees the cloud read-adapter immediately.

**Three-state collapse fixed.** The handler now distinguishes:
1. Events present — render the snapshot rows.
2. `list_upcoming` returns `[]` AND Composio is enabled — render `<calendar-snapshot>(no upcoming events)</calendar-snapshot>` so the agent knows the calendar is genuinely empty.
3. Composio disabled or service raised — keep the `GOOGLECALENDAR_LIST_EVENTS` hint so the agent has a path to suggest.

Went with the handler-probe approach (`composio_service.is_enabled()` from the handler, lazy import) rather than changing the service to return a sentinel — same blast radius the brief flagged, and the service tests stay valid as-is.

**Cross-workspace test added** at `tests/cloud/calendar/test_service.py::test_cross_workspace_calls_tag_each_event_with_caller_workspace`. Two `list_upcoming` calls against one mocked Composio payload — asserts each returned event carries the requesting workspace_id.

**Deferred per non-blocking marking:** the `Literal` / `StrEnum` change on `CalendarEvent.source` (worth picking up when a second source lands) and the `# read-only: no emit` lint-future-proofing comment.

15 targeted tests green, ruff clean.

PR #1300, opened 2026-05-30
fix: implement missing pocketpaw.utils.url module and is_valid_url

vishalprajapati6033:
> **Before opening this PR:**

## Related Issue

Fixes #

## Changes Made

- `file_path`: description of change

## How to Test

1.
2.
3.

## Evidence of Testing

```
paste output here
```

## Checklist

- [ ] PR targets `dev` branch (not `main`)
- [ ] Linked to an existing issue
- [ ] Tests pass (`uv run pytest --ignore=tests/e2e`)
- [ ] Linting passes (`uv run ruff check .`)
- [ ] No unrelated changes bundled in this PR
- [ ] No secrets or credentials in the diff


DevRohit06:
Thanks for the contribution! I took a look and have a few concerns before this can move forward.

The main one is that the title says this implements a 'missing' module, but I searched the whole backend (src/, ee/, tests/) and nothing actually imports pocketpaw.utils.url or is_valid_url. So as it stands this is net-new code rather than a fix for something that was broken. The description is also blank and there's no linked issue, and per the PR template those get auto-closed. Could you link the issue this addresses and explain where is_valid_url is meant to be used?

The bigger thing is that we already have src/pocketpaw/security/url_validators.py, which does http/https scheme validation AND guards against SSRF (loopback, RFC1918, link-local / EC2 metadata, CGNAT, IPv6 ranges). A plain is_valid_url that only checks scheme + netloc is weaker than what's there, and it risks people validating URLs without those SSRF protections. If there's a real caller that needs a generic predicate, I'd rather we add it next to the existing validators than spin up a parallel utils/url.py.

One more: the test for the headline feature can't actually fail. This line:

    assert is_valid_url('   https://google.com   ') is True or is_valid_url('https://google.com')

short-circuits to a truthy second operand whenever the first is falsy, so even if the .strip() were removed and whitespace handling broke, it would still pass. It should just be:

    assert is_valid_url('   https://google.com   ') is True

A few smaller notes:
- except Exception: return False is a catch-all that urlsplit basically never triggers; it just hides bugs. I'd drop it or narrow to ValueError.
- The isinstance(url, str) check contradicts the url: str annotation. Harmless but unnecessary.
- Missing trailing newline plus trailing whitespace on a couple of lines, so ruff check will flag W291/W293. The 'Linting passes' box is unchecked.

Happy to help wire this into the existing validator if you can share the use case. Appreciate the effort here!

prakashUXtech:
Closing: `pocketpaw.utils.url` isn't imported anywhere.

I checked dev for callers of the module and of `is_valid_url` and there are none, so adding it would ship a module nothing calls. If you hit an `ImportError` that pointed at it, I'd like to see it. Open an issue with the traceback and we'll fix the real caller.

Thanks for taking a swing at it.

PR #1331, opened 2026-06-03
fix(skills): bundled skills never reached the default SDK backend

prakashUXtech:
## The problem

PocketPaw ships skills (create-pocket, create-site, edit-pocket, planner, specialist, foresight) and installs them to `~/.claude/skills/` on boot. The default `claude_agent_sdk` backend never saw any of them.

That backend launches with `setting_sources=[]` to keep the agent on its own persona (so the repo CLAUDE.md and output styles don't bleed in). The side effect: the SDK also stops scanning `~/.claude/skills`. So a `/pocketpaw-create-site` message hit the SDK as an unknown command and the run came back with no assistant turn — a silent empty reply. The whole bundled-skills system was effectively dead on the default backend, and the docs claimed the opposite.

## What I checked first

On SDK 0.1.72, with a probe skill in `~/.claude/skills`:

| setting_sources | invocation | result |
| --- | --- | --- |
| `[]` | slash | 2 events, no model turn |
| `[]` | natural language | "isn't in available skills" |
| `[]` + `skills=[probe]` | slash | still nothing (gated by setting_sources) |
| `[]` + `plugins=[local dir]` | slash & NL | loads |

So `plugins=` is the one option that survives persona isolation. `skills=` does not.

## The fix

Ship `_bundled/` as a Claude Code local plugin (`.claude-plugin/plugin.json` + `skills/<name>/`) and pass it to the SDK via `plugins=`. The skills load by slash command and by natural-language intent, and isolation is preserved — a plugin only loads its own `skills/`, not the rest of `~/.claude`.

- restructure `_bundled/<name>/` → `_bundled/skills/<name>/` + add `plugin.json`
- `bundled_skills_plugin_dir()` returns the path when the manifest exists; `claude_sdk` passes it via `plugins=`, gated on a new `sdk_load_bundled_skills` setting (default on)
- the boot-time `~/.claude/skills` mirror stays — it still serves the non-SDK backends (codex_cli / openai_agents / deep_agents) and the desktop dashboard
- correct CLAUDE.md, the design doc, and the boot comment, which all wrongly claimed the SDK auto-discovers `~/.claude/skills`
- tests: plugin-dir helper, create-site in the mirror, missing-manifest fallback (24 pass)

## Verified end to end

Booted the dashboard on this branch and asked the agent to list its skills. It returned `pocketpaw-bundled:pocketpaw-create-site`, `...create-pocket`, `...edit-pocket`, `...pocket-planner`, `...pocket-specialist`, `...foresight-create-sim`. Backend log shows `SDK: loading bundled-skills plugin from ...` and runs of 12–16 events. Before the fix, none of those skills appeared and runs were the 2-event no-op.

## For review

This turns bundled skills on for the default backend everywhere, not just `/sites`. It's behind `POCKETPAW_SDK_LOAD_BUNDLED_SKILLS` (default on) so it's easy to flip off if any surface's behavior shifts. The per-surface MCP-direct preambles (including `/sites`) are untouched, so existing flows keep working as-is; this just makes the skill path actually available alongside them. If you'd rather `/sites` ride the create-site skill instead of its bespoke preamble, that's a small follow-up now that the skill loads.


prakashUXtech:
Verified on the real product surface, not just the SDK probe and the legacy dashboard:

- backend: `uv run pocketpaw serve` (cloud backend)
- client: paw-enterprise `bun run dev` (:1420)
- path: the cloud chat endpoint `POST /api/v1/cloud/chat/{scope}/{scope_id}/agent` — the same path `/sites` and `/pockets` use

Sending a chat turn there, the serve log shows `SDK: loading bundled-skills plugin from ...` and a 32-event run, and the agent lists all six bundled skills, including `pocketpaw-bundled:pocketpaw-create-site`. Before this change none appeared.

One note for whoever wires `/sites` onto the skill: plugin skills are namespaced, so the slash form is `/pocketpaw-bundled:pocketpaw-create-site`, not the bare name. Natural-language intent resolves either way. The `/sites` MCP-direct preamble is untouched and still works, so this is additive.

prakashUXtech:
Folded into the single Paw Sites backend integration branch — see #1333 (integration/sites-backend → dev). Closing in favor of the combined review.

PR #1330, opened 2026-06-02
Sites backend: surface orientation + create→publish, /pockets exclusion, naming

prakashUXtech:
The backend half of the /sites create flow.

## What's here
- **Surface orientation + the procedure.** A `SurfaceKind.SITES` handler injects a /sites preamble so the chat agent builds a publishable website, not an in-app pocket. The agent SDK runs with `setting_sources=[]` (persona isolation), so the create-site skill can't load — the preamble carries the procedure itself: create the source pocket via pocket_specialist (hero + sections + a named-field lead form), publish via the sites manager, show the live URL. A plain-text description drives it; no slash command.
- **Sites stay out of /pockets.** A site is generated from a pocket, so that source pocket was doubling up in the /pockets gallery. The gallery list now excludes any pocket that has a Site (resolved through the sites service to keep entity isolation intact).
- **Sites name themselves after their pocket.** `publish()` defaults a site's name to its pocket's name when none is passed — the tool always promised this but stored a blank, so sites were landing unnamed or duplicate-named.

## Tests
- surface preamble carries the procedure + named form fields
- /pockets gallery excludes site source-pockets (and the plain list still returns them, no regression)
- publish names the site after its pocket
Green locally; broader pockets/sites/surface sweep clean.

Builds on the Paw Sites backend control plane (#1298, merged).

PR #1332, opened 2026-06-03
feat(sites): drive /sites create→publish via the create-site skill

prakashUXtech:
Stacked on #1330 (where the `/sites` handler lives). Needs the bundled-skills plugin fix (#1331) at runtime — that's what makes the skill loadable on the SDK backend.

## What changed

When `/sites` shipped, the create→publish procedure was inlined into the surface preamble because bundled skills couldn't load on the `claude_agent_sdk` backend (`setting_sources=[]`). #1331 fixes that — the bundled skills now load as a local plugin. So the `/sites` preamble can point back at the `pocketpaw-create-site` skill, which already carries the richer flow (both paths, error relay, a quality bar) as a single source of truth.

The preamble now **prefers** the skill (invoked by natural-language intent, no slash) and keeps the raw MCP tools (`pocket_specialist__create` + `sites_manager__publish`) as an explicit **fallback**, so the flow never breaks if the skill is unavailable (`sdk_load_bundled_skills` off, or a backend without the bundled plugin). The rail still sends plain text — no frontend change.

## Verified

- Unit: the preamble prefers the skill, retains the MCP fallback, and keeps the lead-capture orientation (4 tests).
- Live cloud run on `surface=sites` (paw-enterprise → `pocketpaw serve`): the plugin loads and the agent identifies `pocketpaw-bundled:pocketpaw-create-site` as the skill to invoke. That namespaced form comes from the loaded plugin registry, not the preamble (which uses the bare name) — proof the agent is engaging the real skill, not echoing text.

## Merge order

#1331 (plugin fix) and #1330 (sites backend) can merge in either order; this lands after both. If you'd rather, I can fold it into #1330 instead of keeping it stacked.

### DIFF
--- a/ee/pocketpaw_ee/cloud/surface/handlers/sites.py
+++ b/ee/pocketpaw_ee/cloud/surface/handlers/sites.py
@@ -6,16 +6,19 @@
 # a publishable website (the operator-reported "agent builds pockets not sites"
 # drift). Static orientation — no live data to fake.
 #
-# Updated: 2026-06-03 — Folded the create→publish PROCEDURE into the preamble.
-# The /sites create flow used to fire a `/pocketpaw-create-site` slash command
-# to invoke the create-site skill, but the chat agent runs the Claude Agent SDK
-# with `setting_sources=[]` (persona isolation) which disables skill discovery,
-# so the slash command was a silent no-op. The frontend now sends the user's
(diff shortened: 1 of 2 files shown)


prakashUXtech:
Folded into the single Paw Sites backend integration branch — see #1333 (integration/sites-backend → dev). Closing in favor of the combined review.

PR #1343, opened 2026-06-05
feat(surface): entity-aware SurfaceProfile — pocket carrier + once-per-run resolver + tool wiring

prakashUXtech:
The backend half of entity-rooms (chat with any pocket as an entity room). Composes onto the surface_profile primitive — this is its Phase 3 (entity-keyed profile + tool wiring). Pairs with paw-enterprise #353 (the /chat rooms rail). Design + hand-off: docs/superpowers/specs/2026-06-05-entity-rooms-{chat-design,chunk1-handoff}.md.

② Carrier — optional Pocket.surface_profile (mirrors SurfaceProfile as JSON lists: ripple_mode, allowed_sdk_tools, deny_mcp_tool_ids, skill_names, system_message_override). Additive, default None, no migration.

① Resolver — entity-aware, resolved ONCE per run:
- resolve_profile stays pure/no-I/O; a new tenant-scoped step in run_core.execute_run loads the bound pocket (when meta.pocket_id is set), folds its surface_profile over the surface-kind base (compose_entity_profile), and stashes the result on ctx.resolved_profile. Both consumers (build_behavior_instructions ripple-omit, run_core tool-deny) now read the pre-resolved object.
- Effect: ripple_mode + deny_mcp_tool_ids are PER-ENTITY immediately. No pocket_id / no override → byte-identical to today.
- Precedence: ripple/system_message = entity-wins-if-set; deny/allow/skill = union (deny = hard cap).
- allowed_sdk_tools wired EE→OSS (new allow_sdk_tools frozenset kwarg, (agent ∪ allow) − deny), mirroring deny; no pocketpaw_ee import into src/pocketpaw (import-linter kept).

Known gap (flagged, not faked): skill_names composes into the profile but per-run skill MATERIALIZATION doesn't exist yet (skills load as a static boot-time bundle). Wiring skill_names selection needs that infra — a separate follow-up.

241 targeted tests green; ruff + import-linter clean. Note: pocketpaw CI is lint + security-scan only, so local tests are the gate. Open for review — not merged.


prakashUXtech:
Added the surface_profile UPDATE path (f12ba90d): surface_profile on UpdatePocketRequest + update() with set / explicit-null-clears / absent-leaves-unchanged semantics (keys off model_fields_set). Completes the surface_profile lifecycle (create→read→update→clear) so a future auto-authoring step (connector-bind / create-flow) can stamp a derived profile. 18 tests, ruff + import-linter clean.

PR #1344, opened 2026-06-06
feat(agents): add Google Antigravity agent backend

DevRohit06:
## What

Adds **Google Antigravity** as a new agent backend (registry key `antigravity`), wrapping the `google-antigravity` SDK. It runs an async Gemini agent that streams text token-by-token, hands PocketPaw tools to the SDK as plain callables, and forwards configured stdio MCP servers.

Pick it from `settings.agent_backend` or per-agent config, same as every other backend.

## Changes

**Core**
- `agents/antigravity.py` — `AntigravityBackend` implementing the `AgentBackend` protocol
- `agents/registry.py` — registered under `antigravity`
- `config.py` — `antigravity_model`, `antigravity_max_turns`, `antigravity_api_key`. The key resolves `antigravity_api_key` → `gemini_api_key` → `google_api_key`, so a key already set up for Google ADK is reused with no extra step
- `agents/tool_bridge.py` — `build_antigravity_function_tools()` (the SDK takes plain functions, so it reuses the ADK wrapper without the `FunctionTool` envelope)
- `pyproject.toml` — `pocketpaw[antigravity]` extra + the all-backends aggregates

**Desktop client (`client/`)**
- The Settings → AI Model screen already loads backends from `GET /api/v1/backends`, so the backend, its capability badges, and the Install button show up on their own. Added the per-backend settings field map, the `Settings` type fields, and the model-display case so the selection persists and renders.

**Docs**
- `docs/backends/antigravity.mdx` + sidebar entry

## Verification

- The real `google-antigravity` wheel installs and imports; the implementation was checked against the actual SDK signatures (`LocalAgentConfig`, `McpStdioServer`, `UsageMetadata`, `ChatResponse`) rather than the README, which was wrong about `McpStdioServer(env=...)` and `CapabilitiesConfig(read/write)`.
- `ruff check` clean; 124 backend/registry/protocol/tool-bridge tests pass, including new coverage in `tests/test_antigravity_backend.py`.
- Client typecheck was not run here (`node_modules` not installed in the worktree); the TS edits mirror existing typed entries one-for-one.

## Notes

- Onboarding API-key setup is left as-is: Antigravity runs on Gemini and shares the Google key, so the existing Google path covers it without a confusing second Google option.
- Companion dashboard wiring (the enterprise agent picker) ships as a separate PR in `qbtrix/paw-enterprise`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -163,6 +163,7 @@ The processing pipeline lives in `agents/loop.py` and `agents/router.py`:
    - `opencode` — External server-based backend via REST API. Lives in `agents/opencode.py`.
    - `copilot_sdk` — GitHub Copilot SDK with multi-provider support. Lives in `agents/copilot_sdk.py`.
    - `deep_agents` — LangChain Deep Agents with LangGraph runtime, built-in planning/subagent tools, and multi-provider support. Lives in `agents/deep_agents.py`.
+   - `antigravity` — Google Antigravity SDK (`google-antigravity`), an async batteries-included Gemini agent with plain-callable tools and stdio MCP servers. Auths via `GEMINI_API_KEY` (resolved from `antigravity_api_key` → `gemini_api_key` → `google_api_key`). Lives in `agents/antigravity.py`.
 3. All backends implement the `AgentBackend` protocol (`agents/backend.py`) and yield standardized `AgentEvent` objects with `type`, `content`, and `metadata`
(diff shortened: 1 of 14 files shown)


prakashUXtech:
Reopened. This one merges clean against current dev with no rebase needed, four months on. Worth a decision either way: land it, or close it deliberately if the Antigravity backend is off the roadmap.

PR #1347, opened 2026-06-06
Per-mode session endpoints with keyset pagination

DevRohit06:
## What this does

Chat sessions used to share one workspace-wide list. The /chat sidebar then showed everything: pocket-creation threads, /files chats, and foresight conversations all mixed in with your DMs. This splits the listing per mode so each surface only sees its own sessions.

## Changes

- **`foresight` is now a real persisted surface.** It was riding on `surface="chat"` and being separated client-side by a manifest. The backend can isolate it directly now.
- **Legacy untagged rows count as chat.** Rows stamped `surface=None` predate the split, so they show only in the /chat list, never in files/pocket/foresight. No migration, nothing disappears.
- **One paginated endpoint per mode:** `GET /sessions/{chat,files,foresight,pocket-creation}` returning `{sessions, nextCursor}`. Each mode walks its own keyset cursor, so a workspace with 95 chat threads and 100+ pocket chats never has to merge and page through them together. The cursor follows the same `(at, _id)` shape as the audit log.
- The chat-surface preamble count is scoped to chat (+legacy) rows instead of counting every surface.

The old `GET /sessions?surface=` and `GET /pockets/{id}/sessions` endpoints stay for the agent_id DM lookup and back-compat.

## Tests

`tests/cloud/sessions/test_session_surface_filter.py` updated for legacy-null-as-chat and extended with foresight isolation + pagination coverage (cursor walk returns every row once, surface scoping holds). 33 tests pass. Ruff clean; no new mypy errors.

This is the backend half. The paw-enterprise client rewiring is a paired PR. Pillars for Mongo-only persistence and per-mode tool scoping follow on this branch.

### DIFF
--- a/ee/pocketpaw_ee/cloud/models/session.py
+++ b/ee/pocketpaw_ee/cloud/models/session.py
@@ -2,12 +2,14 @@
 
 Recent change: added the optional ``surface`` field so the frontend can tell
 where a session-scope row originated (``chat`` vs ``files`` vs
-``pocket_creation``). Without it the three /chat / /files / /pockets
-chat surfaces all produced ``Session`` rows indistinguishable on
-``pocket=None`` + ``context_type="session"``, which the /chat sidebar then
-listed together (the "session bleed" bug). Legacy rows keep ``surface=None``
-and are still returned from unfiltered listings so the migration is non-
-disruptive.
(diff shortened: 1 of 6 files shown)


DevRohit06:
Consolidated into #1348 (rebased onto current dev). Closing to keep a single PR for the session-isolation backend.

PR #1349, opened 2026-06-06
Per-mode tool allow-lists: lean context for files / foresight / sites

DevRohit06:
Stacked on #1348.

## Why

Every chat mode loaded every MCP tool, so the Files and Foresight agents carried the full pocket / sites / tasks / meetings / decisions / foresight surface even though they only need a slice of it. That bloats the context window. This scopes a mode to its own tools while keeping the things you want available everywhere.

## What's available everywhere (no matter the mode)

The OSS backend always keeps:
- ripple widget tools (UI rendering stays on),
- the pocket lifecycle — read / create / edit / plan (`ALWAYS_ALLOWED_MCP_SERVERS` + the pocket-creation grant), so you can make and edit pockets from any mode,
- connectors (composio), matched by server so its dynamic tool ids pass through,
- built-in Read / Write / Bash / Web / Skill — the allow-list only ever narrows `mcp__*` ids.

## What each mode adds

- **Foresight** → the foresight scenario tools.
- **Files** → nothing extra. Document scaffolding runs on the built-in Write/Edit tools, so the mode is genuinely lean (just general + built-ins).
- **Sites** → the sites authoring tools; the svelte-create sub-mode still denies the two ripple-create ids on top (deny runs after allow, so `create_svelte` survives).
- **Chat** stays unmapped and unrestricted — the do-anything surface.

## Mechanism

Two commits: the first adds `SurfaceProfile.allow_mcp_tool_ids` and threads it through `pool.run` into the Claude SDK backend (mirroring the existing deny path), with the universal grant. The second populates the per-mode lists.

The profile table is built lazily and memoized, and **degrades to unrestricted** if the tool-id constants can't be imported — tool-scoping can never break chat.

## Tests

`tests/test_claude_sdk_allow_mcp_tools.py` (filter logic + grant + always-allowed servers) and new cases in `tests/cloud/test_surface_profile.py` (each mode's allow-list, chat unrestricted, sites allow+deny interaction). Existing surface + runs suites stay green (101 tests).

## For review

The split between "general everywhere" and "specialized per mode" is a judgment call — tasks / meetings / decisions are currently treated as specialized (dropped from scoped modes, kept in broad chat). Easy to promote any of them to the general set if you'd rather they stay everywhere.

### DIFF
--- a/ee/pocketpaw_ee/cloud/auth/password_policy.py
+++ b/ee/pocketpaw_ee/cloud/auth/password_policy.py
@@ -20,7 +20,7 @@ from fastapi_users.exceptions import InvalidPasswordException
 
 logger = logging.getLogger(__name__)
 
-MIN_LENGTH = 12
+MIN_LENGTH = 6
 _HIBP_RANGE_URL = "https://api.pwnedpasswords.com/range/{prefix}"
 _HIBP_TIMEOUT = 3.0
 _CACHE_TTL_SECONDS = 24 * 60 * 60
--- a/ee/pocketpaw_ee/cloud/chat/runs/run_core.py
+++ b/ee/pocketpaw_ee/cloud/chat/runs/run_core.py
@@ -454,10 +454,14 @@ async def _drive_agent_loop(
         # prompt-sniffing gate in claude_sdk.py). ``surface_context is None`` is
         # the legacy path — no deny.
         surface_deny: frozenset[str] = frozenset()
+        surface_allow: frozenset[str] | None = None
         if ctx.surface_context is not None:
-            surface_deny = resolve_profile(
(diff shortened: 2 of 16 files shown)


DevRohit06:
Heads up: dev's #1343 (entity-aware SurfaceProfile) reworked the exact surface/tool code this PR touches and already consumes `allowed_sdk_tools`. Merging this as-is would drop dev's entity-folding, so the per-mode MCP allow-list is being rebuilt cleanly on the new architecture as a fresh PR. Leaving this open for reference until that lands.

DevRohit06:
Superseded by the clean rebuild on dev's entity-rooms SurfaceProfile (new PR). The old branch was based on pre-#1343 dev and couldn't be merged without dropping dev's entity-folding.

PR #1369, opened 2026-06-08
feat(fabric): materialize a standalone Person entity on invite accept

prakashUXtech:
## What ships

Accepting a workspace invite now materializes a standalone Fabric `Person` for the new member — the identity "spine" the VIP-onboarding flow reads. It's built from the member's own profile (name, email, avatar) plus the inviting admin's context (role, team, focus, suggested pic), with provenance recorded (who invited them, and that it was admin-seeded).

## How it's wired

A new `ee/cloud/people` module (a frozen `Person` domain + a materializer service). Writes go through the Fabric journal store, not the legacy SQLite store — that's the blessed object-lifecycle path, it's natively multi-tenant via the event scope, and the `fabric.object.*` events are the emit-on-write. The object id is deterministic (`person-{workspace}-{user}`), so re-accepting updates the existing Person rather than duplicating it. The materialize call in `accept_invite` is wrapped defensively: a journal hiccup never rolls back the accepted invite — the Person can be re-materialized later.

Standalone by design — it models identity, a different axis from the per-pocket `PocketSurfaceProfile` / entity-profile work. No import, no dependency.

## Tests

`uv run pytest tests/cloud/people/ tests/cloud/workspace/` — 108 passing, including 11 new (7 materializer unit tests + 4 accept-path: create-with-fields, idempotent-no-dup, no-context-still-creates, survives-materialization-failure). The 4 failing tests are the pre-existing `delete_preview` / `domains-verify` ones, unrelated to invites (same on a clean dev).

## Follow-ups (next slices, not this PR)

- No read API yet — the member-facing read surface (so the welcome can show the context) lands with the wave-3 work.
- C4: this adds a small `people` module under ee/cloud; I'll fold the C4 regen into the feature's final pass rather than per-slice.

## Stacked

This branches off `feat/vip-invite-context` (#1368) since it reads the new invite `context` field, so the diff shows only the Person changes. When #1368 merges, **merge it with `--rebase`** (not squash) so this stays clean — then this rebases onto dev.

Closes #1366

PR #1368, opened 2026-06-08
feat(workspace): carry optional admin context on invites

prakashUXtech:
## What ships

Workspace invites can now carry optional admin context — a one-line `focus` (what the new member will own) and a `profile_pic` reference. It's accepted on create, stored on the invite, returned on the invite response, and readable when the invite is accepted. Fully optional: omit it and invites behave exactly as before.

This is the first slice of VIP onboarding. It defines the context shape the rest of the flow reads — the personalized welcome (#368), the member's Person entity (#1366), and the agent that greets them (#1367).

## How it's wired

Threaded through the EE four-file shape: an `InviteContext` on the Beanie model, a frozen domain mirror with no Beanie import, and an `InviteContextDTO` on both the create request and the invite response, with mappers in `service.py`. `focus` and `profile_pic` are length-capped on the DTO.

## Tests

`uv run pytest tests/cloud/workspace/` — 97 passing, including 4 new tests for the round-trip (create → store → response → accept-readable) and the empty/optional path.

Heads up: 4 unrelated tests fail on this branch (`test_delete_preview_*`, `test_domains::*verify*`). I confirmed they fail the same way on a clean `dev` — pre-existing (DNS-domain verification + a delete-preview count drift), nothing to do with invites.

## Out of scope

The Fabric Person (#1366), the agent block (#1367), and the invite-modal UI (#367) — separate issues.

Closes #1365

Issue #1365, opened 2026-06-07
feat(workspace): carry admin context on workspace invites


prakashUXtech:
## What ships

A workspace invite can carry optional admin context — a one-line `focus` ("what they'll own") and a profile-pic reference. Role and group already ride on invites; this adds the two missing seed fields. The context persists on the invite doc and survives acceptance, so downstream slices can read it.

## Why now

This is the first slice of VIP onboarding and it unblocks everything else: the personalized welcome, the Fabric Person entity, and the oriented agent all read from this context. Backend lands first so the frontend and Person work have something concrete to build against.

## Scope (vertical slice)

- `ee/pocketpaw_ee/cloud/workspace/domain.py` — add a `context { focus, profile_pic }` shape on the invite model
- `ee/pocketpaw_ee/cloud/workspace/dto.py` — accept `context` on the create-invite request, return it on the invite DTO
- `ee/pocketpaw_ee/cloud/workspace/service.py` — persist context on the invite doc, carry it through to the accept path
- `tests/cloud/workspace/` — round-trip coverage

## Acceptance criteria

- [ ] Create-invite accepts and stores `context { focus, profile_pic }`
- [ ] The context is returned on the invite DTO
- [ ] The context is available at accept time
- [ ] The field is optional — an empty/absent context creates the invite with no error
- [ ] New tests cover the round-trip (create → persist → DTO → accept)
- [ ] Existing tests pass

## Out of scope

- The Fabric Person materialized on accept (separate slice)
- The agent orientation block (separate slice)
- Any invite-modal UI (paw-enterprise slice)

## Reference

PRD: `docs/design/drafts/2026-06-07-vip-onboarding-prd.md` (build chunk 1)
Epic: VIP onboarding v1

PR #1370, opened 2026-06-08
feat(workspace): surface admin context on the invite preview

prakashUXtech:
## What ships

The invite preview (`GET /workspaces/invites/{token}/preview`) now surfaces the admin's `context { focus, profile_pic }` on the ready states, so the member-facing accept → welcome flow can carry the invitee's focus + suggested pic forward (the invite itself is consumed at accept). Null when the invite has no admin context.

## Why

The member welcome (#368) needs `focus`/`pic` at welcome time, but the invite is gone by then. The frontend already calls preview before accepting — this is the cleanest place to surface it. It's the enabling slice for #368 and #369 (flagged as a follow-up in #1369).

## How

Reuses the existing `InviteContextDTO` (from #1365) on `InvitePreviewResponse`, mapped from the invite in `preview_invite`. Scoped to the `ready_*` states — the terminal states (expired / revoked / already_accepted / not_found) keep their minimal shape, consistent with how they already omit role / workspace_name.

## Tests

2 new tests (preview surfaces context; no-context is absent), written test-first (RED `KeyError: 'context'` → GREEN). `uv run pytest tests/cloud/workspace/` — 99 passing; the 4 failures are the pre-existing `delete_preview` / `domains-verify` ones, unrelated to invites.

## Stacked

Branches off `feat/vip-invite-context` (#1368) since it needs the invite `context` field — a sibling of #1369. Merge **#1368 first with `--rebase`**, then this rebases onto dev.

### DIFF
--- a/ee/pocketpaw_ee/cloud/workspace/dto.py
+++ b/ee/pocketpaw_ee/cloud/workspace/dto.py
@@ -224,6 +224,10 @@ class InvitePreviewResponse(BaseModel):
     group: str | None = None
     group_name: str | None = None
     viewer_email: str | None = None
+    # Optional admin onboarding hints (pp#1365), surfaced on ready_* previews so
+    # the member-facing accept UI can carry focus + profile_pic into the
+    # downstream VIP-onboarding welcome. None/absent when the invite has none.
+    context: InviteContextDTO | None = None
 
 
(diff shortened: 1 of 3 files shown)


prakashUXtech:
Consolidated into the VIP onboarding backend integration branch — see #1375. Closing in favor of the single integration PR (this branch's commit is preserved there).

PR #1380, opened 2026-06-08
feat(senses): Sense catalog and the sense field on connectors

prakashUXtech:
## What this adds

A capability vocabulary that sits above connectors. A Sense like `paw.email.v1` names what an agent can do; a connector declares which Senses it fills through a new `senses:` list in its YAML. A later change binds a Sense to whichever connector a tenant has enabled, so a pocket template can ask for `paw.email` and run on Gmail for one tenant and Outlook for another without edits.

This PR is the catalog half only. No resolver, no execution, no EE code.

## What's in it

- `src/pocketpaw/senses/` with the v1 core vocabulary: email, calendar, code, payments, db, docs. The `paw.*` namespace is closed, so an unknown `paw.*` id is rejected at parse time and the core can't fragment. Vendor ids in the `vendor.domain.vN` shape are accepted freely for the long tail.
- A `senses: list[str]` field on `ConnectorDef`, parsed and validated when a connector YAML loads. Connectors without the key keep working with an empty list.
- Nine connectors annotated: gmail, gcalendar, github, gitlab, stripe, postgresql, mongodb, gdocs, drive.
- `connectors_for_sense()`, a pure index from a Sense to the connectors that declare it.
- A guard test that fails if any core Sense has no connector behind it, so the vocabulary and the annotations can't drift apart.

## Testing

`uv run pytest tests/senses tests/connectors` reports 107 passing.

## One thing to know

`registry._scan()` swallows per-connector parse errors, so a bad `paw.*` id on a connector added at runtime would drop that connector quietly instead of erroring. The guard test catches it for in-repo connectors before merge; hardening the runtime path is left for a later change.

## Next

A stacked PR adds the resolver that binds a Sense to a tenant's connector and runs read actions through it.

### DIFF
--- a/connectors/drive.yaml
+++ b/connectors/drive.yaml
@@ -9,6 +9,9 @@ display_name: Google Drive
 type: knowledge
 icon: cloud
 
+senses:
+  - paw.docs.v1
+
 auth:
   method: bearer
   credentials:
--- a/connectors/gcalendar.yaml
+++ b/connectors/gcalendar.yaml
@@ -7,6 +7,9 @@ display_name: Google Calendar
 type: communication
 icon: calendar
 
+senses:
+  - paw.calendar.v1
+
 auth:
   method: oauth2
   credentials:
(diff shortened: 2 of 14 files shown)


prakashUXtech:

PR #1381, opened 2026-06-08
feat(senses): resolver that binds a Sense to a tenant's connector

prakashUXtech:
Stacked on #1380 — review and merge that one first. This branch is based on `feat/sense-catalog`, so the diff here is only the resolver.

## What this adds

The EE half of the Sense tier. Given a Sense and a workspace, it picks the connector that fills the Sense and runs an action through the existing connector execution path.

## How it resolves

- Takes the connectors that declare the Sense (from the catalog) and intersects them with the connectors the workspace has actually enabled.
- None enabled: returns nothing, so the caller can prompt the user to connect a provider.
- Exactly one: uses it.
- More than one: uses a stored preference if there is one; otherwise picks deterministically and marks the result ambiguous so the caller can ask the user to choose.

## Read-first

`execute_sense()` only runs an action whose trust level is `auto` (the read tier). Anything that needs approval — confirm, restricted, or an action with no trust set — is refused and never reaches the execute call. Both refusals (no provider, needs approval) come back as structured results rather than 500s.

## Pieces

- `ee/pocketpaw_ee/cloud/senses/`: the resolver, the read-first execute path, a `SenseFiller` seam so a skill or KB filler can register later, and an EE-side preference store. The provider preference lives next to connector enablement, not in the soul file.
- A `definitions` accessor and `get_definition()` on the connector registry so EE reads the catalog without touching private state.

## Testing

`uv run pytest tests/cloud/senses tests/cloud/connectors` reports 21 passing. `lint-imports` is clean — OSS still does not import EE.

## Scope

Resolves at workspace scope for now; pocket-scoped narrowing is noted in the code for later. The agent-facing `sense_execute` MCP tool is a separate change that waits on the connector MCP work (#1376) landing on dev.

### DIFF
--- a/ee/pocketpaw_ee/cloud/models/__init__.py
+++ b/ee/pocketpaw_ee/cloud/models/__init__.py
@@ -71,6 +71,7 @@ from pocketpaw_ee.cloud.models.pocket import Pocket, Widget, WidgetPosition
 from pocketpaw_ee.cloud.models.pocket_backend import PocketBackendCredential
 from pocketpaw_ee.cloud.models.project import Project
 from pocketpaw_ee.cloud.models.read_state import ReadState
+from pocketpaw_ee.cloud.models.sense_preference import WorkspaceSensePreference
 from pocketpaw_ee.cloud.models.session import Session
 from pocketpaw_ee.cloud.models.site import Site, SiteDomain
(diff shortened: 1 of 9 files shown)


prakashUXtech:

PR #1383, opened 2026-06-08
feat(senses): list_senses + sense_execute MCP tools

prakashUXtech:
Stacked on #1381 (which is on #1380). Review and merge those first — the diff here is only the two MCP tools.

## What this adds

The agent-facing consumer of the Sense tier. Two tools on the existing `pocketpaw_connectors` MCP server let the cloud chat agent work by capability instead of by provider:

- `list_senses()` — the capabilities (Senses) that resolve to an enabled connector in the current pocket, each with the bound connector and whether the choice was ambiguous.
- `sense_execute(sense, action, params)` — runs one read action against a Sense without naming the connector. The resolver picks the provider.

So the agent can be told "check my email" and reach Gmail (or whatever fills `paw.email` for that tenant) without the prompt naming a connector.

## Read-first

Both tools delegate to the resolver's `execute_sense()`, which owns the read-first gate. The MCP layer does not gate on its own — it translates the resolver's structured result into the MCP envelope. A write or an unmapped action comes back as a refusal, never an execution. A successful call flattens the underlying `ExecuteActionResponse` into the same shape `connector_execute` returns, so the agent gets structured fields rather than a stringified object.

## Pieces

- Two handlers + two `@tool` registrations added to `ee/pocketpaw_ee/agent/mcp_servers/connectors.py`, reusing the identity and response helpers that the connector tools already use.
- A Sense glossary entry and a sense-vs-connector section in `docs/CONNECTORS.md`.

## Testing

`uv run pytest tests/ee/agent/test_connectors_mcp_server` reports 22 passing (15 existing plus 7 for the new tools). The success-path test runs a real `ExecuteActionResponse` through `sense_execute` and asserts the flattened fields, so the serialization stays correct. `lint-imports` is clean.

## Scope

This is the agent consumer. The template `needs:[]` clause and the declarative UI binding are a separate change.

### DIFF
--- a/docs/CONNECTORS.md
+++ b/docs/CONNECTORS.md
@@ -1,5 +1,10 @@
 <!--
   Connectors documentation.
+  Updated: 2026-06-08 (sense-mcp / Sense tier chunk 4) — added the Senses
+  section: the "Sense" glossary entry, the sense-vs-connector distinction, and
+  the two new agent tools (list_senses / sense_execute on the same
+  pocketpaw_connectors MCP server) that address a capability instead of a
+  provider, with the resolver binding it to the tenant's enabled connector.
   Updated: 2026-06-08 (connector-mcp-execution / keystone) — documented the
   agent-callable connector tool surface (list_connector_actions /
   connector_execute on the pocketpaw_connectors MCP server), the v1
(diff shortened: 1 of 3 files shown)


prakashUXtech:

PR #1384, opened 2026-06-08
feat(senses): pocket templates declare the Senses they need

prakashUXtech:
Stacked on #1383 (→ #1381 → #1380). Review and merge those first.

## What this adds

A pocket template can declare the capabilities it needs, by Sense rather than by connector:

```yaml
needs:
  - paw.payments.v1
  - paw.email.v1
```

At pocket-create time, each need is checked against the tenant's enabled connectors. If something has no provider, the create still goes through and the unfilled Senses come back on the `pocket.created` event as `missing_senses`, so the UX can prompt the user to connect one. This is the piece that lets one vertical template run across tenants with different connector stacks.

## Why it doesn't block

Creation is never blocked on a missing provider — it mirrors the existing `strict=False` tolerance in the create path. A missing Sense is a prompt, not an error. A resolver hiccup is swallowed and logged, never fatal.

## Pieces

- `needs: list[str]` on `PocketTemplate`, with a validator that runs each id through the Sense validator at template-load (a malformed or unknown `paw.*` id is rejected, same rule as the connector `senses:` field).
- `_check_template_needs()` in the pocket service: resolves each need for the workspace and returns the unfilled ones. `create()` logs a warning and attaches `missing_senses` to the `pocket.created` event. No new model field, no migration.

## Testing

`uv run pytest tests/unit/test_template_schema.py tests/cloud/pockets/test_template_needs.py` reports 45 passing. `lint-imports` clean. `needs` is template metadata, so it is not merged into the rippleSpec — the compile path is unchanged.

### DIFF
--- a/ee/pocketpaw_ee/cloud/pockets/service.py
+++ b/ee/pocketpaw_ee/cloud/pockets/service.py
@@ -89,6 +89,13 @@ dispatcher + temporal scheduler call to obtain a typed ``PocketTemplate``
 from a pocket. A stale / missing slug never breaks pocket creation: the
 loader's ``strict=False`` mode returns ``None`` and the rippleSpec is
 left unmodified so a later resolver run can retry.
+Changes: 2026-06-08 (feat/sense-template-needs, Sense tier chunk 6a) —
+``create`` now reads the template's ``needs:`` Sense ids and, for each,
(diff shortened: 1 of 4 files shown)


prakashUXtech:

PR #1385, opened 2026-06-08
feat(senses): pocket data source can read from a Sense

prakashUXtech:
Stacked on #1384 (→ #1383 → #1381 → #1380). Review and merge the base chain first.

## What this adds

A pocket data source can read from a Sense. Until now RFC-04 sources hydrated state from an HTTP GET only. This adds a `sense` source type:

```json
{
  "name": "inbox",
  "type": "sense",
  "sense_id": "paw.email.v1",
  "action": "gmail_search",
  "params": { "q": "is:unread" },
  "bind": "state.inbox"
}
```

The executor resolves the Sense to the tenant's connector, runs the read action, and writes the result into `state.inbox`. So a pocket spec can bind live data by capability without naming the provider — the declarative counterpart to the agent's `sense_execute` tool.

## Ripple needs no change

The resolved value lands in `state` like any other source, so ripple reads it through the existing expression resolver. This is entirely server-side.

## Read-first and failure handling

Execution goes through the resolver's `execute_sense`, which owns the read-first gate, so a write or unmapped action is refused and never runs. A refusal or a no-provider result folds into the same per-source error aggregation HTTP failures already use, so one bad source never aborts its siblings. `params` are static in v1 — no `{state.x}` evaluation yet.

## Pieces

- `DataSourceDef` (template schema) and `SourceBinding` (runtime) gain `type` (default `http`, backward compatible), `sense_id`, `action`, `params`. `path` becomes optional; a validator requires `path` for http and `sense_id`+`action` for sense.
- `_run_sense_binding()` in the executor, branched ahead of the HTTP path. `run_sources` gained an optional `workspace_id`, threaded from the four callers that already had it in scope.

## Testing

`uv run pytest tests/unit/test_template_schema.py tests/cloud/test_pocket_source_executor.py` reports 70 passing (14 new). Full sense + connector + pocket-source regression is green (219). `lint-imports` clean.

## Known follow-up

`run_sources` still validates the backend base URL even when every selected source is a Sense, so a sense-only pocket with no HTTP backend would be rejected at that gate. Out of scope here; worth revisiting if sense-only pockets become a real shape.

### DIFF
--- a/ee/pocketpaw_ee/agent/pocket_router/router.py
+++ b/ee/pocketpaw_ee/agent/pocket_router/router.py
@@ -229,6 +229,7 @@ async def _run_tier0(
             auth_header=auth_header,
             token=token,
             only_source=classification.op_args.get("source"),
+            workspace_id=workspace_id,
         )
         errors = result.get("errors") or []
         if errors:
--- a/ee/pocketpaw_ee/cloud/pockets/refresh_scheduler.py
+++ b/ee/pocketpaw_ee/cloud/pockets/refresh_scheduler.py
@@ -163,6 +163,7 @@ async def _refresh_one_pocket(pocket: dict) -> None:
                 auth_header=auth_header,
                 token=token,
                 only_source=key,
+                workspace_id=workspace_id,
             )
             if result.get("errors"):
                 logger.debug(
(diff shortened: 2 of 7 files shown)


prakashUXtech:

Issue #1393, opened 2026-06-09
feat(push): wire notifications into agent / guardian / meeting events


DevRohit06:
## What ships

Real product events start producing browser notifications. The events that already notify the Tauri desktop app — agent finished, guardian blocked an action, meeting starting — now call the unified `notify()` so browser users get them too. Also resolves the double-notify case when someone has both the desktop app and a browser tab open.

## Why now

#1392 built the send path and qbtrix/paw-enterprise#385 made the browser receive it. This is the last slice: connect `notify()` to the events users actually care about, so push isn't just a test trigger.

Blocked by pocketpaw/pocketpaw#1392, qbtrix/paw-enterprise#385.

## Scope (vertical slice)

- Wire `notify(user, payload)` into the v1 event set: agent-complete, guardian-block, meeting-start (relates to qbtrix/paw-enterprise#234)
- Double-notify dedupe: prefer the WebSocket transport when a live connection exists, fall back to Web Push otherwise, so a user with both open doesn't get two notifications
- Tests: each event triggers a `notify()` call; dedupe picks one transport when both are available

## Acceptance criteria

- [ ] Agent-complete, guardian-block, and meeting-start each fire a notification through `notify()`
- [ ] A user with both a live WS and a push subscription gets one notification, not two
- [ ] Browser-only users still get push; Tauri-only users still get the native toast
- [ ] All existing tests pass
- [ ] New tests cover the event wiring and the dedupe

## Out of scope

- New event types beyond the v1 set — follow-up
- Per-user notification preference UI — a later pass; this issue uses the "prefer WS when live" default

## Reference

- PRD: `docs/design/drafts/2026-06-09-pwa-push-notifications-prd.md`
- Blocked by pocketpaw/pocketpaw#1392, qbtrix/paw-enterprise#385
- Related: qbtrix/paw-enterprise#234 (meeting scheduling notifications)

Issue #1391, opened 2026-06-09
feat(push): push-subscription store + VAPID key endpoints


DevRohit06:
## What ships

The backend can store a browser's push subscription and serve the VAPID public key a browser needs to subscribe. After this, a subscription round-trips and persists, even before anything sends a notification. Verifiable with curl or a unit test using a fake subscription.

## Why now

Both the send side (#3) and the browser side (#4) need somewhere to register and a key to subscribe with. This is that foundation, and it's independent of the frontend — testable on its own.

## Scope (vertical slice)

- `PushSubscription` model (Beanie doc): `user_id`, `endpoint`, `keys` (`p256dh`, `auth`), `expirationTime?`, `user_agent`, `created_at`. Endpoint is the unique key.
- `GET /api/v1/push/vapid-public-key` — returns the tenant's public key
- `POST /api/v1/push/subscribe` — upsert the subscription for the authed user
- `POST /api/v1/push/unsubscribe` — remove by endpoint
- Per-tenant VAPID keypair generation and storage. Payloads stay on the tenant box — no third-party push service (see open question #3 on where deploy provisions the key).
- Tests: subscribe upsert, duplicate-endpoint dedupe, key endpoint returns the public half only

## Acceptance criteria

- [ ] A POST with a subscription persists it and is idempotent on the endpoint
- [ ] The key endpoint serves the public key and never the private one
- [ ] Unsubscribe removes the row
- [ ] The VAPID keypair is per-tenant and the private key never leaves the backend
- [ ] All existing tests pass
- [ ] New tests cover the slice

## Out of scope

- Actually sending a push — #3
- Service worker and the browser subscribe flow — qbtrix/paw-enterprise#4
- Wiring real events — #5

## Reference

- PRD: `docs/design/drafts/2026-06-09-pwa-push-notifications-prd.md`

PR #1396, opened 2026-06-09
feat(cloud): Web Push send + dead-subscription pruning

DevRohit06:
## What this does

Adds the send side of browser Web Push: the backend can now push a notification to a user's stored subscriptions and clean up the ones that have gone dead. Builds on the subscription store from #1391. This is the push transport only — wiring it to real product events and the WebSocket-vs-push dedupe is the next PR (#1393).

> Stacked on #1391 (`feat/push-subscription-store`). Review/merge that one first.

## Changes

- `send_to_user(workspace_id, user_id, payload)` in the push service — fans out to every stored subscription for the user, signs each with the tenant's VAPID private key (via the existing decrypt chokepoint, so the key never leaves the backend), and POSTs to the browser vendor endpoint.
- The blocking `pywebpush` call runs off the event loop via `asyncio.to_thread`.
- On a `410 Gone` or `404` from the push service, the dead subscription is deleted — and only that one (matched by its unique endpoint). A transient error (5xx, network) is logged and skipped, never pruned, and never aborts the rest of the fan-out.
- `PushPayload` (title/body + optional url/icon/tag) and a `SendResult` summary (sent/pruned/failed counts).
- Contact for VAPID claims reads from `CLOUD_PUSH_CONTACT`, defaulting to a non-personal `mailto:`.

## Tests

12 new tests (30 total with #1391's): fan-out across multiple subscriptions, tenant-key signing, 410/404 pruning, pruning only the dead endpoint, one bad endpoint not aborting the rest, unexpected errors swallowed, the empty no-op case, and the configurable contact. `pywebpush.webpush` is mocked in every test — nothing hits the network.

## Notes

Part of the PWA + Web Push work (design + PRD in paw-workspace `docs/design/drafts/`). Follow-up: #1393 wires this into agent/guardian/meeting events with desktop-vs-browser dedupe.

### DIFF
--- a/ee/pocketpaw_ee/cloud/__init__.py
+++ b/ee/pocketpaw_ee/cloud/__init__.py
@@ -812,6 +812,16 @@ def mount_cloud(app: FastAPI) -> None:
     register_meeting_notification_listeners()
     register_meeting_calendar_listeners()
 
+    # Push notifications fan-out (#1393) — v1 product events
+    # (agent.stream_end / instinct.approval.created / meeting.started) →
+    # ``dispatch.notify``, which forks WS-vs-Web-Push so a user with both the
+    # desktop app and a browser tab open is notified exactly once. Same
+    # constraint as the other bus subscribers: register AFTER init_realtime
+    # installed the singleton bus.
+    from pocketpaw_ee.cloud.push.listeners import register_push_event_listeners
+
+    register_push_event_listeners()
+
     # In-process daily-snapshot scheduler — opt-in via env var.
     #
     # Default OFF in tests + dev (each pytest run would otherwise spawn a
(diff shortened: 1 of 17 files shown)


DevRohit06:
Folded a one-commit fix into this branch (938a821a).

The send path passed the tenant VAPID key to `webpush()` as a raw PKCS#8 PEM string. pywebpush hands a string to `py_vapid.Vapid.from_string`, which only parses raw/DER base64 keys — a PEM gets its newlines stripped, the `-----BEGIN-----` header base64-decodes to garbage, and it throws an ASN.1 "invalid length" error. Every send failed, so no Web Push ever reached a browser. Caught it debugging a live install where `send_to_user` returned sent=0 failed=N against real FCM subscriptions.

Fix: parse the PEM into a `Vapid01` instance with `from_pem` once per fan-out and hand `webpush` the instance (which it accepts directly). Confirmed end to end — FCM returned 201 and the notification rendered.

The existing send tests mock `webpush`, so the real key parsing never ran and one test actually asserted the broken PEM string was passed. Updated that assertion to expect the parsed signer and added a regression test that the key handed to `webpush` can really sign a claim set. Full push suite green (56), ruff clean.

Heads-up for the stack: #1397 and #1605 sit on this branch, so they'll want a rebase to inherit the fix.

PR #1399, opened 2026-06-09
Studio and Code surfaces with a media generation MCP server

prakashUXtech:
## What this adds

Two new chat surfaces, built the same way `/sites` is:

- **Studio** (`surface=studio`) — the user describes an image or video, the agent generates it, and the result lands in a gallery pocket. Backed by a new in-process MCP server `pocketpaw_media` with two tools: `image_generate` (Gemini, reusing the existing image path) and `video_generate` (Replicate/Kling over httpx, polled to completion).
- **Code** (`surface=code`) — the agent edits and runs code in the workspace with the SDK's Bash/Read/Write/Edit/Glob/Grep tools. No new execution layer; the surface profile scopes the toolset and a `code` skill drives the edit → run → verify loop.

Both profiles set `ripple_mode="off"` so the agent generates media or edits code rather than trying to build a Ripple dashboard.

## Why an MCP server instead of a Tool

The cloud chat agent runs on the Claude Agent SDK backend, which only sees SDK built-ins plus MCP servers registered through the `pocketpaw.mcp_servers` entry point. A plain `BaseTool` is invisible to it, so media generation is wired as `CloudMediaMcpProvider`.

## Files

- `ee/pocketpaw_ee/agent/mcp_servers/media.py` — the media MCP server
- `ee/pocketpaw_ee/cloud/surface/handlers/{studio,code}.py` + enum/registration in `domain.py` and `service.py`
- `src/pocketpaw/bundled_skills/_bundled/skills/{studio,code}/SKILL.md`
- `config.py` — `replicate_api_token`, `fal_api_key`, `video_model`
- tests under `tests/cloud/`

## To run it live

- Image generation uses the existing `POCKETPAW_GOOGLE_API_KEY` (Gemini).
- Video generation needs `POCKETPAW_REPLICATE_API_TOKEN` (model defaults to `kwaivgi/kling-v2.0`). Until that's set, `video_generate` returns a plain "set the token" error rather than failing oddly.

## Tests

77 cloud tests (surface/studio/code/media) and 240 tool tests pass; `ruff` is clean. Two pre-existing collection errors in `tests/cloud/extraction/` (missing `pypdf`) are unrelated and were skipped.

## Notes

Pairs with the paw-enterprise PR that adds the `/studio` and `/code` routes. Frontend expects a `pocket_created` event with `pocket.type == "studio"` / `pattern == "gallery"` after each Studio generation. Not for merge yet — this is for review and a morning smoke.

### DIFF
--- a/ee/pocketpaw_ee/agent/mcp_servers/media.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/media.py
@@ -145,20 +145,31 @@ def _build_gallery_spec(media: list[dict[str, Any]]) -> dict[str, Any]:
                 ],
             }
         )
+    # version + a non-dashboard intent make pocketsStore.toRippleEnvelope pass
+    # the spec through untouched, and Ripple render it in NODE mode (a bare
+    # {"ui": [...]} got intent="dashboard" stamped on, which makes Ripple read
+    # `widgets` — always empty here — and render "No widgets yet"). `ui` must
+    # be a single root node, not a list: NodeRenderer takes one node.
     return {
-        "ui": [
-            {
-                "id": "studio-title",
-                "type": "text",
-                "props": {"text": "Generated media"},
-            },
-            {
-                "id": "studio-grid",
-                "type": "grid",
-                "props": {"columns": 3, "gap": "1rem"},
-                "children": tiles,
-            },
-        ]
+        "version": "2.0",
+        "intent": "gallery",
+        "ui": {
+            "id": "studio-root",
+            "type": "container",
+            "children": [
+                {
+                    "id": "studio-title",
+                    "type": "text",
+                    "props": {"text": "Generated media"},
+                },
+                {
+                    "id": "studio-grid",
+                    "type": "grid",
+                    "props": {"columns": 3, "gap": "1rem"},
+                    "children": tiles,
+                },
+            ],
+        },
     }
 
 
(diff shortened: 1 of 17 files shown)


prakashUXtech:
Live-smoked this on a local stack this morning: backend from this branch, frontend from the paired paw-enterprise branch, driven in a real browser.

What passed as shipped:
- /studio renders gallery + chat, the surface stamp fires, and the agent stays in media mode instead of building a dashboard
- missing-key paths for both tools return the plain 'set the key' message with no stack trace (video genuinely reaches the tool before erroring)
- /code: the agent created and ran a script through Bash/Write and the output streamed into the chat rail

Two things broke on first real use, fixed in the follow-up commits:
- Image generation could never work. The default model gemini-2.0-flash-exp is retired on the predict endpoint, and generate_images (predict) is paid-tier only, so free keys fail on every model. fdc30b8e adds a shared generate_image_file() that routes gemini-*-image models through generateContent and keeps imagen-* on predict; the default is now gemini-2.5-flash-image. Verified live up to Google's quota wall — the key on this machine has zero free-tier image quota (limit: 0), so the final hop needs a billed key.
- The /code file tree was permanently empty in cloud mode. The EE Files Tab v2 router shadows GET /files/browse with a required mount param, so the localFs bridge got a silent 422. 983ad7a5 makes mount optional and serves the OSS path shape from the same route, and expands ~ in the OSS resolvers so the IDE can address ~/foo.

Still open:
- a real gallery landing after image gen is only covered by the mocked test until a billed Google key is in place; video needs POCKETPAW_REPLICATE_API_TOKEN
- tests: 18 image/media, 33 files-security, 79 files-v2 — all green, ruff clean

PR #1408, opened 2026-06-10
Add the /belt develop-station surface (BS-2)

prakashUXtech:
Stacks on #1407 (the loom MCP wiring) — review and merge that first.

Adds the `/belt` surface: the develop station of the Belt & Pulley line. The agent on it takes a coding task, grounds itself with loom's `orient()` before reading any code, implements the change in a station worktree, then proposes the resulting diff through the Instinct gate for a human to review. It never applies to the user's branches directly, never pushes or merges, and never claims success unless the gate tool returned ok.

## What's here

- `SurfaceKind.BELT` in `domain.py`.
- A ripple-off `SurfaceProfile` in `service.py`: the coding SDK-tool allowlist (Bash/Read/Write/Edit/Glob/Grep), the `belt` skill, and an MCP allow-list built from the 5 loom orientation tools (`LOOM_TOOL_IDS`, imported from this branch's base) plus the gate tool. It loads through the existing try/except degraded path, so a failed import drops to no MCP restriction rather than breaking chat.
- `handlers/belt.py` — the preamble that enforces the station loop: orient first via loom, develop in the worktree, propose via the gate.
- A bundled `belt` skill carrying the same playbook in skill form, with a worked `belt_propose_change` example.
- Tests for the handler, the profile, and the `belt` wire-string mapping.

## Gate tool id

The Instinct gate tool (`mcp__pocketpaw_belt__belt_propose_change`) is built in a sibling branch (`agent/mcp_servers/belt.py`), so its constant isn't importable here yet. I used the literal id with a comment naming the sibling module; the import reconciles when both PRs land.

## Test evidence

`uv run pytest tests/cloud/surface/ -q`

```
................................................................         [100%]
64 passed in 0.91s
```

`uv run ruff check` on all touched files: `All checks passed!`

### DIFF
--- a/ee/pocketpaw_ee/cloud/surface/domain.py
+++ b/ee/pocketpaw_ee/cloud/surface/domain.py
@@ -48,6 +48,12 @@
 # image + video) and ``CODE`` (/code — the agent edits + runs code). Both get
 # ``ripple_mode="off"`` profiles in ``service.py`` so the agent generates media /
 # edits code instead of defaulting to a ripple ui-spec dashboard.
+# Changes: 2026-06-10 (feat/belt-surface, BS-2 Belt & Pulley stations thin
+# slice) — added the ``BELT`` surface (/belt — the develop station). Its
+# ``ripple_mode="off"`` profile in ``service.py`` scopes the agent to the loom
+# orientation tools + the Instinct gate tool so it orients first, develops in a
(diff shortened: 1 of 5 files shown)


prakashUXtech:
Folded into dev via the integration PR #1413 (merge commits preserve this branch's history there). Keeping this PR as the focused review record.

PR #1409, opened 2026-06-10
Belt code-change gate: propose a diff, approve it, get a PR (BS-3)

prakashUXtech:
Stacks on #1407 (review that first — this branch is based on `feat/belt-loom-mcp`).

## What this does

The Belt & Pulley develop station now has a way to ship code without ever touching the live checkout on its own. The station produces a unified diff and proposes it through Instinct — the human approve/reject layer. A person approves it in The Tray, and only then does an executor apply the diff on a fresh branch and open a PR. The captain still merges on GitHub. Instinct is the mid gate, not an auto-merge.

The two halves mirror the existing pocket-write bridge so there's no new pattern to learn.

**Propose side — `pocketpaw_belt` MCP server (`belt_propose_change`)**
- Checks workspace/user identity from the chat session's ContextVars
- Refuses an empty diff, and caps size at ~1500 changed lines / 200KB with a "split the task" message
- Resolves the `repo` path inside a settings-driven allowlist and refuses anything outside it
- Files an Instinct Action carrying the change as a `_code_change` blob, and returns `ok` only after the store confirms the row — no phantom success. The diff rides verbatim; it's data, never interpolated.

**Apply side — executor (fires on approve)**
- Fetch origin, make a fresh per-action worktree off `origin/<base>`, `git apply --3way` from a temp file
- Branch `feat/belt-<id>`, Conventional-Commits commit (no AI attribution), push
- Open the PR through an injectable opener (default shells `gh pr create`; tests inject a fake)
- `mark_executed` with the PR url, branch, and file count
- The worktree is always removed. An apply conflict or any failure marks the Action failed with a clear reason and leaves no half-state.

**Dispatch wiring**
- The instinct router's approve / bulk-approve routes dispatch the executor by action kind, keyed on the `_code_change` blob. The pocket-write path is untouched.
- `_assert_code_change_workspace` binds a code-change Action to the approver's workspace on the approve, bulk-approve, and reject paths.

Plus the `CloudBeltMcpProvider` entry point and the `belt_repo_allowlist` setting (the allowlist boundary, defaulting to the workspace root when unset).

## Security checklist

- No `shell=True`, no string-interpolated commands — every git/gh call is an argv list through one `_run` chokepoint
- The diff is data: written to a temp file and fed to `git apply <file>`, never echoed or put on a command line. The temp file is removed before commit and in the finally block.
- The repo path is allowlist-gated before any git op, on a resolved path (so `..` can't escape), and re-resolved at execute time (defense in depth)
- No diff content and no secrets in any log line — only action ids, branch names, and file counts
- The worktree is cleaned up on every path, success or failure
- The executor cannot be reached for non-`code_change` kinds — proven by the router-seam tests below

## Tests

`tests/cloud/test_belt_gate.py` drives the whole gate through the real path against a local bare repo (origin) + a seeded clone:
- propose -> approve -> apply: branch + commit land in the bare origin, the fake opener gets the right args, `mark_executed` carries the PR url
- two actions back-to-back with no cross-contamination (distinct branches/PRs, clean base each time)
- reject round-trips for a code_change Action (no PR, no executor run)
- apply conflict -> Action failed, worktree cleaned, nothing pushed
- size-cap rejection, missing identity, repo-outside-allowlist refusal, empty diff

`tests/cloud/test_ee_instinct.py` and `tests/cloud/test_instinct_approval_security.py` add the router-seam coverage the e2e file doesn't reach:
- approve / bulk-approve dispatch the executor only for a `_code_change` Action, never for a plain one
- reject never reaches the executor
- cross-workspace code-change approval / bulk-approval / rejection returns 403 (`instinct.cross_workspace_approval`), and a same-tenant approval passes the gate

Targeted run green: 203 passed across the belt gate + every instinct test module. `ruff check` clean on all touched files.

## Docs

Internal subsystem — the new `belt_repo_allowlist` setting carries its own description in `config.py`. No public CLI/API reference change.

### DIFF
--- a/ee/pocketpaw_ee/agent/mcp_servers/belt.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/belt.py
@@ -1,0 +1,440 @@
+# belt.py — in-process MCP server exposing the Belt & Pulley develop-station
+# code-change gate to the claude_agent_sdk cloud chat backend. Created:
+# 2026-06-10 (feat/belt-gate, BS-3).
+#
+# What this file does: clones the media.py shape — a single
+# ``create_sdk_mcp_server`` with an SDK import-guard, ``SERVER_NAME`` /
+# ``*_TOOL_ID`` allowlist constants, ContextVar-sourced identity (the same
+# ``current_workspace_id`` / ``current_user_id`` / ``current_session_mongo_id``
+# accessors in ``ee.cloud.chat.agent_service`` the media / sites servers read),
+# and the ``_error_response`` / ``_success_response`` helpers. The single tool
+# id namespaces as ``mcp__pocketpaw_belt__belt_propose_change`` so the Claude
+# Code allowlist machinery matches it (a sibling PR hardcodes this exact id).
+#
+# One SDK @tool def:
+#   * belt_propose_change — the develop station agent produces a unified diff
+#     and proposes it THROUGH Instinct (the human approve/reject layer, "sudo
+#     for agents"). This tool does NOT apply anything: it validates the inputs
+#     (identity present; diff non-empty and under the size cap; repo resolves to
+#     an existing git repo INSIDE the settings-driven allowlist) and then files
+#     an Instinct Action carrying a ``_code_change`` blob. A human approves it
+#     in The Tray; the apply-on-approve executor (ee/cloud/belt/executor.py)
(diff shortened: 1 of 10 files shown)


prakashUXtech:
Folded into dev via the integration PR #1413 (merge commits preserve this branch's history there). Keeping this PR as the focused review record.

PR #1444, opened 2026-06-12
Make workspace-scoped connectors visible to the agent surface

prakashUXtech:
Found live on the Nerve stack: a connector enabled workspace-wide showed as Connected in the UI, but the chat agent's list_connector_actions returned [] — list_pocket_connectors and is_connector_bound_to_pocket only matched scope=pocket rows. Both now also match scope=workspace rows. Workspace scope is the tenant boundary, so every pocket in the workspace may see and (trust-gated) use those connectors; cross-tenant filtering is unchanged.

This commit was authored during today's demo session inside the nerve worktree and rescued from there — it was never on a pushed branch. One file, +32/-11.

### DIFF
--- a/ee/pocketpaw_ee/cloud/connectors/service.py
+++ b/ee/pocketpaw_ee/cloud/connectors/service.py
@@ -25,6 +25,14 @@
 #   ``backend_type="connector"`` backend. Keeps the WorkspaceConnector Beanie
 #   read in THIS service (the owner of the connector docs) so the pockets
 #   service never imports the connector model.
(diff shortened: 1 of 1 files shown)


prakashUXtech:
Superseded by #1445, which merged this morning with the identical service.py change (same Or/And scope query, same name dedupe) plus the MCP-handler reach for unanchored chats, the sense_execute parity fix, and the CONNECTORS.md update. Nothing here is missing from dev — closing without merge. Two of us hit the same bug from different directions within hours, which is its own kind of confirmation the fix was right.

PR #1445, opened 2026-06-12
Connectors: workspace-scoped connectors reachable from the agent surface and unanchored chats

prakashUXtech:
## The bug

Enable a connector workspace-wide and the connectors page shows it as Connected, but the agent in any chat says there are no connectors. The two agent-side consumers only matched `scope="pocket"` rows: `list_pocket_connectors` (behind the `list_connector_actions` MCP tool) and `is_connector_bound_to_pocket` (the tenant gate in `connector_execute`). Workspace-enabled connectors were invisible to the agent, and unanchored chats (plain DM/group threads, `pocket_id=None`) short-circuited to a "bind a connector" message before ever asking the service.

## The fix

**`ee/pocketpaw_ee/cloud/connectors/service.py`** — both readers now match `scope="workspace"` OR (`scope="pocket"` AND this pocket's id) via beanie `Or`/`And` operators. The listing dedupes by connector name when the same connector is enabled at both scopes. Docstrings updated to state the new contract.

**`ee/pocketpaw_ee/agent/mcp_servers/connectors.py`** — dropped the `pocket_id=None` short-circuits in `list_connector_actions` and `connector_execute`. Both handlers now call the service with `pocket_id or ""`: the pocket arm of the query matches nothing for an empty id, so unanchored chats reach exactly the workspace-scoped set. Execute builds `ExecuteActionRequest(scope="workspace", pocket_id=None)` when unanchored and keeps `scope="pocket"` when anchored, so credentials resolve from the right scope. Test expectations in `tests/ee/agent/test_connectors_mcp_server/test_mcp_tool.py` updated to match.

**`docs/CONNECTORS.md`** — the agent-tool-surface section described the old pocket-only behavior; updated to the new reach.

## Why this is safe tenancy-wise

Workspace scope IS the tenant boundary. A workspace-scoped row passing from every chat in that workspace leaks nothing across tenants: the query is still filtered on `workspace_id`, and cross-workspace checks stay False. Pocket-scoped connectors remain private to their room — pocket A's binding is still unreachable from pocket B and from unanchored chats. The read-first trust gate is untouched: `confirm`/`restricted` actions are still refused before execution.

## Tests

- `tests/ee/agent/test_connectors_mcp_server/` — 22/22 green.
- `tests/cloud/test_connectors_execute.py` — 9 passed; 6 failures are pre-existing and environment-sensitive (401s from the auth posture), verified to fail identically on the base commit (`origin/dev`) with the base file versions.

Proven live on the nerve demo stack: from an unanchored chat the agent enumerated 14 actions across the workspace-enabled connectors and executed an `application_count` read end to end.

### DIFF
--- a/docs/CONNECTORS.md
+++ b/docs/CONNECTORS.md
@@ -1,5 +1,12 @@
 <!--
   Connectors documentation.
+  Updated: 2026-06-12 (workspace-scope reach) — the agent tool surface now
+  reaches workspace-scoped connectors: list_connector_actions returns the
+  current pocket's bound connectors PLUS the workspace-enabled ones (deduped
+  by name), unanchored chats (no pocket) reach exactly the workspace-scoped
+  set, and connector_execute passes for pocket-bound OR workspace-scoped
+  rows (executing with workspace-scope credentials when unanchored). The
+  read-first / write-blocked trust gate is unchanged.
   Updated: 2026-06-11 (connector cookie/session auth) — documented two new
   auth methods on the DirectREST engine: `cookie` (emits a Cookie: header from
   a declared credential, name set via auth.credential) and `header` (emits an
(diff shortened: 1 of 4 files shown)


prakashUXtech:
Added 8337f18e after a sweep for the same bug class: sense_execute kept the pocket_id=None refusal while list_senses already resolved workspace-scoped providers for unanchored chats — so the agent could list a sense it wasn't allowed to run. The handler now passes pocket_id=None through to execute_sense, which handles it natively and owns the read-first gate. The sweep found no other live instance of the class; surface-profile derivation stays pocket-only on purpose (workspace connectors shouldn't force-load skills into every pocket).

PR #1450, opened 2026-06-12
fix(cloud-chat): give pocket-anchored chats real context about their pocket

prakashUXtech:
Chatting with a pocket-anchored room, the agent knew almost nothing about the pocket. Asked "what's this pocket about?", it described a fully composed applications-triage template as "an empty shell, 0 widgets registered". Two causes:

1. Only home pockets got pocket context in the system prompt. Every other anchored pocket shipped a bare `<current-pocket id="..."/>` tag: no name, no description, no spec. The Pocket document was already fetched during scope resolution and then dropped.
2. The `get_pocket` agent view leads with the legacy `widgets[]` array, which is empty on template-instantiated pockets. Agents read `widgets: []` first and stop there; the real layout lives in `rippleSpec`.

What this PR does:

- New `summarize_ripple_spec()` in `pockets/spec_ops.py`: a pure helper that turns a rippleSpec into light metadata (ui node count and types, state keys, sources, action keys, legacy widget count). Degrades to a fixed shape on None/empty/malformed specs.
- `ScopeContext` carries a pocket summary, populated in `_resolve_pocket` and `_resolve_session` from the doc already in hand. Zero extra DB reads.
- `build_behavior_instructions` renders a `<pocket-summary>` block for every anchored pocket: name, description, type, template/pattern, spec metadata, plus a hint to call `get_pocket` for the full spec. Home pockets keep their existing HOME_POCKET_PROMPT and backend summary unchanged; a test pins the old block as a byte-identical prefix of the new one.
- `get_pocket`'s agent view now leads with a `_summary` field built from the same helper, and says outright when `widgets[]` is legacy and empty.

Since member-authored spec content now flows into system prompts, everything interpolated is sanitized at both seams: whitespace collapse, angle brackets replaced with lookalikes so no forged closing tag survives (case variants and split-across-fields included), 120-char per-item clamp, query strings and fragments stripped from source paths so embedded credentials can't land in prompts. Capped lists carry honest "+N more" markers. Review ran an adversarial battery against both the prompt path and the real Beanie agent_view path; tests pin the single-closing-tag invariant.

Out of scope, noted for later: connector enumeration in the summary, and credentials embedded in path segments (pre-existing exposure via the raw rippleSpec the view already returns).

Tests: 382 passing across the touched suites (spec_ops, scope resolution, behavior instructions, agent view, chat, runs, pocket router). Side effect worth knowing: pocket_mutation "replace" frames embed the agent view, so they now carry `_summary` too; extra-key-tolerant consumers are unaffected.

Internal change only; no living doc describes the pocket chat context surface (checked).

### DIFF
--- a/ee/pocketpaw_ee/agent/mcp_servers/connectors.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/connectors.py
@@ -142,9 +142,7 @@ async def _list_connector_actions_handler(args: dict) -> dict:  # noqa: ARG001 â
         # Unanchored chats (pocket_id=None) pass "" â€” the pocket arm of the
         # service query matches nothing, so only workspace-scoped connectors
         # come back. Anchored chats get pocket-scoped + workspace-scoped.
-        connectors = await connectors_service.list_pocket_connectors(
-            workspace_id, pocket_id or ""
-        )
+        connectors = await connectors_service.list_pocket_connectors(workspace_id, pocket_id or "")
     except Exception as exc:  # noqa: BLE001
         logger.warning("list_connector_actions failed", exc_info=True)
         return _error_response(f"list_connector_actions failed: {exc}")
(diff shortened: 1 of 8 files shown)


prakashUXtech:
Rebased onto current dev (five connector commits landed after this branched). Also carries a one-commit chore: dev's tip had an unformatted connectors.py that failed the ruff format gate for every PR — formatted it here so the gate goes green for everyone.

PR #1468, opened 2026-06-15
Make chat and pocket replies use ripple's rich widgets, not primitive rebuilds

prakashUXtech:
## What

Ripple's catalog is deep — composite dashboards, app-shell, master-detail, comparison-table, data-grid, timeline, audit-log, and more. But the chat agent and the pocket specialist kept building from the basic primitives (flex / text / table / stat) instead of reaching for those typed widgets. This is the "ripple has great components but the chat and pockets don't use them" problem.

This is a prompt-only change in two files. No logic changes.

## How I found it (and confirmed the fix)

I ran a prompt eval: take the *exact* production prompts, give a sub-agent a realistic request, have it author the spec under those prompts' rules, then score the result for widget richness vs. the ideal typed widget. Same harness before and after, so the delta is the signal.

The baseline isolated two real leaks:

| Surface | Before | What it did |
|---|---|---|
| Chat (data in hand) | **3.25 / 10** | rebuilt a comparison as a plain `table`, a ranked list as `table`+`stat`, a timeline as text rows |
| Complex "app for X" pockets | **3 / 10** | declined `app-shell` / `master-detail` / `comment-thread`, hand-built from `split`/`flex`/`each` |
| Composite dashboards | wrapped | sometimes surrounded a `pipeline-dashboard` with redundant sibling `stat` tiles (duplicating its internals) |

Simple pockets were already fine (8–10) — the prompt's rules carry them. The gap was chat across the board, and complex pockets.

## The fix

- **`_inline.py` (chat):** a "REACH FOR THE TYPED WIDGET" block — a shape→widget map with **inlined prop schemas + worked examples** for `comparison-table`, `data-grid`, `timeline`, `kv-table`, `status-dot`, `pricing-table`. Because the schemas are inline, there's no `get_inline_widget_help` round-trip to discourage reaching for them. Also reframed the "six core widgets cover ~90%" opening so it reads as the base, not the ceiling.
- **`_pockets.py` (pockets):** two rich worked examples added to `_CREATION_EXAMPLES_MCP` — a composite dashboard (`pipeline-dashboard`, all data in props, with an explicit "don't wrap it in sibling stat tiles" note) and a full app (`app-shell` + `sidebar` + `master-detail` + `comment-thread`). The examples there were only a basic todo + an espresso viewer, so complex briefs had no rich pattern to imitate.

## After (same eval)

| Surface | Before | After |
|---|---|---|
| Chat (data in hand) | 3.25 | **9.1** — data-grid, comparison-table, timeline, status-dot |
| Pocket floor (recipe-fetch disabled) | 8.2 | **9.2** |
| support-app | 3 | **9** — app-shell + master-detail + comment-thread |
| sales-pipeline | 8 | **10** — one clean pipeline-dashboard |

57 prompt tests pass (`test_ripple_catalog`, `test_widget_diversity`, `test_pocket_prompts_single_source`, `test_inline_prompt_source_of_truth`).

## Honest caveats

- The eval is a **proxy** — Claude sub-agents authoring under the real prompts, not the live runtime (the local backend is degraded, no API key). The model family matches production (Claude) and the before/after uses one harness, so the delta is trustworthy; absolute production numbers may differ.
- The fix raises the **floor** (it works even when the agent skips the optional `kb search` recipe fetch). It doesn't change the recipe mechanism itself — a separate, larger option would be to auto-inject the matching recipe at dispatch instead of relying on the agent to fetch it.
- Minor residuals: a status reply sometimes hand-rolls the row frame instead of using `kv-table` (still uses `status-dot`); no-fetch pocket specs occasionally guess a prop name that the persist-time alias-fix reconciles.

Findings writeup: `docs/design/drafts/2026-06-15-ripple-authoring-widget-richness.md`.

### DIFF
--- a/src/pocketpaw/ripple/_inline.py
+++ b/src/pocketpaw/ripple/_inline.py
@@ -641,7 +641,8 @@ Final self-check before sending:
 ✔ Matched the reply SHAPE to its typed widget (comparison-table / data-grid /
   timeline / kv-table / status-dot / pricing-table) — did NOT rebuild it from a
   plain table / text / flex
-✔ Used a core widget, an inlined typed widget, or called `get_inline_widget_help` BEFORE emitting any other type
+✔ Used a core widget, an inlined typed widget, or called `get_inline_widget_help`
+  BEFORE emitting any other type
 ✔ Multi-step / wizard / intake flow → called `start_flow`, not a hand-authored or `set`-stepped spec
 ✔ Pending Instinct approvals → Approve/Reject buttons `api`-POST the route, NOT chat.send
 ✔ Open Tray affordance navigates to /deep-work (not a chat.send)
@@ -706,7 +707,8 @@ source-card    — { source, title, url?, color? }
 callout        — { variant:"info|insight|warning", title, text }
 definition-list— { items:[{term,definition}], layout?:"inline|stacked" }
 metric         — { label, value, trend?, description? }
-funnel         — { data:[{label,value}], title?, sort?:"descending|ascending|none", colors?, height? }
+funnel         — { data:[{label,value}], title?, colors?, height?,
+                   sort?:"descending|ascending|none" }
 gauge          — { value, max?, label?, title?, color?, height? }
 
 Worked examples — the typed-widget choice in action (data already in hand):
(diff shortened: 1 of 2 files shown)


prakashUXtech:
Closing as superseded. dev's chain-flow-v2 rewrite of `ripple/_inline.py` already carries the widget-richness guidance this PR added, and this branch's only working copy was orphaned uncommitted work on a since-merged branch. If a specific richness gap survives chain-flow-v2, we'll reopen a fresh change against current dev.

prakashUXtech:
Rebased onto current dev (was 425 behind). Two conflicts, both in the header changelog, resolved keeping every entry in date order. The rebase surfaced three E501s the branch had introduced, so there's a follow-up commit wrapping those prompt lines; the prompt text itself is unchanged. Ruff clean, module imports and the funnel/gauge rule is intact.

Heads up: `_inline.py` and `_pockets.py` have uncommitted edits in the main checkout right now, so check for overlap before merging.

PR #1475, opened 2026-06-16
fix(uploads): accept SVG logos and serve them without an XSS hole

prakashUXtech:
## What's wrong

The white-label Branding panel says "Upload logo — SVG or PNG", but picking an SVG fails with `unsupported_mime`. PNG works fine. The logo upload goes through the generic uploads flow, and the backend decides the type by sniffing magic bytes. An SVG is XML text, so it has no binary image signature — the sniff falls back to the declared content-type, and `image/svg+xml` was never on the image allow-list. So every SVG bounced.

## The fix

Two parts, both needed.

**Accept SVG, narrowly.** An upload is treated as an SVG only when all three agree: it declares `image/svg+xml`, the filename ends `.svg`, and the bytes start (after an optional BOM/whitespace) with `<?xml` or `<svg`. A bare `text/xml`/`text/plain` blob, or a non-SVG file that just claims to be one, is still rejected — this doesn't open the allow-list to arbitrary XML or HTML.

**Serve it safely.** This is the part that needs a careful look. An SVG is active content — it can carry `<script>`, inline `on*` handlers, `javascript:` URLs, or a `<foreignObject>` smuggling HTML. A logo rendered through `` never runs that script, but a victim who opens the asset's signed URL directly would be in a script-capable context. So uploaded SVGs get three independent defenses:

1. **Sanitized on upload** — `<script>`, `on*` handlers, `<foreignObject>`, `javascript:` URLs, and `<!DOCTYPE>`/`<!ENTITY>` are stripped before the bytes are stored.
2. **Forced download** — SVG stays out of `INLINE_MIMES`, so the response carries `Content-Disposition: attachment` (plus `X-Content-Type-Options: nosniff`).
3. **Locked-down CSP** — SVG responses send `Content-Security-Policy: default-src 'none'; style-src 'unsafe-inline'; sandbox`, so even a forced inline render can't execute, fetch, or escape its origin.

Both serving paths (the core download route and the workspace download route) share one helper so they can't drift apart.

## Security review requested

This is security-sensitive — it's stored-XSS territory. Please review the serving defenses specifically:

- Is `Content-Disposition: attachment` + the CSP enough for a direct navigation to the signed URL, given how logos are actually fetched (`` with cookie auth)? Is there a case where a browser would still render the SVG inline and run script?
- The sanitizer is a dependency-free regex pass (no `bleach`/`lxml`), deliberately treated as defense-in-depth behind the attachment + CSP layers rather than the sole barrier. Any bypass worth closing now versus relying on the serving headers?
- The accept gate keys on declared content-type + extension + signature. Anything that should tighten it further?

I leaned on layered serving defenses over a heavyweight sanitizer on purpose, partly to avoid pulling a fresh dependency past the supply-chain release-age window. Happy to add a vetted sanitizer later if review prefers it.

## Tests

- `tests/uploads/test_svg_safety.py` (new) — the accept gate, the sanitizer, and the response headers in isolation.
- `tests/uploads/test_service.py` — SVG accepted; stored bytes are sanitized; a declared-SVG-without-signature and a real-SVG-with-wrong-extension are both rejected; PNG still works. Updated the two old tests that asserted SVG rejection.
- `tests/uploads/test_router.py` and `tests/cloud/uploads/test_router.py` — SVG round-trips and comes back with attachment + nosniff + CSP, never inline; a script-laden SVG comes back sanitized.

120 tests pass across the touched files (`uv run pytest tests/uploads/ tests/cloud/uploads/test_router.py`).

## Docs

No user-facing upload/mime doc lists the allow-list or the serving rules — the validation logic is the source of truth and is commented inline. Internal change, no doc drift.

## Note on the EE test fixture

The workspace upload router tests gate on an RBAC dependency that wasn't satisfied by the local test fixture, so the whole file 401'd locally (independent of this change). Added a `current_active_user` override to the `ee_client` fixture so the RBAC gate resolves — this un-breaks the pre-existing tests in that file too. Test-only; no production code involved.

### DIFF
--- a/ee/pocketpaw_ee/cloud/uploads/router.py
+++ b/ee/pocketpaw_ee/cloud/uploads/router.py
@@ -66,9 +66,10 @@ from fastapi import (
 )
 from fastapi.responses import StreamingResponse
 
-from pocketpaw.uploads.config import INLINE_MIMES, UploadSettings
+from pocketpaw.uploads.config import INLINE_MIMES, SVG_MIME, UploadSettings
 from pocketpaw.uploads.errors import NotFound
 from pocketpaw.uploads.factory import build_adapter
+from pocketpaw.uploads.svg_safety import svg_response_headers
 from pocketpaw_ee.cloud.license import require_license
 from pocketpaw_ee.cloud.shared.deps import (
     current_user_id,
@@ -577,15 +578,18 @@ async def download(
         rec, it = await _SVC.stream(file_id, user_id, workspace)
     except NotFound as e:
         raise HTTPException(status_code=404, detail="not found") from e
-    disposition = "inline" if rec.mime in INLINE_MIMES else "attachment"
-    return StreamingResponse(
-        it,
-        media_type=rec.mime,
-        headers={
+    # SVG is active content: force a download + a script-blocking CSP so a
+    # direct navigation to this URL can't execute embedded script. The logo
+    # still renders via <img src> (script-safe). See svg_response_headers.
+    if rec.mime == SVG_MIME:
+        headers = svg_response_headers(rec.filename)
+    else:
+        disposition = "inline" if rec.mime in INLINE_MIMES else "attachment"
+        headers = {
             "Content-Disposition": f'{disposition}; filename="{rec.filename}"',
             "X-Content-Type-Options": "nosniff",
-        },
-    )
+        }
+    return StreamingResponse(it, media_type=rec.mime, headers=headers)
 
 
 @router.delete("/{file_id}", status_code=204)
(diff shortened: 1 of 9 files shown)


prakashUXtech:
Rebased onto current dev (was 403 behind). One conflict at the top of `uploads/service.py`: dev's docstring now mentions thumbnail generation, so I kept dev's docstring and the branch's change note above it.

Verified on the rebase: `tests/uploads/` and `tests/cloud/uploads/` pass 122/122 with the ee group installed, including both SVG serving tests. Ruff clean. SVG is still absent from `DEFAULT_ALLOWED_MIMES` on dev, so the gap this closes is real.

PR #1481, opened 2026-06-17
Fix/connectors oauth

AMRITESH240304:
### The Problem
Three disconnected state systems were causing conflicting connection statuses between the **Tools tab**, **Data Sources tab**, and the **Agent**:
1. **OAuth TokenStore** — Had valid tokens 
2. **Connector Registry StateStore** — Had no configuration 
3. **Naming Mismatch** — The backend expected `"drive"` while the frontend sent `"google_drive"` 

The OAuth flow successfully saved tokens to the `TokenStore` but never registered them with the `ConnectorRegistry`. As a result, both the Agent and the Data Sources tab reported *"no connectors connected"* despite valid tokens existing. Additionally, expired or revoked refresh tokens triggered a fail-delete loop during status queries.

---

#### 1. Backend Changes (`pocketpaw`)

* **Connector Naming Alignment:** * Renamed `drive` ➡️ `google_drive` in the YAML definition, registry native adapter set, and adapter dispatch logic to unify naming conventions across the frontend and OAuth systems.
* **OAuth ➡️ Registry Reconciliation:**
    * Introduced `_auto_connect_if_oauth()` to automatically sync valid OAuth tokens into the connector registry during status queries. 
    * It safely invokes `get_valid_token()` (handling refresh attempts) before updating the registry, preventing the fail-delete loop when refresh tokens are dead or revoked.
* **Endpoint Synchronization:**
    * `GET /connectors` and `GET /connectors/{name}/status` now both trigger the auto-reconciliation routine.
    * `GET /tools` syncs tokens and accurately identifies expired states (yielding `"expired"` if no refresh path remains).
    * The OAuth callback routine now explicitly auto-connects the registry immediately after saving tokens.
* **New Endpoints Added:**
    * `GET /connectors/{name}/actions` — Fixed a `404` error the agent encountered when requesting the action list.
    * `POST /oauth/integrations/disconnect` — Securely deletes tokens and cleanses the corresponding connector from the registry.

#### 2. Frontend Changes (`paw-enterprise`)

* **State Alignment:** Integrated UI handling for the new `"expired"` status badge, updating state mappings inside the Data Panel layout.
* **Toggle Synchronization:** Synchronized status badges with toggle logic, preventing `"connected"` badges from displaying if a category toggle is explicitly turned off.
* **Component Cleanup:** Added an explicit disconnect action calling the new backend endpoint to safely remove active individual OAuth connections.
* **Onboarding Fault Tolerance:** Gracefully handles instances where the cloud connectors endpoint might be missing (e.g., local or OSS mode).

---

### Files Changed

* `connectors/drive.yaml`
* `connectors/registry.py`
* `connectors/adapters/gdrive.py`
* `api/v1/connectors.py`
* `api/v1/tools.py`
* `api/v1/oauth_integrations.py`
* `src/lib/components/os/PocketDataPanel.svelte`
* `src/lib/core/pockets/api.ts`
* `src/lib/core/runtime/api.ts`
* `tests/`

### DIFF
--- a/connectors/drive.yaml
+++ b/connectors/drive.yaml
@@ -4,7 +4,7 @@
 # Auth is OAuth 2.0 (bearer token) — either from the credential broker at
 # dispatch time or from GOOGLE_OAUTH_TOKEN for local dev.
 
-name: drive
+name: google_drive
 display_name: Google Drive
 type: knowledge
 icon: cloud
--- a/src/pocketpaw/api/v1/connectors.py
+++ b/src/pocketpaw/api/v1/connectors.py
@@ -128,6 +128,121 @@ class ConnectorStatusResponse(BaseModel):
 # ── Routes ───────────────────────────────────────────────────────────────────
 
 
+# Map OAuth service names to connector names for auto-connect reconciliation.
+_OAUTH_SVC_TO_CONNECTOR: dict[str, str] = {
+    "google_gmail": "gmail",
+    "google_calendar": "gcalendar",
+    "google_drive": "google_drive",
+    "google_docs": "gdocs",
+    "spotify": "spotify",
+}
+
+
+async def _auto_connect_if_oauth(
+    reg: ConnectorRegistry,
+    connector_name: str,
+    pocket_id: str,
+) -> bool:
+    """Check if *connector_name* has valid OAuth tokens and register it.
+
+    Returns True if the connector was newly connected by this call, False
+    if it was already connected or no OAuth tokens exist.
+
+    This bridges the gap between the OAuth TokenStore (used by the Tools
+    tab) and the connector registry's state store (used by the Data Sources
+    tab and agent tools). Existing tokens saved before the auto-connect
+    feature get picked up on the first status query.
+    """
+    # Find the OAuth service name that maps to this connector.
+    svc = None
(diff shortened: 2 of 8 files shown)


prakashUXtech:
Reviewed this properly — oriented in the connectors/OAuth subsystem and exercised the change with running code. The core fix is solid: calling `get_valid_token()` before `reg.connect()` genuinely dodges the old set-then-delete loop (verified — a revoked token now writes no state), and the happy-path reconciliation works.

I pushed two commits to help it land:

1. **Tests** (`tests/connectors/test_pr1481_oauth_reconcile.py`) — the change shipped without coverage. Three tests: the auto-connect happy path, the fail-delete-avoidance guarantee, and the rename migration.
2. **A migration fix** — the `drive` → `google_drive` rename keys existing persisted state under the old name, so a user who connected before the rename would have their connection silently orphaned (`definition_missing`). The registry now runs a one-time idempotent rename migration on construction, so existing `drive` connections carry over.

One thing from my first read I'm **withdrawing**: I flagged the hardcoded `"default"` pocket in the OAuth callback as a multi-tenancy gap — but these `/api/v1/connectors` + `oauth_integrations` routes are the single-tenant OSS surfaces (per `state_provider.py`), and the cloud path scopes its own rows separately. `"default"` is correct here.

Full connector suite passes (158) including the new tests. Nice fix on a genuinely confusing three-state bug.

PR #1509, opened 2026-06-19
Sites: workspace-scope Fabric object types (sovereign discovery foundation)

prakashUXtech:
First foundational slice of sovereign zero-setup discovery (SZD-2). Fabric object *types* were a global catalog — `define_type` had no `workspace_id` and `fabric_object_types` had no `workspace_id` column — so a type defined in one tenant's workspace was visible to all. For the discovery feature (and tenant sovereignty generally), discovered types must stay private to the workspace that owns them.

## What changed
- `workspace_id` on `define_type` + the `ObjectType` model + the `fabric_object_types` table (idempotent additive migration; legacy rows keep `workspace_id = NULL` = global, for backward compatibility).
- Every object-type read is workspace-scoped: `get_type_by_name`, `list_types`, `stats`, dedup, and the unique index (now `(workspace_id, LOWER(name))`, replacing the global name index).
- `workspace_id` threaded through `ingest_records` / `ensure_type` so connector ingestion mints types in the right workspace.

## Verified
- New `tests/test_fabric_type_scoping.py` (8 tests): workspace A can't read/reuse workspace B's types; legacy NULL stays global; same-name types coexist across workspaces; migration idempotency; backfill.
- 6 existing fabric tests updated to the scoped semantics (they encoded the old global-catalog behavior); reviewed and confirmed legitimate.
- Re-ran the touched suites: 179 fabric tests pass; the migration runs twice cleanly.

## Notes
- Reverses the earlier W4a choice to leave types unscoped — superseded by the 2026-06-19 decision to scope types per workspace.
- Behavior change: an owner's empty (object-less) type is now visible to its owner — it's first-class tenant data now, not derived from object rows.
- Two non-blocking review follow-ups: a test helper (`_seed_customers`) still defines an unscoped type, so one stats assertion passes for the wrong reason; and a one-line comment on the backfill/dedup row-factory ordering. Neither affects correctness.

Foundation for SZD-5a (the `_fabric_objects` proposal type). Part of the sovereign zero-setup discovery slice — design/PRD/tasks in `docs/design/drafts/2026-06-19-sovereign-zero-setup-discovery*`.

PR #1510, opened 2026-06-19
Sites: fabric.objects + fabric.query ripple sources (render Fabric in a Pocket)

prakashUXtech:
SZD-1, stacked on #1509. A Pocket couldn't render Fabric data — only `workspace.pockets` and `workspace.members` ripple sources existed; nothing resolved a Fabric ObjectType into widget rows. The sovereign zero-setup discovery starter-Pocket needs that.

## What changed
- `fabric.objects` source — resolves an ObjectType (by `type_id` and/or `type_name`) → widget rows, optional `limit`/`offset`. Properties spread to the top level for column-oriented widgets; id/type/provenance on reserved keys.
- `fabric.query` source — the filtered variant (adds a `FabricQuery` filters bag). Trivial given FabricQuery already supports filters, so included rather than deferred.
- Both workspace-scoped: `workspace_id` comes from the server-built resolve context (never the spec), threaded to `FabricStore.query`, which applies its W4a tenant guard. A missing workspace returns `[]` — never an unscoped read. Rows capped at 500 to keep resolved specs lean.
- Auto-registered (the `@register` decorators live in `ripple_sources.py`, already eager-imported at startup) — no wiring change.

## Verified
- New `tests/cloud/test_fabric_objects_source.py` (5 tests): workspace A resolves only A's objects, never B's (and the mirror from B). Re-ran with the resolver regression suite: 19 pass.

## Notes
- No paw-enterprise change needed: the resolver replaces source markers with plain row lists before the spec reaches the frontend, so these render like any other widget data.
- Follow-up (SZD-6, the starter-Pocket slice): teach the pocket-authoring skill that `fabric.objects`/`fabric.query` are bindable sources so the author emits them.

Stacks on #1509. Part of sovereign zero-setup discovery.

### DIFF
--- a/ee/pocketpaw_ee/cloud/ripple_sources.py
+++ b/ee/pocketpaw_ee/cloud/ripple_sources.py
@@ -6,6 +6,18 @@ live here, next to the entities they read.
 
 Tenancy rule (from CLAUDE.md ee/cloud rule 7): every Mongo read MUST
 scope by ctx.workspace_id.
+
+Updated 2026-06-19 (SZD-1 — fabric.objects ripple source): added two
+Fabric-backed sources so a rippleSpec can render discovered Fabric data, not
+just pockets/members. ``fabric.objects`` resolves an ``ObjectType`` (by
+``type_id`` and/or ``type_name``) into widget rows; ``fabric.query`` is the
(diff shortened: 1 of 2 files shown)


prakashUXtech:

PR #1511, opened 2026-06-19
Sites: _fabric_objects Instinct proposal type (gated ontology writes)

prakashUXtech:
SZD-5a, stacked on #1509. The Instinct gate had no way to propose "create these Fabric objects" — it was sealed to `_pocket_write` / `_code_change` / `_external_action` / `_belt_plan`. Sovereign zero-setup discovery needs to stage a proposed Fabric ontology for human review and only write on approve. This adds that proposal type, mirroring the shipped `_external_action` pattern exactly.

## What changed
- New `ee/pocketpaw_ee/cloud/fabric_proposals/` — `propose_fabric_objects` (schema-1 blob: object_types + objects + links + the tenancy `workspace_id`; anchors `pocket_id=workspace_id`; opens the `agent.proposed` chain) + `execute_approved_fabric_objects` (reuses the idempotent, workspace-scoped `ingest_records`; links deduped on `(from,to,link_type)`; single `_fail` chokepoint; owns `decision.completed` on approve).
- `router.py` — `_assert_fabric_objects_workspace` guard, called in all four paths (approve / reject / bulk-approve / bulk-reject) before any mutation; approve→executor, reject→router-owned chain close.

## Verified
- New `tests/cloud/test_fabric_objects_gate.py` (14 tests): all four cross-workspace paths reject (403, no write); re-approve doesn't duplicate (created=0 / updated=2 / links=0); chain is `[agent.proposed, human.corrected, decision.completed]`, one terminal per path; a schema-999 blob is refused. Regression: external_action_gate (27) + fabric_ingest + instinct_decision_events (12) green.
- Independently re-ran the 14 tests + confirmed the guard is invoked in all four router dispatch functions; distrust review returned SPEC_PASS, no cross-tenant path.

## Notes
- Links reference endpoints by `(source_connector, source_id)` natural key (Fabric ids don't exist at propose time); resolved post-ingest, a link to a missing endpoint is skipped (logged), not failed — the right default for a self-contained ontology proposal.
- No caller wires `propose_fabric_objects` yet — that's the discovery surface (SZD-6). The helper is exported, ready.

Stacks on #1509. Pairs with `_pocket_create` (SZD-5b, next). Part of sovereign zero-setup discovery.

PR #1512, opened 2026-06-19
Sites: _pocket_create Instinct proposal type (gated Pocket creation)

prakashUXtech:
SZD-5b, stacked on #1511. The second discovery gate type: propose a starter Pocket for review; approve creates it. Mirrors `_fabric_objects` (#1511) and `_external_action` exactly.

## What changed
- New `ee/pocketpaw_ee/cloud/pocket_proposals/` — `propose_pocket` (schema-1 `_pocket_create` blob: `pocket_spec={name, rippleSpec, templateSlug?}` with `workspace_id` + `user_id` as SEPARATE top-level fields so the correction flow can't move tenancy/owner — tenancy keys inside `pocket_spec` are stripped defensively; anchors `pocket_id=workspace_id`; opens the `agent.proposed` chain) + `execute_approved_pocket_create` (schema + workspace/owner + idempotency guards → `CreatePocketRequest.model_validate` → `pockets.service.create`; owns `decision.completed` via a single `_fail` chokepoint).
- `router.py` — `_assert_pocket_create_workspace` called in all four paths alongside the fabric guard; approve→executor, reject→router-owned close.

## Verified
- New `tests/cloud/test_pocket_create_gate.py` (18 tests): approve creates a Pocket owned correctly; 4-path cross-workspace gate (all 403, PENDING, nothing created); `model_validate` resolves camelCase `rippleSpec` + fails invalid specs cleanly; schema mismatch refused; idempotent; chain `[agent.proposed, human.corrected, decision.completed]`, one terminal per path; tenancy never in the editable spec.
- Regression: fabric_objects_gate + external_action_gate + bulk_actions + approval_security = 60 pass. Independently re-ran the 18 tests + confirmed the guard fires in all four router functions.

## Notes
- The empty-`workspace_id` guard short-circuit mirrors `_fabric_objects` (inherited from `_external_action`) — not exploitable (propose requires non-empty; executor fails-closed), but worth folding into any future fail-closed guard hardening.

Stacks on #1511. Completes the two-gate-type pair. Part of sovereign zero-setup discovery.

### DIFF
--- a/ee/pocketpaw_ee/cloud/pocket_proposals/__init__.py
+++ b/ee/pocketpaw_ee/cloud/pocket_proposals/__init__.py
@@ -1,0 +1,32 @@
+# ee/cloud/pocket_proposals/__init__.py — the gated starter-Pocket-create Instinct
+# proposal type (a peer of ``_fabric_objects`` / ``_pocket_write`` /
+# ``_code_change`` / ``_external_action`` / ``_belt_plan`` / ``_artifact_change``).
+# Created: 2026-06-19 (SZD-5b — _pocket_create proposal type) — a proposed starter
+#   Pocket (a rippleSpec + name, optional template slug) staged for human review
+#   through the Instinct gate. "Sovereign zero-setup discovery" infers a starter
+#   Pocket for a tenant and proposes "create this Pocket"; a human approves or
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1513, opened 2026-06-19
Sites: StructuredShapeDigester — reverse-engineer an ontology from sampled records

prakashUXtech:
SZD-3. The discovery pipeline samples a connector's records and must turn them into a candidate ontology. This is the pure-logic digester: structured records → an `OntologyDraft` (object types, objects, links, and per-type primary-key + projection), each inference scored with a confidence.

## What changed
- New `ee/pocketpaw_ee/discovery/` — a `Digester` Protocol + `StructuredShapeDigester` (no DB/network/async): infers object types from record shapes, property types by value-type voting (str/number/bool/date incl. ISO strings + native datetime), the primary key (named-field priority `id`/`_id`/`<type>_id` → high-cardinality-unique fallback → none), and FK links from field values matching another type's key set.
- `OntologyDraft` carries the inferred `source_id` field + field projection so a `FabricMapping` (what `ingest_records` needs) builds directly via `to_fabric_mapping_kwargs()`.

## Verified
- New `tests/ee/test_structured_shape_digester.py` (14 tests): type + property-type inference; native-datetime→date; all three key-inference paths; FK links; FabricMapping round-trip; and the three degradation cases — empty, nested/polymorphic (no crash), no-clear-key (objects-only, low confidence). Re-ran independently: 14 pass.

## Notes (for SZD-6, the consumer)
- `source_id_field` is `""` when no key was inferred — the proposal step must gate on `key_confidence` before building a FabricMapping (a keyless type would skip every record).
- Links reference endpoints by `(type_name, source_id)`, not Fabric ids (the draft predates persistence) — resolved to real object ids after ingest.

Independent of the other slices (off dev). Feeds SZD-4 (the orchestrator). Part of sovereign zero-setup discovery.

### DIFF
--- a/ee/pocketpaw_ee/discovery/__init__.py
+++ b/ee/pocketpaw_ee/discovery/__init__.py
@@ -1,0 +1,29 @@
+# pocketpaw_ee/discovery/ — sovereign zero-setup ontology discovery (digesters).
+#
+# Created: 2026-06-19 (SZD-3 / feat/szd-3-digester) — the digester layer for
+# sovereign zero-setup discovery. A Digester samples a connector's records and
+# reverse-engineers a candidate ontology (an OntologyDraft) that downstream code
+# turns into FabricMapping objects (for fabric_ingest.ingest_records) and a
+# fabric-objects proposal (object_types + objects + links).
+#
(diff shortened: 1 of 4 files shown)


prakashUXtech:

PR #1514, opened 2026-06-19
Sites: DiscoveryRun orchestrator — sample connectors (local path) -> ontology draft

prakashUXtech:
SZD-4, stacked on #1513. Ties sampling→digest: a `DiscoveryRun` samples a workspace's bound connectors and runs the SZD-3 digester to produce an `OntologyDraft`.

## What changed
- New `ee/pocketpaw_ee/discovery/run.py` — `DiscoveryRun.run(workspace_id, connector_ids, opts)`. Reads via the workspace-scoped, pocket-less path: `registry.ensure_connected(name, "ws:<workspace_id>")` then `adapter.execute(read_action, params)`, reading records from `ActionResult.data`. **Never calls `sync()`** (which needs a pocket and returns only counts). Drives the tenant's LOCAL registry directly (the cloud LOCAL-exec seam 503s).
- Read actions: explicit `opts.read_actions` map, or auto-select of GET/HEAD + auto-trust (safe-reads-only) actions — never mutating ones.
- Deterministic digest for slice 1 (no LLM); `opts.refine=True` raises `NotImplementedError` with a note that a refine pass must use the on-box/tenant model, never a cloud model on tenant raw data.
- Sampling capped (200/connector); draft carries a "based on N of M" meta + per-connector rollup. Permission degrades to allow-all with `permission_enforced: false` until #1498 lands.

## Verified
- New `tests/ee/test_discovery_run.py` (8 tests): the read goes through `ensure_connected("ws:...")` + `execute`, and asserts `sync()` is NEVER called; sampling cap; empty connector → empty draft; unresolved connector skipped; mutating actions excluded from auto-select; permission filter; refine NotImplemented. Re-ran with the digester suite: 22 pass.

## Notes
- No single canonical "list_records" action across connectors → auto-select is a heuristic (GET/HEAD + auto-trust); a connector whose only read is confirm-trust must be supplied via `opts.read_actions`. The discovery surface (SZD-6) configures this.

Stacks on #1513. Feeds SZD-6 (wires the draft into the gate proposals). Part of sovereign zero-setup discovery.

### DIFF
--- a/ee/pocketpaw_ee/discovery/__init__.py
+++ b/ee/pocketpaw_ee/discovery/__init__.py
@@ -6,8 +6,11 @@
 # turns into FabricMapping objects (for fabric_ingest.ingest_records) and a
 # fabric-objects proposal (object_types + objects + links).
 #
-# Pure logic: no DB, no network, no async. The first impl is
-# ``StructuredShapeDigester`` (structured ``{type: [records]}`` input).
+# The digester is pure logic (no DB, no network, no async); the first impl is
+# ``StructuredShapeDigester`` (structured ``{type: [records]}`` input). The
+# ``DiscoveryRun`` orchestrator (SZD-4) ties sampling → digest together: it reads
+# a workspace's bound connectors via the pocket-less ``ensure_connected(name,
+# "ws:<workspace_id>")`` path and feeds the sampled records to the digester.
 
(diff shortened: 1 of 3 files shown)


prakashUXtech:

PR #1521, opened 2026-06-20
Sovereign zero-setup discovery slice 2: unstructured exhaust + governed rules

prakashUXtech:
## What this is

Slice 2 of sovereign zero-setup discovery. Slice 1 (merged in #1515) discovers a Fabric ontology and a starter Pocket from a tenant's **structured** connectors. Slice 2 adds the two pieces that were deferred:

- **KbCompileDigester** discovers from **unstructured** exhaust (ticket bodies, email and chat text). It compiles the text into a kb-go wiki on the tenant's own box, reads the compiled articles back, and infers an ontology from them. It plugs into the existing `Digester` interface, so slice-1's run/propose/approve flow needed no rework.
- **Rules-discovery** reverse-engineers draft governed rules (the Edra-grade layer) from a tenant's Instinct exhaust, specifically its correction history, and stages them as `_instinct_rule` proposals through the gate. A human reviews, edits, and approves; the approved rule is persisted as a workspace-scoped, owned, editable record. This generalizes the existing "same field corrected 3 times" soul-memory promotion into a structured, gate-governed rule.

The whole thing runs on the tenant's box. Raw exhaust never leaves it.

## The sovereignty guarantee

kb-go has two ingest paths: a keyless on-box one (`convo ingest`) and a cloud one (`kb ingest` / `kb build`) that POSTs text to Anthropic. KbCompileDigester only ever calls the keyless path. That constraint is enforced as a test, not left to code review. A tripwire asserts `kb ingest` and `kb build` are never invoked across a digest, and the E2E re-asserts it across the full run.

## Design decision made without the captain (flagged for review)

Rules-discovery had no prior data model. No rule store, no enforcement path. With the captain out, rather than stall I made the call and wrote it up in `docs/design/szd-slice2-build-plan.md` (S2-R0):

- The **rule model** reuses `CelExpression` for the `when` and the existing `require_approval` / `notify` / `block` action literal. No new primitives.
- **Persistence** is an EE Beanie `RuleDoc`, alongside the gate types.
- **Scope boundary:** slice 2 takes a discovered rule all the way to persisted, owned, and editable (the Edra parity, white-box owned rules that never left the box), plus a `get_active_rules(workspace_id)` read API. Wiring those rules into the **live gate dispatch** is left for slice 3. Changing how the gate evaluates every action is high-stakes and deserves a look before it ships. The read API is the seam.

If you'd have modeled the rule differently, this is the place to push back.

## What's in it

| Slice | What |
|-------|------|
[...]

PR #1527, opened 2026-06-21
Finish discovery: trigger + on-box categorization + refine + edit-in-review (F1-F4)

prakashUXtech:
## What this is

The first half of finishing sovereign zero-setup discovery end to end. Slices 1 and 2 (in #1515 and #1521) built the discovery engine, but a customer couldn't actually reach it, it degraded on real messy text, and review was all-or-nothing. This PR closes those gaps.

**Stacked on #1521** (slice 2). Review and merge that first; this targets the slice-2 branch so the diff shows only the finish work.

## What's in it

- **F1 — the front door.** A new `POST /api/v1/cloud/discovery/run` endpoint enumerates the workspace's enabled connectors and fires a discovery run, returning 202 with a run id. The proposals surface as pending Instinct actions the review panel already polls. Until now discovery could only be started from a script.
- **F2 — works on real messy text.** `kb convo ingest` tags every article "conversation", so unstructured discovery used to collapse to one objects-only type. This routes exhaust through kb's `prepare → on-box model → accept` so real tickets and emails get domain categories (SupportTicket, RefundRequest). When no on-box model is configured it falls back to the old deterministic path, and it logs which path ran so a silent regression is visible.
- **F3 — the on-box refine pass.** `opts.refine` used to raise. It now cleans the inferred ontology (merge, rename, drop spurious links) using the tenant's own Ollama model. It fails closed on sovereignty (never a cloud call) and soft on availability (Ollama down returns the deterministic draft, stamped `refine=unavailable`, never an error).
- **F4 — edit before you approve.** A new `PATCH /instinct/actions/{id}/proposal` lets a reviewer tighten a rule's CEL, rename a type, drop a link, or fix a key on a pending proposal, then approve the edited version. It re-validates per blob type, records the change as a correction (feeding the learning loop), and never flips status.

## Sovereignty

F2 and F3 both call a model, and both hard-pin `force_provider="ollama"` so a tenant cloud key can never route raw text to Anthropic. That pin is asserted by mechanical tests (the resolved client has `api_key is None`). The existing `kb ingest`/`kb build` tripwire is extended and now also lives as a runtime assertion inside the `_kb` seam, not just in tests.

## Security (F4)

A discovery proposal carries the tenant id in two places: the top-level `workspace_id` and `rule_spec.scope.workspace_id` (the executor scopes the saved rule by the second one). The edit endpoint pins both from the original blob and 403s any client attempt to change either. Both pins have dedicated tests.

## Testing

All slices ship unit + integration tests; the discovery suite stays green. ruff clean, import-linter green, mypy introduces no new real errors.

Does not merge on green. Captain reviews.

### DIFF
--- a/docs/design/szd-finish-build-plan.md
+++ b/docs/design/szd-finish-build-plan.md
@@ -1,0 +1,227 @@
+All anchors confirmed across all four maps (the kb-go category hardcode is at `convo.go:625` exactly as Map C states, with the `"conversation"` filter also at `:766`/`:820`). I have everything I need to write the plan.
+
+---
+
+# Sovereign Zero-Setup Discovery — Finish-Build Plan (FULL polish)
+
+**Worktrees:** pocketpaw integration = `…/pocketPaw/.claude/worktrees/szd2-int` (branch `feat/szd-slice2-discovery`); paw-enterprise UI = `…/paw-enterprise/.claude/worktrees/szd2-ui`. kb-go lives at `…/paw-workspace/kb-go` (separate repo).
+
+**Six gaps → six slices:** F1 trigger (#1), F2 categorization (#3), F3 refine (#5), F4 edit-in-review (#4), F5 real-connector E2E (#6), **F6 live rule enforcement (#2 — captain-review gate)**.
+
+---
+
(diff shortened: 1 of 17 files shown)


prakashUXtech:

PR #1528, opened 2026-06-21
Finish discovery: live rule enforcement + end-to-end proof (F6 + F5)

prakashUXtech:
## What this is

The piece that makes a discovered rule actually do something. Until now an approved rule was saved as an owned record but never consulted, so "refunds over $500 need approval" got approved and then nothing happened. This wires approved workspace rules into the live action gate.

**This is the high-stakes slice, isolated for your review.** It touches the gate that evaluates every action. It is default-OFF, never mutates the template, and fails safe on every error path. **Stacked on the finish-core PR** (F1-F4); review that first.

## The approach (the design you signed off on)

A discovered rule and a template `InstinctRule` already share the same shape: same `when` (CEL) and same `action` (`require_approval`/`notify`/`block`). So enforcement is not a new evaluation path. Each active discovered rule is converted into an `InstinctRule` and run through the exact composer machinery that already ships. No new verdict logic, no new outcome mapping.

- **Where:** `gate_action` in `instinct_dispatch.py`, just before `resolve_instinct`. It merges discovered rules into a *copy* of the template (the template object is never mutated, so every other reader is unaffected). The pure composer is untouched, so the import-linter purity contract still holds.
- **The flag:** `instinct_enforce_discovered_rules`, default False. When off, `get_active_rules` is never called and the composer gets byte-identical arguments. A test asserts the read is never even attempted on the default path.
- **Fail-safe:** a `get_active_rules` read failure falls through to the template (fail open, a DB hiccup can never block an action). A discovered rule whose CEL errors is dropped by a guarded probe, never a 404, never a silent block. Discovered rules are inferred and lower-trust, so they fail open per rule; template rules keep their loud-fail. A discovered rule can only ever add a block or escalate, never relax the template floor.
- **Outcomes (free):** `block → BLOCK`, `require_approval → ESCALATE` (human gate / the existing triage lanes), `notify → existing notify path`.

## F5 — the end-to-end proof

A full integration test drives the finished stack as one flow: trigger an unstructured run, get domain categories from the mocked on-box model, stage three proposals, edit a rule, approve, then fire a governed action and watch it escalate with the flag on and proceed with the flag off. It asserts no cloud LLM call happened, `kb ingest`/`kb build` were never invoked, and the deterministic floor holds when the model is down.

A short by-hand runbook (`docs/design/szd-finish-live-runbook.md`) covers the truly-live run against a real connector and a real Ollama.

## Testing

The enforcement test file covers backward-compat, behavior change, inertness, precedence, and every fail-safe path. The relevant gate files pass clean in isolation. ruff clean, composer purity contract green.

Does not merge on green. Captain reviews this one closely.

### DIFF
--- a/docs/design/szd-finish-live-runbook.md
+++ b/docs/design/szd-finish-live-runbook.md
@@ -1,0 +1,165 @@
+<!--
+docs/design/szd-finish-live-runbook.md
+Created: 2026-06-21 (F5 / feat/szd-finish-enforce). A hands-on checklist for the
+captain to run the FINISHED Sovereign Zero-Setup Discovery feature against real
+infra once — a real connector, a real on-box model, the live gate — and confirm
+by eye what the F5 automated E2E (tests/ee/test_szd2_finish_e2e.py) proves with
+mocks. Pairs with that test: the test pins the logic deterministically; this
+runbook is the truly-live smoke pass that can't be automated cheaply (it needs a
+running Ollama + a bound connector). No PII — use non-sensitive seed data only.
+-->
+
+# Sovereign Zero-Setup Discovery — live run runbook
+
+A by-hand smoke pass for the finished feature (F1 trigger → F2 categorize → F3
+refine → F4 edit → approve → F6 enforce). The automated E2E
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1529, opened 2026-06-22
Close two discovery follow-ups: run_id correlation + edit-test coverage

prakashUXtech:
Two small follow-ups left over from the Sovereign Zero-Setup Discovery finish. Stacked on #1528 (targets feat/szd-finish-enforce).

## 1. Correlate the trigger run_id with the proposals

The F1 trigger minted a run_id for its 202 response, but the orchestrator minted its own separate id to tag the staged proposals. So a client got back a run_id that matched nothing it could find later.

`run_discovery_and_propose` now takes an optional `run_id`. When passed, it tags the proposal markers with that id instead of minting a fresh one. The F1 service mints the id once and hands it both to the 202 response and to the orchestrator, so the dispatch token a client holds actually points at the proposals it produced. Leaving the arg off mints as before, so nothing that called the orchestrator directly changes.

## 2. Happy-path edit tests for the other two blob kinds

The PATCH edit endpoint handles three discovery blob kinds, but only the rule kind had a happy-path edit test (that's where the dual-tenancy risk lives). Added the missing two:

- rename a pending `_pocket_create` proposal (re-validates CreatePocketRequest, stays pending, tenancy/owner pinned)
- drop a spurious link on a pending `_fabric_objects` proposal (re-validates the lists, stays pending, workspace_id pinned)

## Checks

- `tests/ee/test_instinct_proposal_edit.py test_discovery_propose.py test_discovery_rules_propose.py test_szd2_finish_e2e.py` — 28 passed
- adjacent guard run (`test_szd2_e2e.py`, `test_discovery_run.py`) — 9 passed
- ruff check + format clean, import-linter 25/0

The finish E2E now also asserts the 202 run_id equals the proposals' marker run_id, so the correlation is covered end to end.

### DIFF
--- a/ee/pocketpaw_ee/cloud/discovery/service.py
+++ b/ee/pocketpaw_ee/cloud/discovery/service.py
@@ -1,4 +1,10 @@
 # Discovery — service (cloud 4-file rule §5).
+# Updated: 2026-06-22 (feat/szd-finish-followups) — the trigger now mints the
+#   ``run_id`` ONCE and threads it BOTH into the 202 response AND into
+#   ``run_discovery_and_propose(run_id=...)``, so the staged proposals' discovery
+#   markers carry the SAME id the 202 handed the client. The previously-distinct
+#   "optimistic dispatch token vs. internal proposal run_id" split is gone — the
+#   client can now correlate its 202 ``run_id`` to the proposals it produced.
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1533, opened 2026-06-22
Make the discovery model lane configurable

prakashUXtech:
Discovery's categorize and refine passes used to hard-pin the model to a local Ollama, with no way to change it. That is the right default for tenants who can't let anything leave the box, but most businesses are fine running discovery on a cloud model, and forcing local on everyone just makes it slower and the categories weaker for no reason.

This adds a setting so the posture is a choice instead of a constant.

## What changed

New setting `discovery_sovereign_model` (env `POCKETPAW_DISCOVERY_SOVEREIGN_MODEL`), default `True`:

- `True` (default): the model is pinned to the on-box Ollama, exactly as before. `api_key is None`, nothing leaves the box. Regulated and sovereign tenants get the safe default with no action.
- `False`: discovery resolves the workspace's configured provider through `resolve_llm_client(settings)` with no force, so anthropic / openai / openai_compatible / gemini / litellm are honored and a cloud model is allowed. This is the tenant opting in to send discovery's inferred data shape to their own configured model.

The provider decision now lives in one place (`_refine.resolve_on_box_descriptor`), and both the categorize (F2) and refine (F3) lanes go through it.

## What did not change

- Default behavior is identical to before: sovereign-local. The existing F2/F3 tests that assert the default still pass untouched.
- The kb ingest/build tripwire stays unconditional. That path POSTs raw tenant text to the Anthropic KB API and is never correct, sovereign or not, so it stays hard-blocked regardless of this setting.
- Wiring the configured agent backend (e.g. the Claude Code SDK loop) as the discovery one-shot is out of scope here. The agent loop isn't a chat-completions endpoint, so that's a separate, larger piece of work. Left a follow-up note in `_refine.py`.

## Tests

New `tests/ee/test_discovery_model_lane.py`:

- default True forces ollama, `api_key is None`
- False with a configured provider resolves that provider, not forced ollama
- False with nothing resolvable degrades gracefully (refine returns the deterministic draft flagged unavailable, categorize falls back to convo ingest) and never raises or silently leaks to cloud
- the tripwire holds under both postures

All green:

```
37 passed   # model-lane + refine + kb_compile categorize/digester
```

ruff check + format clean, import-linter contracts all kept.

### DIFF
--- a/ee/pocketpaw_ee/discovery/_refine.py
+++ b/ee/pocketpaw_ee/discovery/_refine.py
@@ -1,33 +1,58 @@
-# pocketpaw_ee/discovery/_refine.py — the on-box REFINE pass for discovery (F3).
+# pocketpaw_ee/discovery/_refine.py — the discovery model-lane resolver + REFINE
+# pass (F3).
 #
 # Created: 2026-06-21 (F3 / feat/szd-finish-core) — un-stubs ``opts.refine`` on
 # ``DiscoveryRun``. After the DETERMINISTIC digest produces an ``OntologyDraft``,
-# this pass asks the tenant's ON-BOX model (Ollama) to clean it up: merge
-# near-duplicate types, rename to canonical names, and drop spurious links. It
-# sends the model the DRAFT shape (type names, property names, sample
-# summaries already in the draft) — NOT raw tenant exhaust.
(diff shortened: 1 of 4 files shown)


prakashUXtech:

Issue #1162, opened 2026-05-21
Spike: verify the agent actually solved the original problem, not just that it finished


prakashUXtech:
This is one of six issues distilling Spotify's background-coding-agent posts onto our agent primitives. This one covers the VERIFICATION / OUTCOME primitive: Instinct (the decision-to-action-to-outcome loop), Pawprints (audit), the cloud Mission Control Outcomes strip.

Spotify closes their Part 2 post with an honest admission: "even if we make it to a merged PR, how do we know if it actually solved the original problem?" A merged PR or a completed task is an output, not an outcome. Spotify runs LLM-as-judge on the diff, but that judges the diff, not whether the original problem is solved.

Current state: PocketPaw records Instinct `Action`s and cloud `Task`s. A "Nudge" is not a separate model; it is a `Task` awaiting approval or an Instinct `Action`. The Instinct `Action` already carries an `outcome` field that `mark_executed()` writes, but today it holds a free-text "what happened" string rather than a verified "did this solve the problem" verdict. The cloud Outcomes strip is real, but it only counts `ActionStatus` lifecycle states, and its own DTO docstring calls those numbers unchecked. So the loop is named and partially wired. The RFC's job is to turn `Action.outcome` into a checked verdict and add a success-criteria field to `Task` (and `TaskSpec`), which have none today.

Proposal: file this as a spike / RFC-scoping issue rather than an implementation task. Scope an RFC for outcome verification. The success criteria captured at task creation (the field from the planning-agent sibling issue) get checked after the action completes, and the result lands in Pawprints as a verified outcome instead of a bare completion. Use LLM-as-judge for soft or subjective cases and deterministic checks where the criteria allow it. This is what makes Instinct concrete, and it answers a question Spotify's own internal tooling openly cannot.

This depends on the planning-agent sibling issue (filed in parallel). Success criteria have to be captured at intake to be checkable at the end.

Sources:
- https://engineering.atspotify.com/2025/11/spotifys-background-coding-agent-part-1
- https://engineering.atspotify.com/2025/11/context-engineering-background-coding-agents-part-2

prakashUXtech:
Foundation done in #1169, merged to dev. LLM-as-judge scoring tracked separately in #1168.

PR #1538, opened 2026-06-23
feat(deep_work): stamp an outcome verdict on completed tasks behind a flag (SVL-1)

prakashUXtech:
## What this does

When `deep_work_verify_loop_enabled` is on, a deep_work task that finishes successfully now has its result checked against the `success_criteria` captured at intake, and the verdict is recorded on the task's metadata and logged. The task still completes the same way it does today. This slice only observes and records the verdict; it does not change task status or requeue anything.

The flag defaults to off, so nothing changes for existing deployments.

## Why

The deep_work scheduler marks a task done the moment the agent stops, without checking whether the result actually met its criteria. The verifier for that check (`verify_outcome`, #1162) shipped and has sat idle with no caller. This wires it in as an observe-only first step, so we can see which "done" tasks would have failed their criteria before the loop acts on it.

First slice of the Self-Verifying Loop. Acting on a failing verdict (requeue) lands in SVL-2.

## Scope

- `instinct/verdict_provider.py` (new): `VerdictProvider` protocol + `DeterministicVerdictProvider` wrapping `verify_outcome`. The seam that lets an LLM-judge provider slot in later without touching the executor.
- `config.py`: `deep_work_verify_loop_enabled` (default False) and `deep_work_verify_max_requeues` (default 2, read by SVL-2).
- `mission_control/executor.py`: in the completion branch, when the flag is on, compute the verdict and stamp `metadata["verify_verdict"]`. Guarded so the flag-off path is unchanged. No status mutation.
- Tests: `test_deep_work_e2e.py` (flag-on stamps a matching verdict; flag-off leaves metadata untouched); `test_deep_work_scheduler.py::_make_task` gains `output`/`success_criteria` kwargs.

## Verification

`uv run pytest tests/test_deep_work_e2e.py tests/test_deep_work_scheduler.py tests/cloud/test_instinct_verification.py` -> 52 passed. Ruff clean. Import-linter OSS-core contract holds (no ee import).

## Out of scope

Requeue on failure (SVL-2), the no-progress detector (SVL-3), LLM-judge and source-trace verifiers, and any change to intake or success_criteria authoring.

### DIFF
--- a/src/pocketpaw/config.py
+++ b/src/pocketpaw/config.py
@@ -1,6 +1,13 @@
 """Configuration management for PocketPaw.
 
 Changes:
+  - 2026-06-23: Added the deep_work Self-Verifying Loop flags (SVL-1) —
+    ``deep_work_verify_loop_enabled`` (default False; kill-switch — when
+    False deep_work tasks complete exactly as before) and
+    ``deep_work_verify_max_requeues`` (default 2; the requeue bound read by
+    SVL-2, landed here so later slices don't touch config). SVL-1 only reads
+    the enable flag to stamp an observe-only OutcomeVerdict on a completing
+    task. Env: POCKETPAW_DEEP_WORK_VERIFY_* .
   - 2026-06-18: Added the four layered/learning Instinct gate defaults —
     ``instinct_approval_level`` (default "ASK", dormant),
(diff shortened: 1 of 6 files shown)


prakashUXtech:
Shipped to dev via the stacked rebase in #1540 — SVL-1 is commit 0b419c02 on dev. Closing since the change is merged; no separate merge needed.

PR #1539, opened 2026-06-23
feat(deep_work): requeue tasks that fail their success criteria (SVL-2)

prakashUXtech:
## What this does

Builds on #1538. When the verify loop is enabled and a deep_work task finishes but its result does not meet its `success_criteria`, the task is requeued instead of being marked done. The specific unmet criteria are fed back to the next attempt, the requeue is bounded by `deep_work_verify_max_requeues` (default 2), and a task that still fails after the budget is escalated to BLOCKED for human review (the status the manual-retry endpoint already accepts).

Stacked on #1538 (SVL-1). The flag defaults to off, so nothing changes for existing deployments.

## How it works

- The verdict computed in SVL-1 now drives the outcome. SOLVED or UNKNOWN (no criteria) pass through as DONE and are never mutated; PARTIAL or NOT_SOLVED trigger a requeue.
- Requeue reuses the executor's existing auto-retry mechanism (the same `should_retry` + `execute_task_background` path the error/timeout retry uses), bounded by a separate `verify_requeue_count` so verify requeues and error retries never share a budget.
- The unmet criteria are appended to a cumulative `metadata["verify_feedback"]` log and rendered in the next attempt's prompt under "## Why the last attempt was rejected", so the agent sees exactly what to fix.
- On budget exhaustion the task escalates to BLOCKED with a needs-review activity and an `mc_task_blocked` broadcast. The DONE side-effects (completion log + deliverable save) are gated on the resolved status, so an escalated task does not log completion or save its rejected output as a deliverable.

## Scope

- `mission_control/executor.py`: act on the verdict (requeue / escalate), reuse the existing re-dispatch, gate the DONE side-effects on the resolved status, render feedback in `_build_task_prompt`.
- Tests: `test_deep_work_e2e.py` covers bounded requeue then BLOCKED, convergence to DONE once the agent passes, SOLVED straight to DONE, flag-off unchanged, and that an escalated task does not mark completed or save a deliverable.

## Verification

`uv run pytest tests/test_deep_work_e2e.py tests/test_mission_control_executor.py tests/test_deep_work_v2.py tests/test_deep_work_scheduler.py` -> 134 passed. Ruff clean.

## Out of scope

The no-progress / oscillation detector (SVL-3), LLM-judge and source-trace verifiers, criticality tags, and per-project budget.

### DIFF
--- a/src/pocketpaw/mission_control/executor.py
+++ b/src/pocketpaw/mission_control/executor.py
@@ -1,6 +1,25 @@
 """Mission Control Task Executor.
 
 Created: 2026-02-05
+Updated: 2026-06-23 — Self-Verifying Loop (SVL-2): the completion hook now
+  ACTS on the SVL-1 verdict. A SOLVED / UNKNOWN result still passes through as
+  DONE (a passing or uncheckable result is never requeued or mutated). A
+  PARTIAL / NOT_SOLVED result is requeued: the SPECIFIC unmet criteria (+ their
+  details) are appended to a cumulative ``verify_feedback`` log, the task is set
+  back to ASSIGNED and re-dispatched via the EXISTING error-retry mechanism
+  (``should_retry`` lever + ``execute_task_background``), bounded by
+  ``deep_work_verify_max_requeues`` via a SEPARATE ``verify_requeue_count``
+  counter (independent of the error ``retry_count``). When the budget is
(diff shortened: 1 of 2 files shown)


prakashUXtech:
Shipped to dev via the stacked rebase in #1540 — SVL-2 is commit 8d3fdd01 on dev. Closing since the change is merged.

PR #1550, opened 2026-06-26
feat(stores): per-workspace store factory + physical Fabric isolation

prakashUXtech:
## What this does

Fabric (the ontology store) currently writes every tenant's data to one shared SQLite file at `~/.pocketpaw/fabric.db`. On a shared cloud box two workspaces sit in the same file, separated only by the `workspace_id` WHERE-clause (the W4a read filter). This change gives each workspace its own file at `~/.pocketpaw/workspaces/<workspace_id>/fabric.db`, the same way KB and Soul already isolate per scope by directory.

`get_fabric_store()` becomes a workspace-keyed factory. It resolves the workspace from an explicit argument, then the `current_workspace` ContextVar, then nothing. With a workspace it returns that workspace's file (cached in a bounded LRU). With no workspace and `POCKETPAW_REQUIRE_WORKSPACE_SCOPE` set (cloud), it raises instead of falling back to the shared file, so a missing workspace on cloud fails closed rather than leaking. Single-tenant OSS keeps the old shared file.

The W4a WHERE-filter stays as a second layer.

## Why

First slice of the cloud paw-os multi-tenant offering (design: `docs/design/drafts/2026-06-26-cloud-paw-os-multi-tenant-offering.md`). Per-file isolation is defense in depth: a missed WHERE clause cannot cross a file boundary.

## Security

`workspace_id` reaches a filesystem path, so it goes through a strict allowlist (`^[A-Za-z0-9][A-Za-z0-9_-]*$`) and fails closed on anything else, with a resolved-path containment recheck behind it. A hostile-id test covers `../../../tmp/pwn`, absolute paths, a null byte, url-encoded `..%2f`, a leading dash, and non-ASCII, asserting each one raises and writes nothing outside the workspaces directory. Independent spec and security reviews both passed.

## Scope

Fabric only. Instinct isolation, the non-router ContextVar bridge, the NULL-row migration, and the EE provider registration are the follow-up slices (ISO-2/3/4) that stack on this branch.

## Tests

- `tests/test_fabric_workspace_isolation.py`: two-workspace separation, fail-closed, and the hostile-id matrix (57 passed)
- fabric regression set: 156 passed; ruff clean; the OSS-cannot-import-EE import linter intact
- new env var `POCKETPAW_REQUIRE_WORKSPACE_SCOPE` (cloud sets it; documented in the module docstring)

Stacked PR: ISO-2/3/4 base on this branch, not `dev`.

### DIFF
--- a/ee/pocketpaw_ee/fabric/router.py
+++ b/ee/pocketpaw_ee/fabric/router.py
@@ -43,11 +43,19 @@
 #   scoped on the type's own ``workspace_id`` now). This supersedes the W4a /
 #   fix/fabric-stats note above that called definitions "global in the schema"
 #   — they carry their own workspace column as of SZD-2.
+# Updated: 2026-06-26 (ISO-1 — physical per-workspace isolation) — the store is
+#   no longer a single shared ``~/.pocketpaw/fabric.db``. ``_store`` now takes
+#   the caller's ``workspace_id`` and routes through
+#   ``pocketpaw.stores.get_fabric_store(workspace_id=...)``, so every read/write
+#   in this router hits that tenant's OWN
+#   ``~/.pocketpaw/workspaces/<id>/fabric.db`` file. The per-endpoint W4a
+#   ``workspace_id`` filter args are UNCHANGED — physical file isolation is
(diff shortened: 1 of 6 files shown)


prakashUXtech:

PR #1552, opened 2026-06-26
feat(catalog): model catalog API over the LiteLLM proxy

prakashUXtech:
## What ships

A read-only model catalog API in the enterprise layer, backed by a self-hosted LiteLLM proxy.

- `GET /api/v1/catalog/models?modality=&provider=&q=&capability=` — filtered list
- `GET /api/v1/catalog/models/{id}` — one model (`id` is the URL-encoded `provider/model` key)

The proxy is the source of truth. The catalog reads `/v1/models` and `/model/info`, maps each model onto a typed `ModelCatalogEntry` (modality, pricing, context, capabilities), and caches the assembled list in-process on a TTL. models.dev fills in logos and descriptions when it's reachable, and the catalog falls back to proxy-only data when it isn't.

This is the foundation slice (MCG-1) for the model-catalog work — the picker in paw-enterprise and the agent-backend gateway path build on these endpoints.

## Notes

- Proxy URL and key come from env (`LITELLM_PROXY_URL`, `LITELLM_PROXY_API_KEY`); nothing is hardcoded.
- The catalog lists served-but-undescribed models too (it unions the two proxy endpoints), so it never silently drops one.
- When the proxy is unreachable the endpoints return 502 rather than an empty 200. That's intended.

## Known gaps (carried forward)

- Non-text pricing (image / audio / video) isn't mapped yet, so those entries show no price for now. The fix lands with the multi-modal slices (MCG-6/7). Documented in the catalog README.
- Every entry is `status=available` until a readiness probe exists.

## Live proxy — handoff

This PR ships the app-side API plus a reference `litellm.config.example.yaml`. Standing up the live proxy (Coolify wiring, real provider keys, Postgres/Redis) is a separate deploy step.

## Tests

`uv run pytest tests/cloud/catalog/` — 39 passing. They mock the proxy (no live dependency) and cover the mapper, modality derivation, filtering, the TTL cache call count, the models.dev fail-open path, and the HTTP layer (response envelope, 400/404/502, URL-encoded ids).

### DIFF
--- a/.env.enterprise.example
+++ b/.env.enterprise.example
@@ -95,10 +95,10 @@ POCKETPAW_LICENSE_KEY=
 # The catalog API (GET /api/v1/catalog/models[/{id}]) reads available models from
 # a self-hosted LiteLLM proxy. Point these at the running proxy; until it is
 # reachable the catalog endpoints return 502 (source of truth unreachable).
-#   LITELLM_PROXY_URL      — proxy base URL. Default http://localhost:4000.
-# LITELLM_PROXY_URL=http://localhost:4000
-#   LITELLM_PROXY_API_KEY  — proxy admin/virtual key (optional; sent as Bearer).
-# LITELLM_PROXY_API_KEY=
+#   POCKETPAW_LITELLM_API_BASE      — proxy base URL. Default http://localhost:4000.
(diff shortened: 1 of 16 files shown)


prakashUXtech:
Folded into the integration PR #1559 — the catalog API is part of it. Ships via #1559.

PR #1553, opened 2026-06-26
fix(agents): route per-agent model to every backend, not just claude/openai/google

prakashUXtech:
## What ships

Per-agent model selection now reaches every backend. `AgentPool._build` routed the stored model onto a Settings field with a substring check that only knew claude/openai/google, so a model set for `codex_cli`, `opencode`, `deep_agents`, `copilot_sdk`, or `langchain_react` was silently dropped. The openai case also wrote `openai_model`, which isn't the field that backend reads first (`openai_agents_model`).

This replaces the substring chain with a single `route_model` helper keyed off the existing backend→field map, so all eight backends route correctly.

## Why now

Foundation for the model picker (MCG-4): a picker is pointless if the chosen model doesn't actually reach the backend. This fixes the silent-drop the catalog work surfaced.

## Scope

- `llm/providers/base.py` — new `route_model(settings, backend, model)` over the existing `_BACKEND_MODEL_ATTR`, plus a `langchain_react → deep_agents_model` alias.
- `agents/pool.py::_build` — calls `route_model`; the substring chain is gone; legacy backend names are canonicalized first.
- `deep_agents` / `langchain_react` (`provider:model`) and `opencode` (`provider/model`) strings are written verbatim, not split.
- OSS-only. No catalog existence-validation here — that belongs to the EE picker (MCG-1/4).

## Tests

`uv run pytest tests/test_pool_route_model.py tests/test_agent_pool_entity_profile_runtime.py tests/test_provider_adapters.py` — 58 passing. The integration tests drive the real `_build` path and prove codex / opencode / deep_agents / langchain_react are no longer dropped. A coverage test fails if a future backend is added without a field mapping.

### DIFF
--- a/src/pocketpaw/agents/pool.py
+++ b/src/pocketpaw/agents/pool.py
@@ -50,6 +50,19 @@ Updated: 2026-06-13 (feat/claude-sdk-prewarm) — added ``prewarm``, which eager
   digest matches too. Fire-and-forget: never raises (the backend's ``prewarm``
   swallows its own errors; this guards instance load + prompt assembly). The EE
   trigger fires it as a background task in ``run_core.execute_run``.
+Updated: 2026-06-26 (feat/mcg-3-pool-route-model) — ``_build`` now routes the
+  per-agent ``config.model`` onto the correct backend Settings field via
+  ``pocketpaw.llm.providers.base.route_model`` instead of the old brittle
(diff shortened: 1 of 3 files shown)


prakashUXtech:
Folded into the integration PR #1559 — the route_model fix is part of it. Ships via #1559.

PR #1554, opened 2026-06-26
feat(stores): bridge per-stream workspace onto the store ContextVar + activate the EE provider

prakashUXtech:
Stacks on #1551 (which stacks on #1550). Merge order: #1550 → #1551 → this, each with `--rebase`.

## What this does

ISO-1 and ISO-2 gave Fabric and Instinct their own per-workspace files, resolved from an explicit workspace argument or the `current_workspace` ContextVar. The EE routers pass that argument. This handles the callers that do not: agent tools, MCP servers, and connector ingest, which call `get_fabric_store()` / `get_instinct_store()` with no argument.

It bridges the per-stream workspace onto the OSS ContextVar inside the existing `attach_agent_identity` / `detach_agent_identity` lifecycle (the token rides along as an extra tuple element, so no call site changes). A bare store call from an agent tool now resolves to the caller's workspace file, and detach resets the ContextVar on the same guaranteed path the identity tokens use, so a thrown run cannot leak a workspace into the next task on that loop.

It also activates the EE side of the `StoreProvider` seam: a `CloudStoreProvider` registered on the `pocketpaw.stores` entry-point, delegating to a public OSS helper so the path and allowlist guard stay in one place. The provider is behavior-neutral today (it returns the same per-workspace file store the OSS default builds); its value is proving the entry-point contract end to end, which is the hook for a future per-tier backend swap.

## Tests

- `tests/test_workspace_ctxvar_bridge.py` (9): an agent-path Fabric and Instinct write resolves to the caller's workspace file; fail-closed on a missing context under required-scope; the ContextVar resets when the run body raises; and the factory routes through a registered provider (first end-to-end proof of the entry-point).
- ISO-1/2/3 isolation + router suites: 241 passed; ruff clean; the OSS-cannot-import-EE import linter intact.

## Pre-existing failure (not from this work)

`tests/cloud/test_mission_control_project_filter.py::test_project_id_empty_string_filters_for_unassigned` is order-dependent: it fails run in isolation (asserts one `list_pockets` call, gets two) and passes inside a broader run. It fails on clean `origin/dev` too, previously as an arity `TypeError` that the ISO-2 store-signature change corrected, which revealed the underlying assertion. Outside the isolation scope; worth a separate ticket (same class as the paw-enterprise `localStorage` test-env red).

### DIFF
--- a/ee/pocketpaw_ee/cloud/chat/agent_service.py
+++ b/ee/pocketpaw_ee/cloud/chat/agent_service.py
@@ -168,6 +168,7 @@ from pocketpaw.ripple import (
     get_pocket_prompts,
 )
 from pocketpaw.ripple._pockets import _MCP_POCKET_BACKENDS
+from pocketpaw.stores import current_workspace as _oss_current_workspace
 from pocketpaw_ee.cloud.shared.errors import CloudError, Forbidden, NotFound
 from pocketpaw_ee.cloud.surface import SurfaceContext, SurfaceProfile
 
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1555, opened 2026-06-26
feat(stores): per-workspace migration + isolation lint guard + audit-lock fix

prakashUXtech:
Stacks on #1554 (the ISO stack: #1550 ← #1551 ← #1554 ← this). Merge in order, each with `--rebase`. Last isolation slice.

## What this does

Three things that finish the per-tenant isolation:

1. A one-time migration (`split_workspace_stores.py`) that splits an existing shared `~/.pocketpaw/{fabric,instinct}.db` into per-workspace files. Fabric rows copy by workspace_id; pre-tenancy NULL rows go to a `system0` bucket. Reversible-safe (the source is renamed `.migrated`, never deleted) and idempotent (a per-file marker; a second run is a no-op).
2. An AST lint guard (`test_store_isolation_lint.py`) that fails the build if any code constructs `FabricStore(...)` / `InstinctStore(...)` directly outside the factory, so the isolation cannot silently regress. It carries a planted-violation self-test; only `stores.py` and the migration's source-read are allowlisted, with the justification in the file.
3. The audit-lock fix from the ISO-2 review: `_log_lock` is now keyed by db_path in a process-global registry instead of held on the store instance, so a workspace store evicted and rebuilt by the LRU shares one lock and concurrent audit appends across the eviction boundary cannot fork the chain.

## The Instinct audit chain survives the migration, with integrity

The shared Instinct file has one global hash chain across interleaved tenants. A naive split would orphan each file's hash links, so the migration re-chains each workspace's rows from genesis over their authentic data, giving every migrated tenant a valid `verify_audit_chain`.

Re-chaining is security-sensitive, so before it runs the migration verifies the source: it calls `verify_audit_chain` on the shared file (read-only, before any rename). If the source is intact it proceeds and records the verdict in the marker. If the source is tampered it aborts with `SourceChainTamperedError`, leaving the source untouched; it will not launder tampered rows into a fresh valid-looking chain. `force=True` overrides with a loud warning and records `forced` in the marker. The marker is a JSON attestation: `{migrated_at, source_chain_verified, forced}`.

## Tests

- 22 in the ISO-4 suite: migration (per-workspace chains intact, NULL→system0, row counts preserved, idempotent, source preserved); the 3 source-gate tests (clean records the verdict; tampered aborts and leaves the source untouched; force overrides and the per-workspace chains are still valid); the lint guard (clean scan + planted-violation self-test); and the lock (shares one lock per file, survives a real cap-1 LRU eviction with the chain intact under concurrent cross-eviction appends).
- ISO-1/2/3/4 regression slice: 482 passed; ruff clean; the OSS-cannot-import-EE import linter intact.

Independent security review covered the ISO-1/2 runtime primitives; this slice was verified against committed HEAD.

### DIFF
--- a/src/pocketpaw/_store_locks.py
+++ b/src/pocketpaw/_store_locks.py
@@ -1,0 +1,61 @@
+# src/pocketpaw/_store_locks.py
+# Created: 2026-06-26 (ISO-4 — audit-lock carry from the ISO-2 security review).
+#
+# Process-global registry of per-db-file asyncio locks, keyed by db_path. The
+# Instinct audit hash-chain is serialized by a lock held across the chain
+# read-head + insert in InstinctStore._log: without it, two concurrent _log
+# calls could both read the same prev_hash before either inserts, forking the
+# chain. Until ISO-4 that lock lived on the STORE INSTANCE (self._log_lock).
+#
+# The ISO-2 security review flagged the gap that motivates this module: under
+# the workspace-keyed factory's bounded LRU (ISO-1/2), a workspace's
+# InstinctStore can be EVICTED and a NEW instance built for the same file while
+# an async _log on the old instance is still in flight. Two instances for the
+# same db_path each held their OWN per-instance lock, so they did NOT mutually
+# exclude — re-opening the within-tenant chain-fork window the lock exists to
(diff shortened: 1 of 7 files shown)


prakashUXtech:

PR #1557, opened 2026-06-26
feat(agents): verify Claude-Code-through-LiteLLM routing + gateway probe findings

prakashUXtech:
## What ships

Confirms — and locks in with tests — that selecting the `litellm` provider routes the Claude Code harness through the LiteLLM proxy, plus a findings note from a live gateway probe.

- The routing wiring already existed: the `litellm` provider sets `ANTHROPIC_BASE_URL` from `POCKETPAW_LITELLM_API_BASE`, the spawned `claude` subprocess inherits it, and smart-routing is skipped for non-Anthropic providers so it doesn't clobber the chosen model. This PR adds the unit tests that pin that contract (`tests/test_llm_client.py`).
- A handoff note (`docs/handoff/2026-06-26-mcg2-gateway-probe.md`) capturing the live probe results.

## Gate finding — does prompt-cache survive the LiteLLM hop?

The proxy **preserves cache metadata** — proven against DeepSeek auto-cache (`prompt_cache_hit_tokens: 2944` of 3056 on a repeated prefix). It also relays tool-calls, streaming, and per-call spend faithfully (`/spend/logs?api_key=` shows the rows + cost breakdown).

The Anthropic-specific cache number is **blocked only by the proxy's Anthropic upstream key being invalid/expired** — every Claude model returns `authentication_error: invalid x-api-key`. That's a proxy-credential fix, not a code or design blocker: the gateway demonstrably does not strip cache fields, so once a valid Anthropic key is on the proxy, Anthropic prompt-cache should survive the same path.

## Captain action items (proxy config, not code)

- Fix the proxy's Anthropic upstream key (blocks the Anthropic cache number + Claude-through-proxy in prod).
- OpenAI key is quota-exhausted (429) — `gpt-4o` unusable until topped up.
- `/spend/logs` must be queried with `?api_key=` (unfiltered times out).

## Tests

`uv run pytest tests/test_llm_client.py` — 25 passing, including 2 new tests asserting the `litellm` provider routes `ANTHROPIC_BASE_URL` + the key to the configured proxy.

### DIFF
--- a/docs/handoff/2026-06-26-mcg2-gateway-probe.md
+++ b/docs/handoff/2026-06-26-mcg2-gateway-probe.md
@@ -1,0 +1,143 @@
+<!--
+MCG-2 GATE findings note (NEW FILE, 2026-06-26).
+Records the live-proxy probe results for routing Claude Code through the
+LiteLLM proxy: prompt-cache survival, tool/stream fidelity, spend logging,
+and the captain-config gaps found. Durable deliverable for slice MCG-2.
+-->
+
+# MCG-2 — Claude Code through LiteLLM: gateway probe findings
+
+**Date:** 2026-06-26
+**Proxy:** `https://litellm.hzd.interacly.com` (key alias `paw-enterprise`, `sk-...2qyw`)
+**Slice:** MCG-2 (the GATE) — does routing the spawned `claude` CLI through the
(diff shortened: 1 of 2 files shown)


prakashUXtech:
Folded into the integration PR #1559 — the gateway-probe wiring is part of it. Ships via #1559.

PR #1564, opened 2026-06-26
test: add unit tests for http_utils.is_request_secure

### LINKED ISSUE #1562: Add test coverage for http_utils.is_request_secure
### Problem Statement

src/pocketpaw/http_utils.py contains is_request_secure(request), which makes a
security-relevant decision — whether the original client protocol was HTTPS — and
is used to gate HTTPS-dependent behavior (e.g. secure-cookie decisions) across the
API and dashboard modules.

The function has zero test coverage, yet its logic has several easy-to-break
subtleties: it must read the *first* hop of a comma-separated X-Forwarded-Proto
chain, parse the RFC 7239 Forwarded header (including quoted proto values), be
case- and whitespace-insensitive, and fall back safely on malformed headers. A
silent regression here could downgrade a request that is actually secure (or, worse,
treat an insecure request as secure), with no test to catch it.

### Proposed Solution

Add a new test module tests/test_http_utils.py that exercises every branch of
is_request_secure. No production code changes.

A small helper builds a Starlette/FastAPI Request from an ASGI scope so the URL
scheme and request headers can be controlled directly. Coverage:

- Direct scheme: https -> secure; http with no forwarding -> not secure; direct
  https short-circuits conflicting forwarded headers.
- X-Forwarded-Proto: single value (https/http), first-hop selection in a
  comma-separated chain ("https,http" vs "http,https"), surrounding whitespace,
  case-insensitivity, empty value, and fall-through to the Forwarded header when
  the value is http.
- RFC 7239 Forwarded: proto=https / proto=http, proto alongside other params
  (for=...;proto=https), quoted values (proto="https"), first-entry selection in a
  multi-entry chain, case-insensitive keys/values, whitespace around params,
  missing proto param, empty value, and a malformed value with no assignment.

- tests/test_http_utils.py (new file — unit tests only)

### Alternatives Considered

- Testing the behavior indirectly through the dashboard/API endpoints that call
  is_request_secure. Rejected: heavier, slower, and couples the test to unrelated
  routing/auth setup instead of pinning the header-parsing logic directly.
- Constructing Request objects via a TestClient. Rejected: building the ASGI scope
  directly is simpler and lets each case set an exact scheme + header combination
  with no server round-trip.

### Additional Context

I'd like to pick this up.

### Before submitting

etisamhaq:
## What does this PR do?

Adds a unit test module for `http_utils.is_request_secure`, which was previously
untested. The function makes a security-relevant decision (was the original client
protocol HTTPS?) by parsing the request scheme plus the `X-Forwarded-Proto` and
RFC 7239 `Forwarded` headers. These tests pin down its branches so a future change
can't silently downgrade a secure request. No production code is changed.

## Related Issue

Fixes #1562 

## Changes Made

- `tests/test_http_utils.py` (new): 22 tests covering `is_request_secure`. A small
  helper builds a Starlette/FastAPI `Request` from an ASGI scope so the URL scheme
  and headers can be set precisely. Cases cover:
  - Direct scheme — `https` secure, `http` not, and direct `https` short-circuiting
    conflicting forwarded headers.
  - `X-Forwarded-Proto` — single value, first-hop selection in a comma chain,
    whitespace, case-insensitivity, empty value, and fall-through to `Forwarded`
    when the value is `http`.
  - RFC 7239 `Forwarded` — `proto=https`/`http`, proto alongside other params,
    quoted values, first-entry selection, case-insensitive keys/values, whitespace,
    missing `proto`, empty, and malformed input.

## How to Test

1. `uv sync --dev`
2. `uv run pytest tests/test_http_utils.py -v`
3. `uv run ruff check tests/test_http_utils.py`

## Evidence of Testing

```
$ pytest tests/test_http_utils.py -q
........................                                                  [100%]
22 passed in 0.32s

$ ruff check tests/test_http_utils.py
All checks passed!

$ ruff format --check tests/test_http_utils.py
1 file already formatted
```

## Checklist


etisamhaq:
Hi! Just following up on this PR. Whenever you have time, could you please review it? If any changes are needed, I'm happy to address them. If everything looks good, I'd appreciate a merge. Thanks!

PR #1565, opened 2026-06-26
feat(cloud): per-workspace versioned file writes (file_versions)

prakashUXtech:
## What this adds

A `file_versions` cloud entity for versioned, per-workspace file writes. This is the storage foundation the cloud agent-artifact work builds on.

- `POST /files/write` — create or overwrite a file at a path; returns a file id and version
- `PUT /files/{id}` — inline save; archives the prior blob as a version and bumps the counter
- `GET /files/{id}/versions` and `/versions/{vid}` — version list and single-version fetch

Live content goes to the shared upload `StorageAdapter` (local or S3, env-selected) under a workspace-prefixed key. Version history lives in Mongo. Every read is workspace-filtered.

## Where it came from

This ports the storage core of dewani12's `feature/files` branch onto current dev. That branch had drifted 319 commits behind and predated the cloud restructure, so this is a path and import migration (`ee/cloud/*` → `ee/pocketpaw_ee/cloud/*`) conformed to the current 4-file shape, the tenant-filter rule, and the import-linter contracts. The editor tools and Editor.js UI from that branch are deliberately left out and tracked as a follow-up.

## Multi-tenant correctness

The stored `FileUpload` id is namespaced by workspace (`{workspace_id}:{path}`), so two tenants can each own a file at the same path without colliding on the collection's global unique index. Soft-deleting then recreating a path revives the row and starts a fresh version history.

Independent review caught the collision risk before this opened, plus a few version-handling edges (no-op writes returning a phantom version, hardcoded mime, silent content loss on an archive read failure, off-by-one version labels). Those fixes are in the second and third commits.

## Scope

Storage spine only. No agent-facing tool, no working-directory change, no editor surface. Those are the next tasks in this stack.

## Tests

`tests/cloud/file_versions/` — 9 tests: write plus version bump, two-workspace isolation (no cross-read on the version list or content), same-path-no-collision, soft-delete then recreate starts fresh history, no-op version handling, mime inference, and archive-read-failure abort. The tests assert the logic directly, because the mock Mongo harness does not enforce unique indexes (which is what hid the collision in the first pass).

import-linter: 26 contracts kept. ruff clean.

## Stack

This is the base of a stack. ART-2 (per-tenant agent working directory), ART-3 (jail lifecycle), and ART-4 (deliver tool) follow. Merge this base with `--rebase`, not squash.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -49,6 +49,12 @@ instead of firing an HTTP write, and the new
 GET /workspaces/{ws}/jobs/{job_id} status poll. Jobs run on the shared ARQ
 worker under the synthetic `system:workspace_job` identity and merge their
 results back into the pocket's `state` over the live update bridge.
+
+Updated: 2026-06-26 (ART-1) — documented the Files — Versioned Writes
+section: POST /files/write, PUT /files/{id}, and the two version-history
+reads (GET /files/{id}/versions[/{vid}]). The write path archives each
+prior blob as a FileVersionDoc and bumps a per-file content_version counter;
+every read is workspace-scoped.
 -->
 
 # Cloud REST API Reference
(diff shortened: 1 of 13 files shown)


prakashUXtech:

PR #1566, opened 2026-06-26
feat(cloud): per-tenant agent working-directory jail with fail-closed gating

prakashUXtech:
## What this adds

Per-tenant working-directory isolation for the cloud chat agent. The agent (a Claude Code subprocess) used to run with its cwd set to `settings.file_jail_path`, which defaults to the shared home dir. In multi-tenant cloud that means every tenant's agent writes into the same filesystem. This change jails each workspace's agent into `~/.pocketpaw/workspaces/<workspace_id>/agent/<session_id>/` and fails closed if it can't determine the workspace.

- Cloud run: cwd resolved per-run from the `attach_agent_identity` workspace ContextVar; directory created on demand.
- A cloud run that reaches the agent with no resolvable workspace **raises** instead of falling back to the home dir. Mis-tenanting fails loudly rather than co-mingling.
- OSS and single-tenant dedicated deployments are unchanged: no cloud DB, no resolver, cwd stays `file_jail_path`.
- The cloud-mode signal is `get_client()` truthiness, the existing convention (same gate as `cli.py`).

## Security

Reviewed for spec, quality, and as a dedicated tenant-isolation audit. The boundary holds:

- **Fail-closed is complete** on both the run and prewarm paths, and is locked by a run()-level test that proves the agent never builds SDK options (never spawns) when the workspace is unresolvable. This guards against a future generic cwd fallback silently slipping past green CI.
- **Path traversal:** workspace and session ids are validated against `^[A-Za-z0-9_.-]+\Z` with explicit `.`/`..` rejection before any mkdir. The ids are server-set authenticated values (24-hex ObjectIds), never request input.
- **Warm-client isolation is structural:** the resolved cwd is part of the agent subprocess cache key, so a warm client can never be reused across two tenants' directories.
- **Audit:** a high-severity audit event (`agent.cwd_jail.fail_closed`) fires when the fail-closed path triggers, as fire-and-forget that cannot block or swallow the raise.

## Scope

Working-directory isolation only. Quota / TTL GC / disk-watermark eviction is the next task (ART-3). The deliver-to-blob tool is ART-4.

The one scope-adjacent change is binding workspace identity in `_prewarm_session`: prewarm runs before the stream binds identity, so without it every cloud session would warm a wrong-cwd client and lose its turn-1 warm. It mirrors the run-path binding exactly.

## Known residual (tracked follow-up, not in this PR)

The jail lives in `ClaudeSDKBackend`. A cloud agent configured with a non-claude backend (`deep_agents`, `openai`, `codex`) would bypass the jail and run un-isolated. Reachability is low today: the cloud default is `claude_agent_sdk` and non-claude backends need keys the managed cloud doesn't provision. The fix is to either pin `claude_agent_sdk` for cloud chat agents or lift the jail into a shared backend base, tracked as a follow-up.

Minor: the fail-closed audit task is fire-and-forget and could be dropped under GC pressure (alert delivery only, never the raise) — a one-line follow-up to hold a task reference.

## Tests

`tests/cloud/agents/test_agent_jail.py` + `tests/test_agents.py` + `tests/test_fast_path.py`: two-workspace cwd isolation, per-session and sessionless `_shared` paths, fail-closed when cloud + no workspace, OSS default unchanged, hostile-id rejection (including trailing newline), run()-level fail-closed (no home fallback), and warm-client reconnect on cwd change. 200 in the focused sweep, 367 in the broad sweep. ruff clean.

One pre-existing unrelated red (`test_agent_bridge_emits.py`, a committed Windows path) is untouched by this change.

## Stack

Stacked on ART-1 (`feat/art1-file-versions-port`, PR #1565). Merge the base with `--rebase`, not squash.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -221,6 +221,17 @@ The web dashboard (`frontend/`) is vanilla JS/CSS/HTML served via FastAPI+Jinja2
   env-configurable), marking queued/running `ChatRunDoc`s older than 10 minutes as
   `interrupted` so runs abandoned by a backend restart surface a retry affordance
   instead of leaving clients subscribed forever.
+- **Per-tenant agent cwd jail (cloud, ART-2)**: In multi-tenant cloud each
+  workspace's chat agent runs with `cwd = ~/.pocketpaw/workspaces/<workspace_id>/agent/<session_id>/`
+  (resolved per-run from the `attach_agent_identity` ContextVars in
+  `ee/pocketpaw_ee/cloud/agent_jail.py`) so one tenant's file ops never
+  co-mingle in the shared home dir. It **fails closed**: a cloud run that
+  reaches the backend with no resolvable `workspace_id` RAISES rather than
+  falling back to `~`. OSS / dedicated installs (no cloud DB initialized) keep
+  `settings.file_jail_path` unchanged. Override the jail root with
+  `POCKETPAW_WORKSPACE_JAIL_ROOT` (default `~/.pocketpaw/workspaces`) to anchor
+  it on a data volume. Lifecycle (quota / TTL-GC / disk-watermark eviction of
+  these dirs) is ART-3, not yet shipped.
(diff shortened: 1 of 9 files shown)


prakashUXtech:

PR #1568, opened 2026-06-26
feat(cloud): bound the per-tenant agent jail — quota, TTL GC, watermark eviction

prakashUXtech:
## What this adds

A lifecycle for the per-tenant agent scratch jail introduced in ART-2, so the jail can't fill the box. This is the answer to "what happens at 1000 users": disk scales with *active concurrency*, not total user count, because the jail is pure scratch (durability lives in blob storage) and gets reclaimed aggressively.

Three mechanisms, all cloud-only:

- **Per-workspace quota** checked at run-start. Over quota → the run is rejected cleanly (a terminal `failed` run plus an `agent.jail_over_quota` error frame), the worker survives, no crash. The check early-exits as soon as the jail crosses the limit, so it doesn't stat a huge tree on every run's critical path.
- **TTL garbage collection.** A background sweeper (registered at cloud boot and on the existing 5-minute heartbeat, mirroring the resumable-run sweeper, each isolated so a GC failure can't suppress the other sweeps) deletes idle jails after a grace period.
- **Disk-watermark eviction.** When box disk crosses a high-water mark, the least-recently-used idle jails are evicted first until back under the mark.

## The safety invariant

A jail backing a live run is **never** evicted. This holds via two complementary signals, and a jail is reclaimed only when both agree it's safe:

- the DB active-set (`find_active_run_scopes` — every queued or running run), which protects a freshly-queued run whose jail is still empty;
- the newest mtime across the whole jail subtree, which protects a long in-progress write (a deep `node_modules` write that doesn't bump the top-dir mtime) even if the run were wrongly excluded from the DB set.

Eviction is path-guarded (refuses any path not strictly under the jail root) and idempotent, so concurrent web replicas only double the IO, never corrupt. Reviewed as spec + quality with the eviction-safety property traced explicitly, including the queued-after-snapshot race (benign).

## New env vars (cloud-only)

`POCKETPAW_AGENT_JAIL_QUOTA_MB` (2048), `POCKETPAW_AGENT_JAIL_TTL_GRACE_SECONDS` (3600), `POCKETPAW_AGENT_JAIL_DISK_WATERMARK_PCT` (90), `POCKETPAW_AGENT_JAIL_GC_ENABLED` (true).

## Known limits (by design)

- The sweep is per-process, not distributed-locked — safe because eviction is idempotent and the topology is single-box per tenant.
- If every jail on the box is backed by an active run and disk is over the watermark, the GC correctly evicts nothing (it won't kill a live run to free disk); the per-workspace quota is the backstop. Box sizing should budget roughly `quota × max-concurrent-workspaces`.

## Scope

Lifecycle only. ART-2's cwd resolution and fail-closed logic are untouched. The deliver-to-blob tool is the next task (ART-4).

## Tests

`tests/cloud/agents/test_agent_jail_lifecycle.py` + `tests/cloud/runs/test_run_core_jail_quota.py`: over-quota clean reject (run survives), the quota walk short-circuits, idle jail swept after grace, watermark evicts LRU-idle first, and an active run's jail (session and the sessionless `_shared` dir) is protected while a terminal run's is not. 38 in the agents sweep, 119 in the runs sweep, ruff clean.

## Stack

Stacked on ART-2 (`feat/art2-cwd-jail`, PR #1566). Merge the base with `--rebase`, not squash.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -247,8 +247,9 @@ The web dashboard (`frontend/`) is vanilla JS/CSS/HTML served via FastAPI+Jinja2
   `POCKETPAW_AGENT_JAIL_GC_ENABLED` (default `true`) — escape hatch to disable
   the GC sweep entirely. The GC (`ee/pocketpaw_ee/cloud/agent_jail_gc.py`,
   `sweep_agent_jails`) runs on cloud startup and the same 5-minute heartbeat as
-  the stale-run sweeper, and NEVER evicts a jail whose run is still
-  queued/running/resumable (resolved via `run_service.find_active_run_scopes`).
+  the stale-run sweeper, and NEVER evicts a jail whose run is still queued or
+  running (resolved via `run_service.find_active_run_scopes`; a retried
+  interrupted run re-protects its jail by spawning a fresh queued run).
 - **In-process bus subscribers**: `pocketpaw_ee.cloud._core.realtime.bus.InProcessBus` exposes `subscribe(event_type, handler)` for cloud-side listeners (e.g. the `FileReady` → KB indexer wired in `ee/pocketpaw_ee/cloud/uploads/listeners.py`). Register subscribers from `mount_cloud()` after `init_realtime()` runs. Handler exceptions are logged and swallowed per-handler so one bad listener can't block the rest of the dispatch.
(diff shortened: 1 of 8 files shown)


prakashUXtech:

PR #1569, opened 2026-06-27
feat(cloud): deliver_artifact tool — agent artifacts to per-tenant blob with a real download URL

prakashUXtech:
## What this adds

The payoff of the stack: a cloud agent's built artifact now lands in the tenant's blob storage with a real download URL, instead of the shared container filesystem.

This closes the original bug — a cloud user asks the agent for "a downloadable landing page," the agent writes files to `/home/pocketpaw/...` and serves a preview from `127.0.0.1`, neither of which the user can reach and both of which co-mingle across tenants. Now the agent calls `deliver_artifact` and the user gets a real, per-tenant, presigned download link that appears in their Files.

- **`deliver_artifact` tool** (cloud-only): takes a path in the agent's jail. A single file uploads as-is (mime guessed); a directory is zipped and uploaded as `application/zip`. Routes through the existing `EEUploadService.upload`, so the artifact is workspace-scoped, emits `FileReady` (→ KB), shows in `GET /files`, and returns a presigned download URL.
- **Cloud storage boot guard** `verify_cloud_storage_backend()`: warns when a cloud deploy isn't on the S3 adapter (so the feature can't silently write "blob" artifacts to local disk), and hard-fails under `POCKETPAW_REQUIRE_S3_IN_CLOUD`. Mirrors the memory-backend boot guard.
- **`is_multi_tenant_cloud()`**: one named home for the `get_client()` cloud signal, now used by the boot guard and the ART-2 jail.
- **System-prompt guidance** (cloud): tells the agent to deliver downloadable results via the tool, never by printing a path or starting a local preview server.

## Security

Reviewed for spec, quality, and as a dedicated audit. The boundary holds:

- **Path jail:** `_resolve_in_jail` resolves symlinks before a component-wise containment check against the workspace jail root. `..`, absolute paths, `/etc/passwd`, symlinks pointing out, and another tenant's jail are all rejected; directory zipping skips symlinks so a link can't smuggle its out-of-jail target into the archive. Five escape vectors are tested.
- **Cross-tenant:** uploads are workspace-scoped and the presign is double-gated (workspace scope + an owner/member/admin read check), so workspace A cannot obtain or list workspace B's artifacts. Proven by a two-tenant test.
- **Surface scoping:** the tool is ambient on the default chat surface but filtered out of every restricted surface (studio/sites/belt/foresight), with a test locking the exclusion against a future allowlist edit.
- **No inline-render:** the audit caught that the S3 presigned path served agent-delivered HTML/SVG/JS inline (the app download route already forces `attachment` for non-`INLINE_MIMES`, but the presign didn't). Fixed: `presigned_get` now forwards `Content-Disposition: attachment` for non-`INLINE_MIMES`, mirroring the app route, with a sanitized filename. So a delivered `.html`/`.svg` downloads, it doesn't render on the storage origin.
- **Fail-closed:** the tool errors without a resolvable workspace/user — it never delivers to an unknown tenant. The build is gated on `is_multi_tenant_cloud()`.

## Scope

ART-1/2/3 logic is untouched except a 10-line swap in `agent_jail.py` to use the shared `is_multi_tenant_cloud()` helper. No file_versions/editor/Slice-D.

## Tests

`tests/cloud/agents/test_deliver_tool.py` + `tests/cloud/uploads/test_storage_bootstrap.py` + `test_presigned_disposition.py`: single-file deliver (workspace-scoped, visible in Files, correct mime), directory → zip round-trip, two-tenant zero cross-read, five path-escape vectors, a symlink inside a delivered directory is skipped, the attachment disposition for html/svg (and inline preserved for images/pdf), missing identity, oversize cap, the boot-guard warn/raise matrix, and the cloud-only build gate. 210 in the focused sweep, ruff clean.

A pre-existing unrelated set of 401 auth-harness failures (`uploads` router/folders tests) is untouched by this change.

## Stack

Stacked on ART-3 (`feat/art3-jail-lifecycle`, PR #1568). Merge the bases with `--rebase`, not squash. Full stack: ART-1 #1565 → ART-2 #1566 → ART-3 #1568 → this.


prakashUXtech:

PR #1580, opened 2026-06-28
feat(instinct): finish deny-by-default — read_only marker + template deny-on-no-match

prakashUXtech:
## What

Binding-level Instinct routing already defaults to gated (`requires_instinct=True`). Two gaps remained: there was no way to mark a genuinely read-only action as gate-exempt, and a template-bound pocket fell through to EXECUTE when no rule matched an action. This closes both.

Two commits:
- **AW-6** — `read_only: bool` on `ActionBinding`, gate-7 suppression for read_only actions, and a bidirectional validator so a GET/HEAD binding is admissible only as an explicit `read_only` action. The action surface stays write-only otherwise, so a compromised client cannot smuggle a read verb in.
- **AW-7** — when a bound template has no matching rule for a mutating action, the gate now returns PENDING instead of EXECUTE, behind a per-workspace `instinct_template_default_deny` flag (global default plus a workspace-document override, mirroring how `approval_level` is plumbed). Reads still proceed. BLOCK, a matched EXECUTE rule, and notify-only are never overridden.

## Rollout

`instinct_template_default_deny` defaults **off**, so existing template-bound pockets are unchanged until a workspace opts in. The temporal sweeper is intentionally left unchanged (it inherits the off default; its rows are already planner-cleared).

## Tests

46 targeted tests pass; the broader instinct + executor sweep passes 246; `ruff` clean.

## Known nuance (carried forward, out of scope)

A `read_only` GET that skips the park still flows through the write-allowlist and the gate-8 write executor. AW-7 guarantees the template flip does not gate reads. Routing `read_only` actions to `source_executor` instead is a deeper refactor, noted in code and docs.

## Design

`docs/design/drafts/2026-06-24-aiam-whitespace-hardening{,-prd,-tasks}.md` — Slice 2 (AW-6, AW-7). This finishes deny-by-default end to end; the binding default shipped earlier in W2a.

### DIFF
--- a/docs/concepts/pocket-write-actions.mdx
+++ b/docs/concepts/pocket-write-actions.mdx
@@ -75,7 +75,9 @@ A write binding lives inside `rippleSpec.actions` on the pocket. The minimum sha
 |-------|---------|
 | `method` | One of `POST` / `PUT` / `PATCH` / `DELETE`. Validated server-side; the client cannot override. |
 | `path` | Resolved path appended to the backend `base_url`. `:placeholder` segments are filled at call time. |
-| `requires_instinct` | When `true`, runs validation gates but parks the write for human approval via the Instinct surface. |
+| `requires_instinct` | Defaults to **`true`** (deny-by-default): runs validation gates but parks the write for human approval via the Instinct surface. Set `false` only with `instinct_exempt: true` to fire ungated. |
+| `instinct_exempt` | When `true` (paired with `requires_instinct: false`), the explicit, auditable opt-out of the gate. Rejected by the model validator alongside an `instinct_policy`. |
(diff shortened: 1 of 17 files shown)


prakashUXtech:
Adversarial security review (per-branch, refute-by-default) flagged one HIGH on this branch and confirmed it: the template deny-by-default flag was threaded through the router path but not the temporal sweep, so a scheduled mutating action with no matching rule would execute ungated while a human-initiated one parked. Fixed in 493861fc — the temporal dispatcher now resolves the workspace flag and threads it (plus a fail-closed `is_mutating`) into both gate points. New tests prove an uncovered mutating temporal action parks, flag-off is unchanged, and read-only actions are never gated. 63 targeted tests pass. The other review findings on this branch were info/low (read_only routes through the write executor by design).

PR #1590, opened 2026-06-30
feat(agents): native session resume wiring for the Claude SDK backend

prakashUXtech:
Adds the plumbing for an agent to continue a conversation by resuming its native CLI session instead of replaying chat history into the prompt each turn.

- `SessionHandle` (cli_session_id + an opaque session_store), threaded through `AgentPool.run` only when set, matching the existing deny/allow/skills idiom.
- In the Claude SDK backend, a handle carrying a `cli_session_id` sets `ClaudeAgentOptions.resume` and routes that turn down the fresh `query()` launch path. The warm persistent client can't take a fresh resume (its cache key omits it), so a resume turn deliberately bypasses it.
- History injection is skipped on a resume turn (the native session already carries it); the per-turn system prompt is still rebuilt and applied, so resumed turns honor fresh instructions and KB.
- The native session id is captured from the turn-1 init message and surfaced as a `session_id` event for the executor to persist.

Without a handle, behavior is unchanged.

Tests: resume sets `resume` and uses the fresh path, per-turn system prompt is honored on resume, turn-1 id is captured, legacy path unchanged. 6 new plus regression, green.

First of 6 stacked PRs (base: dev). Live two-turn continuity is confirmed by hand separately, since nested `claude` calls can't run in the build sandbox.

Out of scope here: the durable store, the supervisor, and the executor wiring (later PRs in the stack).

### DIFF
--- a/src/pocketpaw/agents/backend.py
+++ b/src/pocketpaw/agents/backend.py
@@ -22,6 +22,15 @@ per-entity skill subset the Claude SDK backend materializes into a per-run local
 plugin). The ``system_message_override`` field is applied UPSTREAM in
 ``AgentPool.run`` (it swaps the base system prompt before assembly), so it never
 reaches a backend as a kwarg — it rides the existing ``system_prompt`` channel.
+
+Updated: 2026-06-30 (feat/session-supervisor SS-1) — adds the ``SessionHandle``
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1591, opened 2026-06-30
feat(agents): tenancy-keyed SessionStore for durable native resume

prakashUXtech:
Backs native resume with our own store, so a session survives a backend restart and two tenants on one box can't read each other's transcripts.

- `InMemorySessionStore` (OSS reference) and `MongoSessionStore` (ee, durable) both implement the SDK's `SessionStore` protocol. Every key is `(workspace_id, project_key, session_id, subpath)` and the store is bound to a workspace at construction, so a store for tenant B never sees tenant A's rows.
- The Claude SDK backend passes the store through to `ClaudeAgentOptions.session_store` when present. On a resume turn the SDK materializes the transcript from the store into a throwaway temp dir, then mirrors new turns back via `append`.
- The store reaches the OSS backend opaquely, so `src/pocketpaw` never imports `pocketpaw_ee`.

Known limitation: the Mongo `append` is read-modify-write, not an atomic `$push`. Fine at current volume, flagged for concurrency hardening.

Tests: store round-trip, durability across a simulated restart, tenant isolation, and the backend passing the store into options. 22 green (OSS plus mongomock).

Stacked on the SS-1 branch.

### DIFF
--- a/ee/pocketpaw_ee/cloud/agent_sessions/__init__.py
+++ b/ee/pocketpaw_ee/cloud/agent_sessions/__init__.py
@@ -1,0 +1,14 @@
+# ee/pocketpaw_ee/cloud/agent_sessions/__init__.py — the agent-session entity.
+#
+# Holds the Mongo-backed ``SessionStore`` (feat/session-supervisor SS-2) that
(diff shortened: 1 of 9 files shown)


prakashUXtech:

PR #1592, opened 2026-06-30
feat(cloud): durable map from chat session to native CLI session id

prakashUXtech:
Records which native CLI session belongs to each `(workspace, session, agent)` so any later turn, including a cold one after a restart, can resume the right session instead of starting over.

- New `AgentSessionRuntimeDoc` (collection `agent_session_runtimes`) with a unique index leading on `workspace`.
- `runtime_service.set_cli_session_id` (upsert) and `get_cli_session_id` (tenant-filtered lookup). The service is the only importer of the doc; reads always filter by workspace.
- Internal index with no HTTP surface, so no DTO/router and no realtime event.

Tests: persist then look up, upsert/backfill keeps a single row, foreign-tenant and absent keys return None. 6 green.

Stacked on the SS-2 branch.


prakashUXtech:

PR #1593, opened 2026-06-30
feat(agents): SessionSupervisor with warm tiers, per-tenant quota, turn-1 capture routing

prakashUXtech:
The lifecycle engine that owns one runtime per `(workspace, session)` and decides, per turn, whether to reuse a warm process or take a fresh launch.

- COLD and WARM tiers (LIVE reserved). A second turn within the TTL can reuse a warm slot; idle slots are reaped; a busy runtime (active_runs > 0) is never reaped or evicted, the same guard the agent pool uses.
- Per-tenant warm quota plus a global cap, evicting the oldest idle warm session in scope. One tenant hitting its cap never evicts another tenant's sessions.
- Turn 1 of a session, before any native id is known, always routes to its own fresh launch, so the id the SDK captures binds to that session rather than a shared warm client. This closes the turn-1 capture gap noted in the SS-1 PR.
- Crash, reap, and evict all demote to COLD while keeping the cli_session_id, so the next turn resumes from the store.

The supervisor doesn't spawn processes. The executor supplies the warm slot and a teardown callback, which keeps this a pure policy engine, tested with a fake clock and fake teardowns.

Known limitation: COLD runtimes stay resident in memory; a COLD-TTL prune is the follow-up for long SaaS uptime.

Tests: warm reuse, idle reap, busy-guard, per-tenant quota including a busy-skip case, turn-1 routing with foreign-slot and expired-slot edges, crash to COLD. 12 green.

Stacked on the SS-3 branch.

### DIFF
--- a/src/pocketpaw/agents/session_supervisor.py
+++ b/src/pocketpaw/agents/session_supervisor.py
@@ -1,0 +1,437 @@
+"""Session Supervisor — agent-session lifecycle across COLD/WARM tiers (SS-4).
+
+Created 2026-06-30 (feat/session-supervisor SS-4): the policy / lifecycle engine
+that owns one ``SessionRuntime`` per ``(workspace_id, session_id)`` and decides,
+for each turn, whether to reuse a live warm slot (WARM) or take a fresh launch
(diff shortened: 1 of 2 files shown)


prakashUXtech:

PR #1594, opened 2026-06-30
feat(cloud): drive supervised native-resume sessions through the executor (flag-gated)

prakashUXtech:
Wires the supervisor, the durable id map, and the per-tenant store into the live chat executor, behind `POCKETPAW_SESSION_SUPERVISOR` (default off).

When on, each agent turn:
- recovers any prior native id for `(workspace, scope_id, agent)`,
- acquires the runtime (turn 1 owns capture, later turns carry the resume id),
- threads a `SessionHandle` pairing that id with this tenant's `MongoSessionStore`, so the agent resumes natively instead of replaying history,
- brackets the run with the busy-counter and persists the turn-1 id,
- demotes to COLD on failure so the next turn still resumes.

When off, `pool.run` is called with no handle and nothing else fires, so the path is unchanged. Any supervisor or store error during a turn falls back to the legacy path for that turn rather than breaking the run.

v1 doesn't bind a live warm slot; every supervised turn resumes from the store. Warm hot-process reuse is a follow-up.

Session identity uses `ctx.scope_id` (stable per conversation, present on the worker path) rather than the Mongo session-doc id, which is None there.

Tests: turn-1 capture, turn-2 resume, flag-off legacy path unchanged, backend-error demotes to COLD. The full chat-runs suite stays green (128).

Stacked on the SS-4 branch.

### DIFF
--- a/ee/pocketpaw_ee/cloud/chat/runs/run_core.py
+++ b/ee/pocketpaw_ee/cloud/chat/runs/run_core.py
@@ -1,6 +1,28 @@
 """Agent-run core — the loop the executor invokes for every chat run.
 
 Changes:
+- 2026-06-30 (feat/session-supervisor SS-5) — ``_drive_agent_loop`` now drives
+  every supervised agent turn through the ``SessionSupervisor`` + the durable
+  ``(workspace, session, agent) -> cli_session_id`` mapping (SS-3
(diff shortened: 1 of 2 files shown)


prakashUXtech:

PR #1595, opened 2026-06-30
test(cloud): adversarial multi-tenant session-isolation gate

prakashUXtech:
A consolidated isolation suite for the session subsystem, meant as a merge gate for the multi-tenant SaaS track. Tests only, no production change.

Covers, adversarially:
- two stores sharing one backing dict or one collection, colliding on the same key. Neither tenant can load, list, or delete the other's transcript, including a guessed-id attempt.
- the id map: a foreign-tenant lookup returns None, and a same-key write from another tenant creates a separate row.
- supervisor quota pressure from one tenant never evicts another's warm session, and a busy runtime is never taken.
- a leaked native id paired with a store bound to another tenant resolves to nothing, checked at the store level and through the SDK's real `materialize_resume_session`.

It also pins the SDK guarantees the design leans on (UUID and traversal validation, safe subpaths, refresh-token redaction, ephemeral temp-dir cleanup, 0600 perms) by calling those helpers directly.

No isolation gaps found.

Stacked on the SS-5 branch. Final PR in the set.

### DIFF
--- a/tests/cloud/test_multitenant_session_isolation.py
+++ b/tests/cloud/test_multitenant_session_isolation.py
@@ -1,0 +1,110 @@
+# tests/cloud/test_multitenant_session_isolation.py
+# Created 2026-06-30 (feat/session-supervisor SS-7) — the durable (Mongo) half of
+# the consolidated, ADVERSARIAL multi-tenant session-isolation GATE. The OSS half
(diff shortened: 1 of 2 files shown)


prakashUXtech:

PR #1613, opened 2026-07-02
test(atlas): eval re-baselined on the full model + integration fixes

prakashUXtech:
## What this is now

Originally the standalone eval harness; retargeted to the top of the atlas stack (base: feat/atlas-fabric) and now carries the integration tail — merge this LAST.

## What ships

- **Eval harness** — 22 real-user intents in `tests/atlas/eval_cases.json`, parameterized pytest with a two-directional strict-hit baseline gate (regressions fail, unrecorded improvements fail too, so the constant stays honest).
- **Re-baseline against the full 237-entry model** — the original 16/18 was measured on the 31-entry hand seed. Every shifted case was re-measured with real searches and decided per-case (decision notes live on the cases): three re-pointed to genuinely better specific answers (kanban/chart/gcalendar over the generic primitive), two were real vocabulary gaps fixed in the authored seed ("roll back", "persist"), and the fabric stemming xfail was promoted per its own rule. **New measured baseline: 21/22 strict hits.** Four new cases cover the widget/skill/connector/sense kinds.
- **EE test-expectation fix** — with pocketpaw_ee installed, 7 tests in `tests/ee/test_mcp_claude_sdk.py` failed because the always-on `pocketpaw_atlas` server wasn't in their strip-list; added it the same way every prior ambient server was (belt, media, fabric, instinct).

## Integration verification (run on this exact tree)

- `tests/atlas/` 130 passed (ee installed) · `tests/ee/` 1724 passed, 2 skipped · flag-mode lane (`POCKETPAW_REQUIRE_WORKSPACE_SCOPE=1`) 1854 passed.
- `tests/cloud/`: 146 failures — proven pre-existing by a controlled experiment (reverting all atlas-touched files reproduces byte-identical failure counts; the suite is `--ignore`d in pyproject addopts and has no CI lane). Flagged for separate triage.
- Live smoke: server booted from this tree, real agent turn called `atlas_search`/`atlas_describe`, answer carried correct primitives + surface routes; tenant scope visible in the warm-client cache key.

### DIFF
--- a/docs/atlas.md
+++ b/docs/atlas.md
@@ -36,6 +36,15 @@ synthetic `fabric:<entity-type>` cards for the CALLING workspace's live
 ontology and atlas_describe answer `fabric:<type>` with properties +
 links. Never compiled into atlas.json; absent/erroring introspector →
 gracefully absent (OSS default).
+Updated: 2026-07-02 (feat/atlas-eval, AT-2) — added the "Eval" section:
+intent→capability ranking-regression harness (tests/atlas/eval_cases.json
++ tests/atlas/test_eval.py), measured strict-hit baseline 16/18.
+Updated: 2026-07-03 (integration/atlas) — re-baselined the eval against
+the full compiled model (237 entries) + fixpoint stemmer: baseline now
+21/22; fabric xfail promoted; three cases re-pointed at the more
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1615, opened 2026-07-02
feat(atlas): surface map + always-on OS primer

prakashUXtech:
## What ships

Agents learn the frontend exists (stacks on #1612, sibling of #1613):

- **21 surface entries** in the atlas seed, one per real paw-enterprise route (verified against `src/routes/`): /, /chat, /pockets, /sites, /belt, /paw-print, /decisions-graph, /mission-control, /agents, /settings, /settings/workspace(/integrations), /knowledge, /files, /studio, /code, /foresight, /calendar, /meetings, /activity, /audit.
- **Primitive → home-route cross-links**: pocket→/pockets, instinct→/paw-print, connector→/settings/workspace/integrations, sites→/sites, belt→/belt. `atlas_describe` answers now carry the route, so agents can say "see it at /sites" after an action.
- **Paw OS Primer** — new always-on block in `context_builder` (same priority + try/except pattern as the skills block): OS identity, one line per primitive generated from the atlas store at build time, and the instruction to call `atlas_search` before guessing and to point users at surface routes. Measured ~1.6K chars (≈400 est. tokens), hard-capped at 2000 chars by the injection budget; atlas load failure degrades to no block, never a broken prompt.

Two honest gaps found while mapping: no dedicated billing route exists (plan info lives on /settings/workspace — the entry says so), and Instinct's decision surfaces are /paw-print + /decisions-graph, not a /decisions route.

## Testing

`tests/atlas/` + `tests/test_agents_md.py` → 53 passed. Primer tests are hermetic (mock memory manager, no machine-global config reads). ruff clean. Reviewed (spec + quality): pass; review's one important finding (non-hermetic tests) fixed before this PR, truncation cosmetics cleaned up.

Known accepted: the primer mentions atlas_search on all backends while the MCP server registers on the SDK backend — same exposure class as the existing skills block.

### DIFF
--- a/docs/atlas.md
+++ b/docs/atlas.md
@@ -1,0 +1,65 @@
+<!--
+docs/atlas.md — the atlas primitive: the runtime OS self-model and its
+two agent tools (atlas_search / atlas_describe).
+
+Created: 2026-07-02 (feat/atlas-core, AT-1) — first end-to-end slice:
+hand-authored paw.atlas/v1 seed (10 primitives), AtlasStore
+loader/search/describe, and the pocketpaw_atlas in-process MCP server
+registered on the Claude Agent SDK backend next to pocketpaw_widgets.
+-->
+
+# Atlas — the OS self-model
+
+Atlas is the runtime's self-model: a hand-authored capability map the
+product's runtime agents (chat agent, pocket specialist) query to learn what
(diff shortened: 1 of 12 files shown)


prakashUXtech:

PR #1617, opened 2026-07-02
feat(shield): real-client-IP access log for CrowdSec (opt-in)

prakashUXtech:
## What

Adds an opt-in HTTP access log to `pocketpaw serve` so a co-located paw-shield / CrowdSec can see traffic. When `POCKETPAW_ACCESS_LOG_PATH` is set, an ASGI middleware appends one nginx-combined line per request to that file; when it's unset (the default), the middleware isn't even added, so existing deploys are unchanged.

## Why

The backend sits behind Cloudflare → Traefik, so the socket peer is a proxy. CrowdSec needs the **real** client IP or it would fingerprint and ban the proxy itself. The middleware resolves the client by precedence: `CF-Connecting-IP` → leftmost `X-Forwarded-For` hop → socket peer. nginx-combined format so the maintained `crowdsecurity/nginx` collection parses it as-is.

## Notes

- Raw ASGI middleware (mirrors `AuthMiddleware`), so WebSocket and lifespan scopes pass through untouched — only HTTP is logged.
- A logging failure can never break a request: the write is wrapped, warns once, and is swallowed in a `finally`.
- Part of the paw-shield containerized-deploy work; the deploy overlay that mounts the log volume + points CrowdSec at it lands separately. This PR is safe on its own (off by default).

## Test

`tests/test_api_access_log.py` — 9 tests covering the IP precedence (CF / XFF / peer), the nginx-combined shape, query-string handling, write-failure safety, and that the middleware is absent when the path is unset. All green locally.

### DIFF
--- a/src/pocketpaw/api/access_log.py
+++ b/src/pocketpaw/api/access_log.py
@@ -1,0 +1,158 @@
+# nginx-combined HTTP access-log middleware for ``pocketpaw serve``.
+#
+# What this does:
+#   A raw ASGI middleware (matching the AuthMiddleware pattern in
+#   dashboard_auth.py — no BaseHTTPMiddleware, so WebSocket scopes pass through
+#   untouched) that, for every HTTP request, appends ONE nginx-combined-format
+#   line to ``Settings.access_log_path``. shield's CrowdSec parses that file to
+#   detect probing/scanning.
+#
+#   The backend sits behind Cloudflare → Traefik, so the socket peer is a proxy.
+#   The log records the REAL client IP resolved by precedence:
+#     CF-Connecting-IP header → leftmost hop of X-Forwarded-For → peer host.
+#   Recording the proxy IP would make CrowdSec ban the proxy (self-DoS).
+#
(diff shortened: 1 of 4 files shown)


prakashUXtech:
Rebased onto current dev (was 264 behind). The only conflict was the config.py changelog docstring, resolved by keeping both entries in date order. `tests/test_api_access_log.py` passes 9/9 against dev, ruff clean. Reopening because the stale bot closed this one, not a human.

prakashUXtech:
Rebased onto current dev (was 365 behind) — clean this time, no conflict. Diff is unchanged: the same 4 files, 411 insertions, nothing removed.

Verified on the rebased commit `030eb944`: 9/9 tests pass, ruff + format clean.

Still held for review, not merging. Worth flagging why this one matters: the Coolify shield overlay is already on main and wires `POCKETPAW_ACCESS_LOG_PATH` + points CrowdSec at `/logs/access.log` — but nothing writes that file until this lands, so an enabled shield stack currently sees no traffic. This PR is the missing half.

PR #1616, opened 2026-07-02
feat(atlas): compiled artifact — connector + senses extraction, CLI, drift check

prakashUXtech:
## What ships

atlas becomes a compiled, deterministic artifact (stacks on #1615):

- **Authored/compiled split** — hand-authored entries move to `atlas/authored/{primitives,surfaces}.json` (verified field-identical through the move); `data/atlas.json` is now the compiled output, checked in like a lockfile with a `generated: true` header. 72 entries total.
- **Connector extraction** — one `connector:*` entry per YAML (35): actions with descriptions and key params in the narrative, declared senses, searchable keywords. `atlas_search("stripe invoices")` now top-hits stripe with `list_invoices` in the detail — the "agent doesn't know what a connector can do" gap is closed.
- **Senses extraction** — 6 `sense:*` entries from CORE_SENSES, each cross-linking its declaring connectors (e.g. `sense:paw.code.v1` → github, gitlab).
- **`pocketpaw atlas build`** CLI (+ `--check` for CI: exit 1 with a diff summary when the artifact is stale). `--check` added to the existing lint job in ci.yml.
- **Startup drift check** — first store load compares compiled vs live connector names; mismatch logs a warning, never raises. Stays quiet when the process can't see any connector YAMLs at all (installed wheel / foreign CWD) — that's an environment fact, not staleness.

## Testing

`tests/atlas/ + tests/test_agents_md.py` → 72 passed; determinism pinned by test AND proven live (built twice, diff empty); `--check` exercised against a genuinely tampered artifact (exit 1) and fresh one (exit 0); ruff clean. Reviewed (spec + quality): pass — the one important finding (drift false-positive with empty live set) fixed here; review verified all 35 connector entries match their YAMLs and the authored move byte-for-byte.

Known minors accepted: hypothetical duplicate connector `name:` across YAMLs isn't deduped by the compiler (registry dedupes; curated dir); extraction tests assume repo-root CWD (matches CI).

PR #1618, opened 2026-07-02
feat(atlas): per-workspace live overlay with fail-closed entitlement filter

prakashUXtech:
## What ships

atlas answers now reflect the calling workspace, not a global view (stacks on #1616):

- **EntitlementProvider protocol + AtlasOverlay** — entries a provider doesn't grant are removed (absent from search, describe, and the known-ids listing; a filtered id is byte-identical to an unknown one — no existence oracle). Connector entries gain `available: true/false` from the real connector seam: `ConnectorRegistry.status(scope_key)`, the same durable-state view `connector_list` answers from. Available connectors rank above unavailable ones only at equal relevance — base scoring untouched.
- **Per-run scope, no process-globals** — the SDK backend resolves the scope from its own run env (`ws:<POCKETPAW_WORKSPACE_ID>` when an isolated cloud run attached tenancy, else the OSS `default` scope). Tenancy attached with a blank id fails closed to a sentinel scope that matches no rows.
- **Warm-client hardening** — the tenant scope is folded into the client cache key, so a scope change structurally forces a fresh subprocess (extends the ART-2 cwd argument).
- **`connect_hint`** — describing an unavailable connector points at the integrations surface, with the route looked up through the overlay so a filtering provider can't leak it.

## Review trail

Spec+quality review: pass. Adversarial security audit: one important finding — blank-workspace-id fail-open toward the shared default scope — **fixed here** (sentinel scope + tests), plus its two defensive recommendations applied (cache-key fold, overlay-routed hint). Audit verified: no store mutation/poisoning, no cross-tenant oracle in error envelopes, scope-key injection contained by the state store's segment sanitizer, no shared-backend route between tenants.

## Honest v1 scope

Two things this does NOT yet do, on purpose:
- Cloud `ws:` availability is plumbed but inert — the cloud connector state store doesn't enumerate tenant rows, so cloud scopes conservatively report all connectors unavailable until the EE availability provider lands (docs + code comments say so explicitly).
- Entitlement filtering is dormant: the OSS default provider grants everything. The filter paths are built and tested; a plan-aware EE provider is the follow-up (tracked with the Fabric slice).

## Testing

119 passed: 23 overlay tests (two-context divergence, fail-closed on raising AND ambiguous grants, no-leak describe, both re-rank cases, no singleton mutation) + 3 new scope-key tests + full atlas suite + claude_sdk allow/deny/warm-lease/prewarm consumers. ruff clean. Live smoke on this machine: genuinely-connected connector shows available, stripe shows unavailable.

### DIFF
--- a/docs/atlas.md
+++ b/docs/atlas.md
@@ -16,6 +16,12 @@ file is now a COMPILED artifact built by `pocketpaw atlas build` from
 hand-authored sources (atlas/authored/) + extracted connector and sense
 entries; CI gates freshness via `atlas build --check`; the store logs a
 stale-artifact WARNING on connector drift at first load.
+Updated: 2026-07-02 (feat/atlas-overlay, AT-5) — live overlay +
+fail-closed entitlement filter: atlas answers now reflect the CALLING
+workspace via a per-run EntitlementProvider (atlas/overlay.py) —
+connector cards carry `available`, unavailable connectors rank below
+available ones at equal relevance and describe points them at the
+integrations surface, and non-granted entries are absent everywhere.
 -->
 
 # Atlas — the OS self-model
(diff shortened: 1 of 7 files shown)


prakashUXtech:

PR #1624, opened 2026-07-03
feat(files): agent library verbs — tag, move, annotate, search

prakashUXtech:
Agent library verbs for the /files prod-ready arc (FL-3) — the piece that makes the Library agentic rather than just searchable.

**Stacked on #1623 (FL-2)** — base is `feat/files-versioning-spine` (which carries FL-1 + FL-2); the diff here is only FL-3's changes.

## What ships

Four builtin tools the agent can call over the Library:
- `tag_file(file_id, tags)` — add/remove tags (persists via FL-1's workspace-scoped setter). Metadata op → journal event.
- `move_file(file_id, folder_path)` — move to a folder. Metadata op → journal event.
- `annotate_file(file_id, note)` — prepend a note. Content mutation → writes a new revertable version via FL-2.
- `search_library(query, limit)` — BM25 search over the workspace's kb-go scope, ranked hits with snippets.

Mutating verbs are `trust_level="medium"` (not auto-approved); search is `"high"`. All four are workspace-scoped with cross-workspace denial tested.

## Load-bearing design note (surfaces a cross-slice seam)

FL-1 Library rows key `FileUpload` by a **bare `file_id`** (globally-unique index). FL-2's editor path keys by the file's **path** (`${workspace}:${path}` internally). Two schemes on one collection. `annotate_file` bridges them — `annotate_upload()` reuses FL-2's archive-then-rewrite mechanism but keyed on the Library row's own `file_id`, so annotations land as real, revertable `FileVersionDoc` rows.

**Follow-up (tracked separately):** the FL-7/8/9 editor calls `PUT /files/{id}` with the Library `file_id`, which the path-based FL-2 handler does not yet resolve — so editing a Library file end-to-end needs the same bridge on the update/list/revert routes, plus a real upload→edit→version→revert integration test. Filed as the next task.

## Tests

`tests/cloud/test_library_verbs.py` (12) + `tests/cloud/file_versions/` (13 regression) → 27 passed. Covers tag add/remove + journal, move + missing-folder reject + journal, annotate writes a revertable version + non-text reject, search scoped to `workspace:{id}`, cross-workspace denial for all four, and fail-closed with no workspace bound. `lint-imports` root + ee → 0 broken (FileVersions contract kept).

PR #1625, opened 2026-07-03
fix(files): resolve Library file_id in the editor version path

prakashUXtech:
Fixes a cross-slice bug that every green PR in this arc hid (FL-16).

**Stacked on #1624 (FL-3)** — base is `feat/files-agent-verbs`; the diff here is only FL-16's changes.

## The bug

The Library and the editor version path key files two different ways on the same collection:
- FL-1 Library rows key `FileUpload` by a **bare `file_id`** (a globally-unique uuid).
- FL-2's editor path (`update_file_content` / `PUT /files/{id}`, list versions, revert) treats the client id as a **path**, namespacing internally to `${workspace}:${path}`.

So when the editor (FL-7/8/9) opens a Library file, it holds the bare uuid and calls `PUT /files/{uuid}` — which the path-based handler resolved to nothing. **Editing a real Library file end-to-end was broken**, even though FL-2's own tests (path-based) and the frontend's tests (mocked fetch) all passed. The seam was invisible to every isolated test.

## The fix

A single `_resolve_upload(workspace, file_id)` helper: try the editor's namespaced id first (preserves the FL-2 path flow exactly), then fall back to a Library row keyed by its bare `file_id`. Both lookups pin `workspace`, so tenant isolation holds. `FileVersionDoc` still keys on the client-facing id inside `service.py` (import-linter contract intact), so list/get/revert find archived rows for either scheme. The router passes `file_id` straight through, so the HTTP path inherits the fix with no router change.

## The test that proves it (no mocks at the seam)

`test_editor_library_bridge.py` — seeds a **real** `FileUpload` Library row by its bare uuid and drives `update_file_content` / `list_versions` / `revert_to_version` against it; only blob-IO is faked (a legitimate storage seam).

- **Red-before** (fix stashed): the two bridge tests fail with `NotFound: file not found` — exactly the bug.
- **Green-after**: `tests/cloud/file_versions/` + `test_library_verbs.py` → 31 passed. `lint-imports` root + ee → 0 broken. Cross-workspace access denied (tested).

### DIFF
--- a/ee/pocketpaw_ee/cloud/file_versions/service.py
+++ b/ee/pocketpaw_ee/cloud/file_versions/service.py
@@ -30,6 +30,21 @@
 #   ``get_version``. A stale ``If-Match`` now raises ``PreconditionFailed``
 #   (412) instead of ``ConflictError`` (409), matching HTTP conditional-request
 #   semantics for the file-version write path.
+# Updated: 2026-07-03 (FL-16, editor↔Library bridge) — the editor version path
+#   (``update_file_content`` and, transitively, ``revert_to_version``) now
+#   resolves BOTH id schemes via the new ``_resolve_upload`` helper: it tries the
+#   editor's workspace-namespaced id (``${workspace}:${path}``) first, then falls
+#   back to a Library ``FileUpload`` row keyed by its OWN bare ``file_id`` (a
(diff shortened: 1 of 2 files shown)


prakashUXtech:

PR #1626, opened 2026-07-03
feat(files): organize_folder batch executor — capped, metered, revertable

prakashUXtech:
The folder-batch executor — the headline Poly-style demo (FL-4): "add a one-line summary to every file in this folder," run as capped, metered, revertable operations.

**Stacked on #1625 (FL-16)** — base is `feat/files-editor-library-bridge`; the diff here is only FL-4's changes.

## What ships

`organize_folder` — a builtin tool that fans out FL-3's single-file verbs (`annotate` / `tag`) over every file in a folder:

- Enumerates the folder's files workspace-scoped, applies the inner verb to each as a **separate metered invocation** (reuses FL-3 internals, no duplication).
- **Hard cap `max_files` (default 200).** Over the cap, it refuses with a clear count-vs-cap message and runs nothing — no silent truncation.
- **Partial-failure isolation:** one file's error doesn't abort the batch; per-file results are collected and a `folder.batch` summary is emitted.
- Annotations write revertable versions (via FL-3 → FL-2), so a batch is undoable file-by-file.

Registered in `_EE_TOOLS`, trust level `medium` (mutating, not auto-approved).

## Tests

`test_folder_batch.py` (10) + FL-3/versions regression → 41 passed. Covers: cap refusal with zero side effects, admin-may-raise / non-admin-cannot, partial failure (2 text succeed + 1 binary fails → both text reach v2, binary untouched), and cross-workspace isolation. `lint-imports` root + ee → 0 broken.

## Follow-ups (flagged, not guessed)

- **Admin cap-raise is dormant** until EE wires a `current_is_admin` marker into `attach_agent_identity` — there's no admin ContextVar today, so `_current_is_admin()` reads an optional getter and defaults to `False`. The default 200 cap works for everyone now; the raise path is implemented and tested against that seam but won't unlock until EE provides the marker. No fake API was committed.
- Over-cap detection scans up to 10000 workspace rows; a folder-scoped query would be cleaner if libraries grow very large.

### DIFF
--- a/src/pocketpaw/tools/builtin/__init__.py
+++ b/src/pocketpaw/tools/builtin/__init__.py
@@ -17,6 +17,9 @@
 #   - 2026-07-03: Added FL-3 Library verbs (EE) — TagFileTool, MoveFileTool,
 #     AnnotateFileTool, SearchLibraryTool — the single-file agent verbs that
 #     make /files agentic. Registered in _EE_TOOLS (guarded by ee/ availability).
+#   - 2026-07-03: Added FL-4 OrganizeFolderTool (EE) — the folder-batch executor
+#     that fans an FL-3 verb (annotate | tag) over every file in a folder as
+#     separate metered invocations, capped + partial-failure isolated.
 
 import importlib as _importlib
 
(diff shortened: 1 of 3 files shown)


prakashUXtech:

PR #1627, opened 2026-07-03
feat(files): document, slides, and spreadsheet edit tools (port #1193)

prakashUXtech:
The document/slides/spreadsheet edit tools — the last dewani12 harvest (FL-5, port of #1193).

**Stacked on #1626 (FL-4)** — base is `feat/files-folder-batch`; the diff here is only FL-5's changes.

## What ships (ported from #1193, credit: dewani12)

Three builtin agent tools, each writing a **revertable version** via FL-2:
- `edit_document` — Editor.js block engine + normalization + transport.
- `edit_slides` — reveal.js deck engine.
- `edit_spreadsheet` — Univer workbook engine + A1 parsers.

Registered in `_EE_TOOLS` (trust `medium`). Backend: `read_current_content` (the read half the tools need, resolving both id schemes tenant-filtered), the restored `*-edit` / `*-context` REST routes, and the `AiEditRequest`/`AiEditResponse`/`SyncEditingContext` DTOs.

Each edit writes via `update_file_content(editor_kind="agent")` — pre-edit content recoverable through `list_versions`/`get_version`, live blob carries the edit, `content_version` bumps, cross-workspace denied.

## Tests

`test_edit_tools.py` (15) + file_versions/library regression → 46 passed. `tests/cloud/file_versions/ + test_library_verbs.py + chat/` → 157 passed (confirms the `CloudAgentChatRequest` change didn't break chat). `lint-imports` root + ee → 0 broken.

## Known follow-up (documented, not a regression)

The chat-side `editor_blocks` / `slides_data` / `spreadsheet_snapshot` transport is a **shim**. In #1193 the chat path injected these into the agent's system prompt via an inline `_run_agent_stream`; current dev replaced that with an out-of-process executor (`RunSpec` → executor → worker), so there's no router-time prompt-injection counterpart. The fields parse and round-trip on `CloudAgentChatRequest` but are **not** threaded through `RunSpec` → worker prompt assembly yet — a follow-up. **The working editor round-trip today** runs through the REST `*-edit`/`*-context` routes (inline `pool.run`, present on dev) + the module-store transport + the registered edit tools (edit-by-file_id, revertable — the primary surface, tested).

PR #1628, opened 2026-07-03
feat(files): auto-tag uploads on ingest

prakashUXtech:
Auto-tagging on ingest (FL-6) — uploaded files get suggested tags from their content, unless hidden from AI.

**Stacked on #1627 (FL-5)** — base is `feat/files-edit-tools`; the diff here is only FL-6's changes.

## What changes

The `FileReady` listener, after extraction, derives a small set of free-form tags and writes them to `FileUpload.tags`:

- `tagging.py` — dependency-free derivation: adapter-provided labels take precedence, then title, then frequency-ranked keywords from captions + text; normalized, deduped, capped at 8.
- **Union, not overwrite** — derived tags merge with any user-applied tags (a user tag survives re-index).
- **`hide_from_ai` gate** — a hidden file returns early, skipping BOTH KB ingest AND tagging. Tag-write failures are contained and never abort ingest.

Bonus: fixed a pre-existing Beanie test-isolation leak (torn-down mongomock db stayed bound to the Document classes, bleeding rows between tests) — this drops the pre-existing `tests/cloud/uploads/` failure count from 26 to 23.

## Tests

`test_auto_tagging.py` (15) + `test_listener_auto_tag.py` (4) → 19 passed here; the broader keyword set → 47 passed. Explicitly covers `test_hide_from_ai_skips_ingest_and_tags` and `test_reindex_unions_with_existing_user_tag`. `lint-imports` root + ee → 0 broken.

## Follow-up (hardening in FL-11b)

The hide gate is currently fail-*open* on indexing (if the upload row can't be loaded at all, indexing proceeds). For a privacy control this should be fail-*closed* — FL-11b will harden it so an unresolvable hide-status skips indexing.

### DIFF
--- a/ee/pocketpaw_ee/cloud/uploads/listeners.py
+++ b/ee/pocketpaw_ee/cloud/uploads/listeners.py
@@ -19,6 +19,16 @@
 #   vector path inherits the same scope variable so embeddings land in
 #   the same kb-go scope as the text article. Workspace uploads (no
 #   ``pocket_id``) keep the previous ``workspace:{wid}`` behaviour.
+# Updated: 2026-07-03 — FL-6 "Auto-tagging on ingest". The listener now
+#   (1) loads the FileUpload row up front and, if ``hide_from_ai`` is set,
+#   returns early — a hidden file is neither KB-indexed nor tagged (this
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1629, opened 2026-07-03
feat(files): public share links with forced-attachment downloads

prakashUXtech:
Public share links (FL-12b) — a file owner mints a link that lets a non-authenticated recipient download the file, safely.

**Stacked on #1627 (FL-5)** — base is `feat/files-edit-tools`; the diff here is only FL-12b's changes.

## Endpoints

- `POST /files/{id}/share` — owner/admin creates a link (license + `uploads.write` gated). Returns the token URL + 7-day expiry.
- `GET /share/{token}` — **PUBLIC, no auth.** The unguessable token is the authorization. Validates active → mints a download via `presigned_get` → 302 redirect. Unknown → 404, revoked/expired → 410.
- `DELETE /files/{id}/share/{token}` — owner/admin revoke.

Feature-flagged behind `POCKETPAW_SHARE_LINKS_ENABLED` (**default OFF**); when off, both surfaces are a flat 404.

## Security posture (reviewed — /security-review passed, no P0/P1)

- **Forced-attachment inherited, not bypassed.** The public route mints through the same `presigned_get` that forces `Content-Disposition: attachment` for non-inline mimes (ART-4), so agent-authored `.html`/`.svg`/`.js` downloads instead of rendering active content on the storage origin. Double-locked by `test_presigned_disposition.py` + the new e2e test.
- **Token:** `secrets.token_urlsafe(32)` (~256 bits), unique-indexed, single-file scoped — a token for file A can never resolve file B.
- **No leak:** flat 404/410, opaque redirect; the route never exposes workspace/owner internals and never requires auth beyond the token. Expired/revoked/unknown tokens never call `presigned_get`.
- **Owner replay identity:** the public route uses the link's stored `owner_id`/`workspace_id` as the read identity so `presigned_get`'s ACL passes — scoped to the one `link.file_id`, re-checks the file still exists.
- **Fails closed:** 503 (not an authed URL) when no presigning backend is configured.
- Independent of `hide_from_ai` (separate gates by design).

## Tests

`test_share_links.py` — 13 passed. Explicitly: HTML→forced-attachment, image→inline-ok, token-A-can't-fetch-B, expired→410, revoked→410 (presign not called), unknown→404, deleted-file→404, cross-workspace create/revoke scoping, flag-off-dark, share-independent-of-hide, token-unguessable-unique. `lint-imports` root + ee → 0 broken.

## P2 follow-ups (non-blocking)

- Add rate limiting to the public GET (defense against enumeration/DoS; the 256-bit token already makes brute-force infeasible).
- The revoke route duplicates the owner-or-admin check inline; could reuse `_assert_can_write` for consistency.
- No TTL cleanup job — expired rows are inert (`is_active()` rejects them); a Mongo TTL index on `expires_at` could prune them.

### DIFF
--- a/ee/pocketpaw_ee/cloud/__init__.py
+++ b/ee/pocketpaw_ee/cloud/__init__.py
@@ -1,5 +1,12 @@
 """PocketPaw Enterprise Cloud — domain-driven architecture.
 
+Modified: 2026-07-03 (feat/files-share-links, FL-12b) — Mounts two share-link
+    routers: ``share_router`` (owner + license gated POST/DELETE
+    /files/{id}/share) and ``public_share_router`` (intentionally
+    unauthenticated GET /share/{token} — the token is the capability). The
+    public route mints its download URL through EEUploadService.presigned_get so
+    it inherits the forced-attachment protection for non-inline mimes. Both are
+    feature-flagged behind POCKETPAW_SHARE_LINKS_ENABLED (default off).
 Modified: 2026-06-24 (integration/billing-credits, BC-6) — Mounts the
     Entitlements resolver router (``GET /entitlements`` -> the caller workspace's
     plan + features + monthly credit allotment). The plan CATALOG read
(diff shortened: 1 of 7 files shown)


prakashUXtech:

PR #1630, opened 2026-07-03
feat(files): fail-closed hide-from-AI indexing gate

prakashUXtech:
Fail-closed hide-from-AI indexing gate (FL-11b).

**Stacked on #1628 (FL-6)** — base is `feat/files-auto-tagging`; the diff here is only FL-11b's changes.

## What changes

FL-6 added the `hide_from_ai` gate but failed **open**: if the metadata row couldn't be loaded to check the flag, it proceeded to index. For a privacy control that's the wrong default — a lookup hiccup could index a file the user marked hidden.

This hardens the gate to fail **closed**: an unresolvable hide status (row absent or store unavailable) skips KB indexing and auto-tagging. A resolvable, unhidden row proceeds as before. Hiding never removes the file from listing/download — it only affects AI indexing.

## Tests

`test_listener.py` — 34 passed. New: `test_hidden_file_skips_index_and_tags` (hidden → no extraction, no ingest) and `test_fail_closed_when_row_unresolvable` (unresolvable row → skipped). The pre-existing collaborator tests now seed a resolvable non-hidden row via the shared helper (they relied on the old fail-open default). `lint-imports` ee → 28 kept, 0 broken.

## Scoped out (needs a kb-go feature first)

Purging content **already** indexed when a file is later hidden requires a `kb delete` subcommand in kb-go, which does not exist today (kb-go stores articles as files with a BM25 index + concept graph; a delete must remove all three). This PR guarantees hidden files are **never indexed going forward**; the retroactive purge is filed as a follow-up gated on adding `kb delete`.

### DIFF
--- a/ee/pocketpaw_ee/cloud/uploads/listeners.py
+++ b/ee/pocketpaw_ee/cloud/uploads/listeners.py
@@ -29,6 +29,14 @@
 #   writes the result back through ``MongoFileStore.set_library_metadata``.
 #   Tag derivation/write failures are contained: a broken tag write must
 #   never lose the KB ingest that already succeeded.
+# Updated: 2026-07-03 — FL-11b "hide-from-AI enforcement". Hardened the hide
+#   gate to fail CLOSED: if the FileUpload row can't be resolved to confirm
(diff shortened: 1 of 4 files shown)


prakashUXtech:

PR #1632, opened 2026-07-03
feat(files): purge KB article on hide-from-AI toggle

prakashUXtech:
Completes the "hide from AI" control: hiding an already-indexed file now purges its content from the knowledge base (FL-11b follow-up).

**Stacked on #1630 (FL-11b)** — base is `feat/files-hide-from-ai`; the diff here is only this change.

## Background

FL-11b (#1630) made indexing fail closed — a file marked hidden is never indexed *going forward*. But content indexed *before* it was hidden stayed in kb-go, because kb-go had no delete surface. That surface now exists (kb-go's new `kb delete`), so this wires the retroactive purge.

## What changes

- **Track the article.** `FileUpload` gains `kb_article_id` + `kb_scope` (backward-compatible defaults). The `FileReady` listener already extracts the article id after ingest — it now persists it (and the scope) to the row via a workspace-scoped `set_kb_article` setter. A tracking-write failure never breaks ingest.
- **Delete method.** `KnowledgeService.remove_article(scope, article_id)` shells to `kb delete` (resilient — logs and returns False on subprocess error, like the other kb calls).
- **Purge on toggle.** When `PATCH /uploads/{id}` flips `hide_from_ai` false→true on a tracked file, it calls `remove_article` then clears the tracking (so a later re-index re-tracks). Best-effort: a purge failure logs a warning but the hide flag still applies and the tracking is kept for a retry. No purge when the file wasn't indexed, was already hidden, or the flag didn't change. The file stays browsable and downloadable — only the KB content is removed.

## Tests

`test_kb_purge.py` — 12 passed: ingest records id+scope (workspace and pocket), hide-toggle purges with the tracked id+scope and clears tracking, untracked/already-hidden/no-change/un-hide paths do NOT purge, a purge failure still hides and keeps tracking, store round-trip + workspace isolation, and `remove_article` maps to `kb delete` and swallows errors. `lint-imports` root + ee → 0 broken.

Also includes a small test fix carried from the rebase: two listener test files now seed a row for FL-11b's fail-closed gate (full listener suite: 32 passed).

## Depends on

kb-go's `kb delete <id> --scope <s>` (separate PR). The purge is unit-tested against that CLI contract with a mocked subprocess, so it verifies without the kb-go build present.

### DIFF
--- a/ee/pocketpaw_ee/cloud/agents/knowledge.py
+++ b/ee/pocketpaw_ee/cloud/agents/knowledge.py
@@ -1,4 +1,10 @@
 # knowledge.py — Agent knowledge service via the kb-go binary.
+# Updated: 2026-07-03 — FL-11b "hide-from-AI purge". Added
+#   KnowledgeService.remove_article(scope, article_id) → `kb delete
+#   <article_id> --scope <scope>`, mirroring get_article. Resilient: logs and
+#   swallows subprocess errors (returns False) like the other kb calls, so a
+#   purge failure never propagates back into the PATCH handler that hides a
+#   file. kb-go's `delete` is idempotent (deleting a missing id is a no-op).
 # Updated: 2026-04-30 — Stage 1.B of "Files as Knowledge". Added
 #   ingest_text_to_scope so callers (notably the FileReady listener) can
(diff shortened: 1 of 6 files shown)


prakashUXtech:

Issue #1633, opened 2026-07-03
Cap assignable member role at the actor's own level (no ADMIN-grants-OWNER)


prakashUXtech:
`update_member_role` (`ee/pocketpaw_ee/cloud/workspace/service.py:691`) enforces an ADMIN minimum via the `workspace.member.role_change` action, but it does not cap the role an actor can assign. An ADMIN can set a member to OWNER, and two admins can promote each other. The REST route (`ee/pocketpaw_ee/cloud/workspace/router.py:198`) has allowed this since before the agent tools existed.

Surfaced during the WA-2 security review of the new agent-facing `member_update_role` tool. The tool path is actually stricter than REST (it requires a human to approve the change in the Instinct Tray), so this is not a regression the tool opens. But now that role changes are reachable through an AI agent, the missing ceiling is worth closing.

Fix in the shared service so both the REST route and the tool inherit it: reject an assignment where the target role exceeds the actor's own role. An ADMIN may grant up to ADMIN; only an OWNER may grant OWNER. The existing owner-demotion and last-owner guards stay as they are.

Issue #1631, opened 2026-07-03
Generalize the Instinct approve/execute spine into a pluggable executor registry


prakashUXtech:
Today the Instinct gate dispatches an approved proposal by branching on which bespoke blob key is present in `Action.parameters` (`_external_action`, `_pocket_write`, `_code_change`, `_fabric_objects`, `_pocket_create`, `_instinct_rule`, `_artifact_change`). Each kind has its own hardwired executor, and the router runs a fixed N-way if/elif on the key. Adding a new proposal kind means editing the router's dispatch, the approve path, the reject path, and the bulk path in lockstep.

We just hit this while shipping workspace-admin tools, which need an 8th `_admin_action` kind so an agent-proposed admin change (a role change, a plan change) executes when a human approves it in the Tray.

It works, but every new gated-write surface repeats the same four-place wiring. The proposal here: refactor the dispatch into a registry where each kind registers its blob key and executor once, and the router looks up the executor instead of branching. The external-action and admin-action executors would then share one approve/execute spine.

Not urgent. We are shipping the `_admin_action` kind as a dedicated executor now, since that matches the existing precedent and carries the lowest risk. This issue tracks the consolidation for when there are enough kinds to justify it.

Reference: `ee/pocketpaw_ee/instinct/router.py`, `ee/pocketpaw_ee/cloud/external_actions/executor.py`.

PR #1634, opened 2026-07-03
feat(ee): workspace-admin agent tools — RBAC-gated, Instinct-approved, role-filtered

prakashUXtech:
A complete agent-facing workspace-administration surface: ~15 tools the runtime chat agent can use to run a workspace, gated by the existing RBAC matrix, with every mutation routed through the Instinct approval gate, and discovery filtered by role so members don't see admin tools they can't use. Design: `docs/design/drafts/2026-07-03-workspace-admin-tools.md`. Stacks on the atlas PR chain (base `feat/atlas-eval`).

## The three layers

- **RBAC is the execution lock.** Every tool calls `check_workspace_action(user, workspace_id, action)` — the existing, tested, audited matrix. Reads execute directly once the gate passes; a failed gate returns a structured deny envelope, never a crash.
- **The Instinct gate is the second lock on writes.** No agent-initiated mutation fires inline. Each write files an `_admin_action` proposal into the Tray; a human approves; the change fires only then, with RBAC **re-checked at execute time** (a proposer demoted after proposing is blocked).
- **Atlas discovery is role-filtered.** A member's `atlas_search("manage the workspace")` doesn't surface admin tools — a role-aware entitlement provider hides `role:*` capability cards the caller can't reach, fail-closed (unknown role → hidden). This is discovery only; the RBAC checks are the real gate.

## Slices

- **WA-1/2** — admin MCP server + the `_admin_action` Instinct gate kind (an 8th proposal kind) + `member_update_role` as the first human-approved gated write. Live-smoked end to end (propose → approve → execute flips the role; a member is denied).
- **WA-4** — 5 read tools (members, settings, connectors, billing usage, audit), each gated to the tier its REST route actually uses.
- **WA-5** — 7 ADMIN writes (member remove, invite create/revoke, connector enable/disable/config, workspace update), each a whitelist entry with a strict arg adapter.
- **WA-6** — 3 OWNER writes (instinct approval level, workspace delete, billing plan change). Billing executes as **propose → Dodo checkout URL**, never a plan mutation — the payment-bypassing `set_workspace_plan` is webhook-only and unreachable from any agent path. **seats_manage was deliberately skipped** — no seat-change service exists; faking one would mean a silent DB write bypassing billing.
- **WA-3** — role-aware atlas provider + 17 admin capability cards (role-tiered to match each tool's gate). Includes a scoped stopword fix to `store._tokenize` (the admin card names were about to pollute already-shipped atlas ranking; adversarial review confirmed it removes that pollution with zero eval regressions and 4 improvements).

## Security

Two adversarial security passes (WA-2 gate kind; WA-5+WA-6 write surface) and a 4-dimension adversarial review workflow on WA-3 — all **PASS**. Verified: no execute-without-approval path, adapters don't smuggle kwargs, invite role capped at admin|member, the plan can't flip without payment, no ADMIN-gated tool reaches an OWNER effect, the discovery filter fails closed at every branch, and OSS never leaks admin cards (the default provider hides all `role:*` entries).

## Follow-ups filed (not silently shipped)

- #1631 — generalize the Instinct approve/execute spine into a pluggable registry.
- #1633 — cap assignable member role at the actor's level (an ADMIN shouldn't grant OWNER; pre-existing in REST, now human-gated on the agent path).
- Seats management has no service in the codebase — a product decision, flagged for the captain.

## Testing

Full admin + gate + atlas suites green (204 in the atlas/ee slice; instinct-router regression intact — the 7 existing proposal kinds unchanged). Artifact deterministic (`atlas build --check`), eval 21/22 held honestly. Live-smoked against real Mongo. Nothing merged — captain review gate.

### DIFF
--- a/ee/pocketpaw_ee/agent/atlas_provider.py
+++ b/ee/pocketpaw_ee/agent/atlas_provider.py
@@ -1,0 +1,189 @@
+# atlas_provider.py — role-aware atlas EntitlementProvider for the cloud chat
+#   agent (WA-3). Created: 2026-07-03 (feat/workspace-admin-tools).
+#
+# The DISCOVERY layer for the workspace-admin tools — NOT the security gate.
+# The RBAC checks INSIDE each admin tool (workspace_admin.py) are the real lock
+# and are already audited; this provider only decides what a member SEES in
+# atlas_search / atlas_describe, so a non-admin doesn't see admin capabilities
+# they can't use. It must NEVER be relied on as the execution gate.
+#
+# Shape: implements the core ``EntitlementProvider`` protocol (overlay.py):
+#   * ``is_granted(entry)`` — a role-BLIND entry (no ``role:*`` in requires)
+#     delegates to the wrapped ``DefaultEntitlementProvider`` (grant, subject to
+#     the existing connector/availability logic). A role-GATED entry is granted
+#     only if the CALLER's resolved workspace role >= the entry's required tier
+#     (owner > admin > member). FAIL-CLOSED: if the role can't be resolved (no
+#     identity, user not found, not a member, error) the role stays ``None`` and
+#     every role-gated entry is hidden. Never raises.
+#   * ``connected_connector_names()`` — pure delegation to the wrapped default
+#     provider (reuse the connector-state seam; don't reimplement).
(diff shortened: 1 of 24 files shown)


prakashUXtech:

PR #1641, opened 2026-07-04
Add a no-mutation Chat mode to the site refine surface

DevRohit06:
## What

The per-site chat at `/sites/[siteId]` treats every message as an edit to apply and republish. There's no way to just ask a question about a site ("what's on the pricing section?", "why does the hero say that?") without the agent going ahead and changing the live page.

This adds a Build/Chat mode to the surface so the sites builder can offer a toggle where Chat answers questions and leaves the site untouched.

## How

- `SurfaceMeta` (domain) and `SurfaceMetaRequest` (dto) grow a `mode` field: `"build"` (default) or `"chat"`. Default keeps current behavior.
- Threaded through `_meta_from_request` (DTO -> domain), next to the existing `engine` / `site_id` hints.
- New `_chat_preamble` in the sites handler: a read-only Q&A preamble. The agent answers questions about the existing site but does not call the pocket specialist edit/merge tool, does not republish, and does not create a pocket. If the user actually wants a change, it tells them to flip the toggle to Build.
- `build_preamble` branches on the refine path: `pocket_id` present + `mode == "chat"` -> chat preamble; otherwise the create/refine branching is unchanged. Build mode is today's behavior, byte-for-byte.

## Tests

Written first (failed red on the missing `mode` field), then implemented:

- `mode="chat"` + pocket_id yields a Q&A preamble that does not surface the edit tool, doesn't republish, and reads as question/answer.
- `mode="build"` (and unset) + pocket_id is identical to the existing refine preamble (regression guard).
- `mode="chat"` with no pocket_id leaves the create preamble unchanged.
- `mode` survives the DTO -> domain mapping, default `"build"`.

Full surface suite green: `78 passed`. Ruff check + format clean on touched files.

## Docs

Internal surface-handler + domain/dto change only; no CLI, runtime API, or memory-architecture surface touched, so no doc updates needed. The Build/Chat toggle UI is a separate frontend change.


DevRohit06:

PR #1643, opened 2026-07-05
fix(atlas): correct stale OS facts and sharpen intent search

prakashUXtech:
## What this fixes

atlas is the runtime OS self-model: the hand-authored map the agent reads to answer "where do I go for X" and "which capability does this". Four of its facts had drifted out of sync with the real OS, so the agent was pointing users at the wrong route. A separate batch of common intents ranked the wrong card first, in one case a destructive one.

Every claim below was checked against the live routes in `dev-frontend/src/routes` before editing, not taken on faith.

## Data accuracy

- **Billing.** atlas had no entry for `/settings/billing` even though it ships live with three tabs (Usage, Plans, Billing), the consumption graph, plan cards, and the wallet, and it's linked from the user menu, the plan badge, and the settings nav. Worse, the workspace-admin entry asserted "there is no dedicated billing route today." Added a `surface:billing` entry and deleted the false line.
- **Deep Work.** The operator/approvals surface moved off `/mission-control` (now a 308 redirect) to `/deep-work` on 2026-06-12, and the UI has always called it Deep Work. atlas still shipped "Mission Control" at the dead URL. Repointed the entry to `/deep-work`, renamed it, and added approvals/tray vocabulary. The id stays `surface:mission-control` (kept as an alias keyword) so nothing downstream breaks.
- **Security.** The owner-gated `/security` console (paw-shield) went live 2026-07-01 with no atlas entry at all. Added `surface:security` with a `role:owner` marker so the overlay hides it fail-closed from non-owners, matching the route's own gate.
- **Widget count.** A Ripple card claimed "~190 widgets"; the real catalog is 151. Corrected to "150+".

## Search relevance

- **"user" reaches the member cards.** The member capabilities only spoke "member", so "remove a user" missed them. Added user synonyms to the list, update-role, and remove cards.
- **A benign action can't rank below a destructive one.** "delete a user" used to surface the irreversible delete-workspace card first, because its name literally contains "delete". The scorer now damps the name-field weight by an IDF over entry names. A token like "workspace" that appears in 8 names no longer hands every card a full name hit and lets seed order pick the destructive action. Keyword, summary, and narrative weights are untouched.
- **Filler words stop hijacking names.** "set up memory" ranked the approval-level and follow-up cards because "set" and "up" hit their names. Added up, set, get, open, and show to the stopword list; the query now reaches the Soul primitive.
- **Orientation and provider actions.** Added "what can you do", help, and capabilities vocabulary to the home surface, and disconnect-provider vocabulary to the disable-connector action.

## Recompile and pins

Rebuilt the artifact (`pocketpaw atlas build`, 256 entries, `--check` green). Every fix is pinned with a new case in the intent-to-capability eval harness, so the strict-hit baseline rises from 21/22 to 30/31. Also added a guard test that flags brittle factual claims (negative route-existence, hedged counts) so a false narrative can't slip past a fidelity-only check again. It catches both of the bugs fixed here.

## Testing

- `uv run pytest tests/atlas/ tests/ee/agent/test_atlas_role_provider.py`: 170 passed
- `pocketpaw atlas build --check`: up to date
- `ruff format --check .` and `ruff check` on touched files: clean
- `lint-imports` and `lint-imports --config ee/pyproject.toml`: both green


prakashUXtech:
Landed via #1646 — the 3-deep atlas stack was collapsed into #1646 for a clean single rebase-merge. All commits from this PR are on dev.

PR #1645, opened 2026-07-05
fix(atlas): harden the compiler and fabric search against edge cases

prakashUXtech:
Five robustness defects in the atlas compiler (`compile.py`) and live Fabric search (`fabric.py`). All are code-only — the compiled `atlas.json` content does not change, so `pocketpaw atlas build --check` still reports up-to-date and no regenerated artifact is committed.

> Stacks on the `/files` atlas-data PR (`fix/atlas-data-accuracy-and-relevance`). The base of this PR is that branch, not `dev`, so the diff here is only the compiler-robustness changes. Merge the base first.

## A — compiler crashed on `display_name: null` (P2)

A connector YAML that parses cleanly but sets `display_name: null` took down the whole compiler — and the CI drift gate — with an `AttributeError` on `.lower()`. The dataclass default only fills an absent key, not a present null, so the null flowed straight into the f-strings. That contradicts the module's own "one bad file can't block the build" stance. Fixed by coercing to `defn.display_name or defn.name` before any string use.

## B — Fabric search ran a 3N sqlite N+1 per query (P2)

`search_entity_types` called `describe_entity_type` for every type on every search. Through the live EE adapter each describe opens three sqlite connections (`entity_exists` + `get_entity_properties` + `list_entity_links`), so a search was 3N connections — and the link list it paid for is never scored on. Search now reads property names only, through a new `list_entity_properties` path on the adapter (one connection), and falls back to describe for older introspectors. The thin search card drops the DB-derived link count; describe still carries links for the detail view.

## C — default connectors dir was CWD-relative (P3)

`DEFAULT_CONNECTORS_DIR` was `Path("connectors")`. Run `compile_atlas` / `write_artifact` / `check_artifact` with defaults from anywhere other than the repo root and it read zero YAMLs, silently dropped every connector and sense entry, and `write_artifact` would then persist a truncated artifact. Anchored to the package location instead, matching how `store.py` locates the packaged data file.

## D — search missed consecutive-capital CamelCase (P3)

The camel-split regex only broke lower→upper, so acronym names (`APIKey`, `HTTPServer`, `IOError`) never split and `api key` / `http server` queries missed them. Added an acronym-boundary alternative so `HTTPServer` → `http server` and `APIKey` → `api key`.

## E — drift diff hid duplicate-id entries (P3)

`check_artifact` built its per-id map with a dict comprehension, so two entries sharing an id collapsed into one: CI failed on bytes, but the printed added/removed/changed lists misreported why. `check_artifact` now surfaces duplicate ids explicitly, and `compile_atlas` asserts id uniqueness so an authoring collision fails loud at build time instead of shipping a silently-deduped artifact.

## Testing

- `uv run pytest tests/atlas -q` — 168 passed (11 new tests across the five findings, each red before the fix and green after).
- `uv run pocketpaw atlas build --check` — up to date; `atlas.json` is not modified.
- `ruff format --check`, `ruff check` on the touched files, and both `lint-imports` configs (main + `ee/pyproject.toml`) all clean.

### DIFF
--- a/src/pocketpaw/atlas/compile.py
+++ b/src/pocketpaw/atlas/compile.py
@@ -29,6 +29,22 @@
 #     installed skills change per install and are deliberately NOT baked
 #     into the artifact (their discovery stays with the system-prompt
 #     skills block).
+# Updated: 2026-07-05 (fix/atlas-compiler-robustness) — three robustness fixes,
+# all output-preserving (atlas.json bytes unchanged):
+#   * FINDING A: ``_connector_entry`` coerces ``display_name`` to a safe string
+#     (``display = defn.display_name or defn.name``) before ``.lower()`` / any
+#     f-string, so a connector YAML with explicit ``display_name: null`` no
+#     longer crashes the WHOLE compile (and CI drift gate) on AttributeError —
+#     honoring the module's "one bad file can't block the build" posture.
+#   * FINDING C: ``DEFAULT_CONNECTORS_DIR`` is anchored to the package location
+#     (``Path(__file__).parents[3] / "connectors"``) instead of a CWD-relative
+#     ``Path("connectors")``, so a default-arg compile / write / check from any
+#     dir still finds the connector YAMLs instead of silently dropping every
+#     connector + sense entry (and persisting a truncated artifact).
+#   * FINDING E: ``check_artifact`` detects duplicate entry ids explicitly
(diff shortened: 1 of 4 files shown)


prakashUXtech:
Landed via #1646 — the 3-deep atlas stack was collapsed into #1646 for a clean single rebase-merge. All commits from this PR are on dev.

PR #1651, opened 2026-07-06
Add an icon-search tool for site authoring

DevRohit06:
## What ships

An in-process `pocketpaw_icons` MCP server that gives the site-authoring agent a `search_icons` tool, so it can pull real open-source icons the same way it already pulls stock photos. Backed by the keyless Iconify public API (200k+ icons), it returns icon names plus real SVG URLs.

It's ambient — surfaced to the cloud chat agent without an opt-in — matching how the stock-images, media, and sites-manager servers already work. The agent only sees in-process MCP servers, so this is the layer icon search has to live in to be reachable.

## Why now

Icons are one of the cheapest, highest-leverage assets for making a generated marketing site look finished instead of generic. This lands the capability on its own so it's usable by the current site-create path immediately, ahead of the larger authoring-crew work that will lean on it.

## Scope

- `ee/pocketpaw_ee/agent/mcp_servers/icons.py` — the server, tool handler, and inlined Iconify search helper. Fails soft: empty query or a provider error returns an error response, never raises.
- `ee/pocketpaw_ee/extensions.py` + `ee/pyproject.toml` — a `CloudIconsMcpProvider` registered ambiently, exactly like the stock provider.
- `ee/pocketpaw_ee/agent/mcp_servers/__init__.py` — the module listing.
- `tests/ee/agent/test_icons_mcp_server/` — 13 tests, all network mocked (no live calls).

## Testing

`uv run pytest tests/ee/agent/test_icons_mcp_server/ -q` → 13 passed. Covers https SVG URLs, style-hint passthrough, empty-query and HTTP/connection errors going soft, limit default and cap, malformed-id skipping, tool-id namespacing, and ambient registration. `ruff check` + `ruff format` clean.

## Out of scope

No crew orchestration and no surface-preamble changes — the tool is registered and reachable; wiring it into a specific authoring stage comes with the crew work.


DevRohit06:

PR #1652, opened 2026-07-06
Extract a full color palette from a reference image

DevRohit06:
## What ships

An in-process `pocketpaw_palette` MCP server with an `extract_palette` tool: give it a reference image URL and it returns a coherent color palette across four roles (primary, secondary, tertiary, neutral), each as a **full 50–900 scale** rather than a single hex.

The full scale is the point. An agent handed real role scales composes far better UI than one handed three flat colors — it has the light tints for backgrounds and the dark shades for text already worked out, so a generated site reads as a chosen brand instead of generic.

How it works: download the image (httpx), thumbnail and quantize it (Pillow — already a dependency), assign dominant colors to roles, then expand each base into a 50–900 scale by holding hue and saturation and stepping lightness. Fails soft everywhere — a bad URL, a download error, or an unreadable image returns an error response, never a crash.

## Why now

Pairs with the icon-search tool as the second asset capability for site authoring, and it's independently useful to the current create path. Both are small, self-contained, and land ahead of the larger authoring-crew work that consumes them.

## Stacked on #1651

This branches off `feat/sites-crew-icons` because both PRs register a provider in `extensions.py` and `ee/pyproject.toml`; stacking keeps them conflict-free. Merge the base (#1651) with `--rebase`, then this rebases onto `dev`.

## Scope

- `ee/pocketpaw_ee/agent/mcp_servers/palette.py` — server, `extract_palette` handler, and a pure `scale_from_base()` helper.
- `ee/pocketpaw_ee/extensions.py` + `ee/pyproject.toml` — `CloudPaletteMcpProvider`, ambient like icons/stock.
- `ee/pocketpaw_ee/agent/mcp_servers/__init__.py` — module listing.
- `tests/ee/agent/test_palette_mcp_server/` — 17 tests, network mocked with a synthetic in-test PNG.

## Testing

`uv run pytest tests/ee/agent/test_palette_mcp_server/ -q` → 17 passed. Covers the scale helper (ordered light→dark, valid hex, hashless/shorthand/grayscale inputs), the synthetic-image happy path (4 roles × 50–900), empty/non-http/download/unreadable errors going soft, and tool-id namespacing. `ruff check` + `ruff format` clean.

## Out of scope

No crew orchestration or stage wiring — the tool is registered and reachable; consuming it in a Branding stage is later crew work.


DevRohit06:

PR #1653, opened 2026-07-06
Add a design-system library and retriever for site authoring

DevRohit06:
## What ships

A bundled library of five portable design systems plus a retriever tool, so the site-authoring agent can pick and adapt a coherent, professional look instead of inventing one from scratch on every prompt. This is the "better design system" foundation: each system is a `DESIGN.md` (machine-readable token scales + prose rationale) paired with a compiled `tokens.css` that both render engines can read as one source of truth.

The five, deliberately distinct across color, type, rounding, and elevation:

- **Northwind** (`clean-saas`) — cool indigo, restrained, trust-first
- **Monolith** (`bold-high-contrast`) — near-mono with an electric accent, brutalist
- **Hearth** (`warm-local-service`) — terracotta and sage, rounded and friendly
- **Obsidian** (`dark-premium`) — charcoal and champagne gold, editorial luxury
- **Quill** (`minimal-editorial`) — paper and ink, whitespace-led, serif display

All original — inspired-by, not brand clones — with full 50–900 color scales generated on a coherent lightness ladder, and `tokens.css` kept in lockstep with each `DESIGN.md`.

Why full scales and rationale matter: an agent handed real role scales, component states, and do's/don'ts composes a site that reads as a chosen brand. Handed three flat colors, it produces the generic centered-everything look. The library gives it grammar, not just paint.

## Why now

Third asset capability for site authoring, alongside icon search and palette extraction. Each system is retrievable today by the current create path, and the set grows by touch-time.

## Stacked on #1652

Branches off `feat/sites-crew-palette` because it also registers a provider in `extensions.py` and `ee/pyproject.toml`. Merge the chain base-first with `--rebase`.

## Scope

- `src/pocketpaw/bundled_design_systems/` — the five systems (`DESIGN.md` + `tokens.css` + `manifest.json` each) and a `bundled_design_systems_dir()` helper. Ships in the wheel via the existing `src/pocketpaw` include (verified in a built wheel).
- `ee/pocketpaw_ee/agent/mcp_servers/design_systems.py` — a `pocketpaw_design_systems` server with `list_design_systems` and `get_design_system`, reading the files directly (no network). Unknown slug returns a soft error listing valid slugs; path traversal is rejected.
- `ee/pocketpaw_ee/extensions.py` + `ee/pyproject.toml` + `mcp_servers/__init__.py` — ambient provider registration, like icons and palette.
- `tests/ee/agent/test_design_systems_mcp_server/` — 14 tests.

## Testing

[...]

PR #1654, opened 2026-07-06
Turn a custom brand color into a full palette scale

DevRohit06:
## What ships

A second tool on the palette server, `scale_from_color`: give it one brand hex and it returns a full 50–900 tint/shade scale (50 lightest, 500 the color you gave, 900 darkest). It complements `extract_palette`, which derives colors from a reference image — this is the path for when the user already knows the exact color.

This is the "custom colors" path directly. A user can say "make the primary `#6B21A8`" and the crew gets real design tokens for it — a full scale with light tints for surfaces and dark shades for text — instead of a single flat value. It's the override primitive behind editing a bundled design system's palette.

## Why now

Answers a concrete gap: the design-system library ships coherent presets, but customization needs a way to swap in an exact color and still get a usable scale. The scale math already existed as an internal helper; this exposes it as a callable tool.

## Stacked on #1653

Branches off `feat/sites-crew-design-systems` because it also touches the palette module in the same stack. Merge the chain base-first with `--rebase`.

## Scope

- `ee/pocketpaw_ee/agent/mcp_servers/palette.py` — the `scale_from_color` tool (input `hex`, optional `role`), `SCALE_FROM_COLOR_TOOL_ID` added to the allowlist tuple. Fail-soft: a malformed hex returns an error, never raises.
- `tests/ee/agent/test_palette_mcp_server/test_mcp_tool.py` — 6 new tests (valid hex → full scale, custom role label, shorthand hex, bad hex and missing hex going soft, tool-id publication); the existing registration test updated for two tool ids.

## Testing

`uv run pytest tests/ee/agent/test_palette_mcp_server/ -q` → 23 passed. `ruff check` + `ruff format` clean.

## Out of scope

No crew wiring — consuming this in the Branding stage to override a design system's palette is later crew work.

### DIFF
--- a/ee/pocketpaw_ee/agent/mcp_servers/palette.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/palette.py
@@ -31,7 +31,15 @@
 # lightest→darkest). Fail-soft: an empty/non-http url, a download/HTTP error, or
 # bytes PIL can't open all return an ``_error_response`` and never raise into the
 # agent.
-"""Agent-side MCP surface for reference-image palette extraction (full scales)."""
+#
+# Updated: 2026-07-06 (feat/sites-crew-scale-from-color, SC-7c) — added a SECOND
+# tool ``scale_from_color`` for the custom-color path: it exposes the pure
+# ``scale_from_base`` helper directly, turning ONE brand hex the user supplied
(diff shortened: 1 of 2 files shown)


DevRohit06:

PR #1656, opened 2026-07-06
Let the site agent actually use the authoring toolbelt

DevRohit06:
## What ships

The `/sites` surface scopes which MCP tools the agent may call to a hard whitelist. Until now that list was only the sites-manager tools (publish, create, edit), so the ambient asset servers — stock photos, icons, palette extraction, and the design-system library — were **silently filtered out**. The agent could not call them on `/sites` even though they were loaded.

This unions those tool ids into the sites allow-list so the authoring flow can actually reach them. It also closes a latent gap: the shipped stock-photo tool, added for the Svelte-track site skill, wasn't reachable on `/sites` for the same reason.

`media` (image/video generation) is deliberately left scoped to `/studio` — site imagery leans on stock first, and generated logo/favicon work belongs to the later Branding stage.

## Why now

This is the load-bearing wiring for the authoring crew. Without it, a crew flow that asks the agent to "retrieve a design system" or "extract a palette" would have those calls quietly dropped — the kind of failure that only shows up during manual testing.

## Stacked on #1654

Branches off `feat/sites-crew-scale-from-color` because it imports the tool-id constants from the icons/palette/design-systems servers. Merge the chain base-first with `--rebase`.

## Scope

- `ee/pocketpaw_ee/cloud/surface/surface_registry.py` — `_load_mcp_tool_ids()` now builds `sites_allow` as the sites-manager tools unioned with stock + icons + palette + design-systems.
- `tests/cloud/surface/test_surface_registry.py` — two tests: every create/refine mode permits the toolbelt, and a simulation of the actual claude_sdk filter predicate proving each toolbelt id survives on `/sites` while an unrelated tool id is correctly filtered out.

## Testing

`uv run pytest tests/cloud/surface/ tests/cloud/test_surface_profile.py -q` → 88 passed. `ruff` clean. Only `sites_allow` changed.

## Out of scope

No preamble or crew-flow changes — this only makes the tools reachable.

### DIFF
--- a/ee/pocketpaw_ee/cloud/surface/surface_registry.py
+++ b/ee/pocketpaw_ee/cloud/surface/surface_registry.py
@@ -195,15 +195,35 @@ def _load_mcp_tool_ids() -> _McpToolIds:
 
     logger = logging.getLogger(__name__)
     try:
+        from pocketpaw_ee.agent.mcp_servers.design_systems import DESIGN_SYSTEM_TOOL_IDS
         from pocketpaw_ee.agent.mcp_servers.foresight import FORESIGHT_TOOL_IDS
+        from pocketpaw_ee.agent.mcp_servers.icons import ICON_TOOL_IDS
         from pocketpaw_ee.agent.mcp_servers.loom import LOOM_TOOL_IDS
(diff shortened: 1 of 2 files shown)


DevRohit06:

PR #1658, opened 2026-07-06
Svelte design-taste skill: bundle it, then give it real design direction

DevRohit06:
## What this is

The svelte-track site crew leans on a `design-taste-svelte` skill for its look. The first commit here bundles that skill; the second gives it the depth it was missing.

## Why the follow-up

The first cut carried the anti-slop guardrails (dials, the CSS "AI tells" list, the static-prerender rules) but dropped the two parts of leonxlnx's taste-skill that actually set a direction. So every site the crew built came out as the same clean house style, just with the obvious tells removed. Cleaner, but same-y.

## What changed

The skill now runs the full framework, in order:

1. **Read the room first.** Before authoring a line, it declares a one-line Design Read: what the business is, who it's for, the vibe.
2. **Pick a direction.** One aesthetic family (clean-tech, soft-premium, editorial-luxury, warm-minimalist, brutalist, dark-tech) that drives the palette, type, materiality, and motion in `app.css`. A variance mandate keeps two similar briefs from landing on the same look.
3. **Set the dials** from the read, then author the sections honouring the direction.

It also picks up the deeper rules from the source: the total em-dash ban, eyebrow restraint, the section-layout-repetition ban, hero discipline, the premium-consumer palette ban, the double-bezel, and a copy self-audit. The pre-flight is now mechanical and checks that the read and the family were actually set, not just the CSS.

`create-svelte-site` now points the author at the Design Read + family pick as the first step, instead of treating the taste skill as an optional add-on.

## Translation note

Everything stays Svelte-5 and static-safe: plain CSS tokens, runes, `@font-face`, CSS-first motion. None of the source's React, Tailwind-utility, `next/font`, or `motion/react` specifics leaked in. Attribution and MIT credit retained.

## Testing

`tests/test_bundled_skills_installer.py` passes for the skills this PR touches (12 passed). The one failing case asserts against `pocketpaw-create-pocket` and fails on an em-dash/mojibake mismatch in the test's own assertion string; it's pre-existing and unrelated to these files.

### DIFF
--- a/ee/pocketpaw_ee/cloud/surface/handlers/sites.py
+++ b/ee/pocketpaw_ee/cloud/surface/handlers/sites.py
@@ -61,6 +61,10 @@
 # and assets. When the flag is OFF, the create branch returns
 # `_create_preamble(meta)` byte-for-byte (no behaviour change). The refine/chat
 # branches (keyed on `pocket_id`) are untouched.
+# Updated: 2026-07-06 (feat/sites-crew-taste, SC-TASTE) — the crew svelte
+# `build_step` now instructs the agent to author the components with the bundled
+# `design-taste-svelte` skill (premium, anti-slop, static-safe styling) on top of
+# the design system's tokens. Svelte-only: the ripple `build_step` is unchanged.
 
 from __future__ import annotations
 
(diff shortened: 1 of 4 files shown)


DevRohit06:
Folded into #1659 (consolidated Sites-crew quality pass). The design-taste-depth commits are part of that PR now.

PR #1659, opened 2026-07-06
Sites crew quality pass: design-taste depth, de-biased palettes, ask-question chips

DevRohit06:
Consolidates three related Sites-crew quality fixes into one PR (previously #1658, #1659, #1661). All three came out of the same thread — the generated sites looked improved but still off — and they stack cleanly, so they're easier to review together.

## 1. design-taste-svelte carries the full framework depth

The first cut of `design-taste-svelte` kept the anti-slop guardrails but dropped the two parts of leonxlnx's taste-skill that set a direction, so every site came out as the same clean house style. This ports them back, translated to the Svelte static-prerender reality:

- **§0 Brief Inference** — read the business / audience / vibe and declare a one-line Design Read before authoring.
- **§2 Design Direction** — pick one aesthetic family (clean-tech / soft-premium / editorial-luxury / warm-minimalist / brutalist / dark-tech) that drives palette, type, materiality, and motion, plus a variance mandate so two similar briefs don't match.
- A more assertive dial baseline, the em-dash ban, eyebrow restraint, the section-layout-repetition ban, hero discipline, the premium-consumer palette ban, and a copy self-audit.

`create-svelte-site` now runs the Design Read + family pick first, instead of treating the taste skill as an optional add-on.

## 2. De-biased design-system palettes (the "Claude colors" fix)

Sites kept landing on the warm terracotta/tan palette. The cause was the design-system library: three of five bundled systems sat in the LLM-default warm zone and a fourth in indigo. Re-tuned four of five (ramps regenerated with the product's own `scale_from_base`, so they stay accessible):

| system | before | after |
|---|---|---|
| warm-local-service | terracotta + sage + Fraunces | honey amber + deep teal on stone, Bricolage Grotesque |
| minimal-editorial | oxblood on warm paper | garnet on a true near-neutral |
| clean-saas | indigo | cobalt (true blue) |
| dark-premium | champagne-gold on warm charcoal | jade-emerald on cool graphite |
| bold-high-contrast | lime | unchanged — already clean |

DESIGN.md, prose, tokens.css, and manifests moved together; every accent contrast-checked; the picker was nudged off defaulting to the warm system.

## 3. Interactive ask_user question chips on the /sites chat

[...]

PR #1661, opened 2026-07-07
Interactive ask_user question chips on the /sites chat

DevRohit06:
## What & why

On the /sites svelte-create flow the agent's clarifying questions come back as plain text, not the interactive "ask a question" chips you get elsewhere. Root cause: that flow runs with `ripple_mode="off"` on purpose (so the agent hand-authors Svelte instead of a Ripple ui-spec), and inline Ripple is what normally renders dynamic chat UI. Turn it off and the agent has no way to render an interactive question.

This restores the ask-a-choice UI **without** re-enabling inline Ripple (which would reintroduce the "builds a Ripple spec instead of a site" problem the off-switch exists to prevent).

## The gap was backend-only

The client is already wired end-to-end for an `ask_user_question` stream frame:

`service.ts` → `chatStore.setAskUser` → `AssistantMessage.svelte` option chips → click → `answerAskUser` sends the choice as the next message.

Nothing on the backend emitted that frame, and the cloud SDK agent can't call the harness `AskUserQuestion` tool — its builtin set is only `Bash/Read/Write/Edit/Glob/Grep/WebSearch/WebFetch` (+`Skill`). So the fix is a tool for the agent to call, plus the emit.

## Changes

- **`ask.py`** — a small in-process MCP tool `ask_user(question, options)` on server `pocketpaw_ask`, mirroring `palette.py`. Near no-op: validates the input and returns an ack that tells the agent to end its turn and wait.
- **`run_core.py`** — a `tool_use` for that tool becomes an `ask_user_question` frame (`{question, options}`) instead of a generic `tool_start`; its internal ack `tool_result` is suppressed. `execute_run` + `chat_router` already forward arbitrary event names, so it reaches the client unchanged.
- **`extensions.py` + `ee/pyproject.toml`** — registered as the `ask` `pocketpaw.mcp_servers` provider.
- **`surface_registry.py`** — added to the /sites allow-list (a hard per-surface whitelist; an id absent there is silently unreachable).
- **`sites.py`** — the crew-flow Phase-1 clarity gate now prefers `ask_user` for the vibe / design-direction choice, then stops and waits.

No client changes.

## Testing

`tests/ee/agent/test_ask_mcp_server/` (16 tests): tool registration, handler happy path + validation fail-soft, dict-option normalization, truncation, and two wiring guards — the `run_core` emit literal must equal `ASK_USER_TOOL_ID` (drift), and `sites_allow` must publish the tool id. Compile + ruff clean; the `test_sites_handler.py` pre-existing failures on this stack are unaffected (count unchanged).

**Not e2e-verified here** — this needs a live cloud run (SDK agent + Mongo + client) to watch the agent call the tool and the chips render + round-trip. Two things to check on deploy: (1) the new entry-point needs a package reinstall (`uv sync --group ee`) to be discovered, same as any `pocketpaw.mcp_servers` addition; (2) confirm the agent reliably calls `ask_user` from the Phase-1 prompt.

Stacked on `feat/sites-crew-create-flow` (which carries the /sites surface + two-phase flow + `ASK_USER_QUESTION` enum this builds on; none of it is on `main` yet).


DevRohit06:
Folded into #1659 (consolidated Sites-crew quality pass). The ask_user chips commit was merged into that branch.

PR #1665, opened 2026-07-08
Universalize the run-start credit gate across agent-run seams

prakashUXtech:
Part 1 of 3 in the billing-enforcement stack (the first-strike from the thesis-gap backlog). Criterion 1: make the run-start billing gate fire on every path that reaches an agent run, not just the sync chat route.

## What changed
- New shared guard `ee/pocketpaw_ee/cloud/credits/guards.py` factors the run-start check out of `run_core`, so every seam blocks an over-budget tenant the same way with no model call when rejected. Three entry points, one per transport: `over_billing_limit` (pure core), `assert_within_billing` (HTTP routes, raises 402), `reject_if_over_billing` (run/stream transport, terminal error frame).
- Wired into the three ungated `AgentPool.run` callers:
  - `agent_bridge` (Slack/WhatsApp/group/DM auto-response) — the gate sits **above** the `_smart_relevance_check` Haiku pre-classifier, so an over-budget workspace spends nothing, not even on the relevance check.
  - `file_versions` (ai-edit / spreadsheet-edit / slides-edit) — raises before the handler's try/except, so it returns a clean 402 instead of a swallowed 500.
  - `planner` — rejects after finding ready tasks, before the claim + run batch; tasks stay `proposed` and run on a later tick once budget frees.
- The worker/executor path (`_reject_if_over_credit_quota`) now delegates to the shared guard and gains the wallet-<=0 check it was missing (before, it only caught the monthly ceiling).

## Flag stays default-off
`billing_enforced` keeps its `default=False`. Enforcement turns on per deployment via `POCKETPAW_BILLING_ENFORCED=true`, which stays the instant kill-switch. OSS/self-host tenants run no ledger and are unaffected. Enterprise plans (uncapped) are a no-op; unknown plans fail closed to Free.

## Tests
- `test_guards.py` covers the shared guard across flag on/off, empty wallet, over-quota, enterprise no-op, and the best-effort side effects.
- `test_agent_bridge_billing.py` proves an over-budget workspace on the group/DM path reaches neither the pre-classifier nor the agent run, and emits one terminal error.
- `test_run_core_credit_quota.py` extended for the new wallet-<=0 worker rejection.
- Both import-linter configs pass, including the OSS->EE boundary the guard relies on. 35 tests green.

Fast-follow: HTTP-route (file_versions) and planner caller-level integration tests. The guard core and the group/DM path are covered here; those two are one-line wirings over the same tested guard.

Stacked: cancel/downgrade (part 2) and SMB caps (part 3) branch off this. Please review, but hold the merge, this is the base of a stack.

### DIFF
--- a/ee/pocketpaw_ee/cloud/chat/runs/run_core.py
+++ b/ee/pocketpaw_ee/cloud/chat/runs/run_core.py
@@ -53,6 +53,13 @@ Changes:
   model call — the no-overspend money guarantee). This is the universal cap that
   covers the worker/executor path; the chat HTTP route ALSO fast-rejects in
   ``agent_router`` so its synchronous caller gets a clean 402 with no DB trace.
+- 2026-07-08 (feat/billing-enforce-gate) — ``_reject_if_over_credit_quota`` now
+  delegates to the shared ``credits.guards.reject_if_over_billing`` so every
+  agent-run seam (this executor, the group/DM bridge, the /files agent ops, the
+  planner) blocks an over-budget tenant identically. The shared helper runs BOTH
+  credit assertions, so the worker/executor leg now ALSO rejects an empty wallet
+  (``check_balance``, 402 credits.insufficient) — previously it caught only the
+  monthly-ceiling case.
 - 2026-06-30 (feat/session-supervisor SS-5) — ``_drive_agent_loop`` now drives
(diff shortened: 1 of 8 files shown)


prakashUXtech:

PR #1666, opened 2026-07-08
Add subscription cancel and stop downgrade double-subscribing

prakashUXtech:
Part 2 of 3 in the billing-enforcement stack. Two money-management bug fixes. **Stacked on #1665** (base is `feat/billing-enforce-gate`, so this diff is only the cancel/downgrade changes). Backend only; the cancel UI is a separate PR.

## (A) Subscription cancel was dead-coded
The provider `cancel_subscription` existed with no callers, no route, no service. Wired end to end:
- `POST /billing/cancel` (workspace-scoped, no body) -> `billing.service.cancel()`.
- `cancel()` loads the workspace's **active** subscription. The `(workspace)` index is non-unique (a workspace keeps historical/cancelled rows), so it filters on `status == "active"` rather than a naive first-match, and returns 402 `billing.no_active_subscription` when there's none.
- The plan revert (`Workspace.plan -> free`) stays reactive on the `subscription.cancelled` webhook. Cancel only adds the initiation side, no duplicate revert.

## (B) Downgrade/switch no longer double-subscribes
`subscribe()` called `create_subscription` unconditionally, so switching plans opened a second parallel Dodo subscription and double-billed. Now, when a workspace already has an active subscription, it runs a guarded cancel-then-create: open the new checkout first (a create failure leaves the current sub untouched), then cancel the old one at the gateway. `event_id` webhook idempotency is unchanged.

## One decision for you
The cancel-old-immediately ordering has a residual gap: if the buyer abandons the new checkout after the old sub is cancelled, they're temporarily reverted to free. It's the lesser evil versus the double-billing bug and is inherent to the checkout-session model. Dodo's SDK exposes an atomic `change_plan` (proration, no re-checkout, no drop-to-free window) that closes the gap cleanly, but wiring it needs a new provider-port method, a `subscription.plan_changed` webhook handler, and a `/subscribe` response-contract change, which is out of scope for a bug fix and tracked as a follow-up in the code. Your call on whether to schedule it.

## Tests
- `test_subscriptions.py` (+11, 24 total): cancel targets the active sub id, 402s on none / only-stale-rows, ignores cancelled rows; switch cancels the old sub and opens exactly one new checkout; a fresh subscribe cancels nothing; the switch stays webhook-idempotent. Tests spy at the SDK-client edge (real DodoProvider, mocked client).
- 147 billing+credits tests green (C1 included, no regression). Both import-linter configs pass.

Hold the merge, this is the stacked base of the caps PR (part 3).

### DIFF
--- a/ee/pocketpaw_ee/cloud/_core/errors.py
+++ b/ee/pocketpaw_ee/cloud/_core/errors.py
@@ -17,6 +17,13 @@ rather than a balance refill; it carries the effective `ceiling` and the
 FL-2 (file-version spine, port of dewani12's #1193) added
 ``PreconditionFailed`` (412) for stale ``If-Match`` optimistic-concurrency
 failures on the file-version write path.
+
+Changed 2026-07-08 (feat/billing-cancel-downgrade): added
+``NoActiveSubscription`` (402, ``billing.no_active_subscription``) — raised by
+the subscription-cancel path when a workspace has no ``active`` subscription to
+cancel (only historical / already-cancelled rows, or never subscribed). 402 (the
+same money-error family as ``InsufficientCredits`` / ``QuotaExceeded``) so the
+client renders a "not subscribed" state rather than a 404-style missing resource.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1668, opened 2026-07-08
Add SMB seat/pocket/connector caps to the plan ladder

prakashUXtech:
Part 3 of 3 in the billing-enforcement stack, and the last backend criterion: real SMB resource caps so a plan can actually be sold. **Stacked on #1666** (base `feat/billing-cancel-downgrade`), so this diff is only the caps changes. Backend only; the PlanCards UI is a separate PR.

## What changed
Mirrors the existing `monthly_ceiling` precedent exactly:
- `plans.py` — `max_seats` / `max_pockets` / `max_connectors` on every `PlanTier`, from tunable dicts with the same fail-closed default for unknown keys.
- `entitlements` — the three caps surface on `Entitlements` and thread through `resolve_entitlements`, with defensive Free defaults.
- `_core/errors.py` — `PocketLimitError` (402 `billing.pocket_limit`) and `ConnectorLimitError` (402 `billing.connector_limit`), mirroring `SeatLimitError`.
- Enforcement, at create/invite/enable time only (never retroactive):
  - **Seats** — the three seat gates source the limit from the plan via `_effective_seat_limit = max(doc.seats, plan.max_seats)`, so an upgrade lifts the cap and a workspace with a higher custom seat count never regresses. `raise_seats_for_plan` on the `subscription.active` webhook resyncs the stored count up on an upgrade (upgrade-only, never strips).
  - **Pockets** — `PocketLimitError` at the ceiling in the create path; the ripple-spec agent path degrades to `None` at the cap (its failure contract) rather than throwing a raw 402 through the agent loop.
  - **Connectors** — `ConnectorLimitError` at the ceiling in `enable_connector`, but only when turning a connector on; an idempotent re-enable is never blocked.

## Two design choices
- Pocket + connector caps are **gated behind `billing_enforced`** (same as the credit gate), so OSS/self-host tenants are unaffected and pay no extra DB read. Seat enforcement stays always-on but is now plan-sourced.
- The cap **numbers are generous placeholders pending your pricing call** — the machinery is tier-agnostic and works with whatever ladder you land on. Free `max_seats` is pinned at 5 (the current `Workspace.seats` default), so no existing workspace regresses.

| tier | seats | pockets | connectors |
|---|---|---|---|
| free | 5 | 200 | 50 |
| go | 10 | 1,000 | 100 |
| pro | 25 | 5,000 | 250 |
| pro_max | 100 | 20,000 | 1,000 |
| enterprise | uncapped | uncapped | uncapped |

## Tests
- New `test_pocket_cap.py`, `test_connector_cap.py`, plus caps in `test_plans.py` / `test_entitlements.py` and seat tests in `test_service_v2.py`. 143 targeted pass: caps on every tier + fail-closed unknown key; seat limit sources from plan and an upgrade lifts it; pocket/connector caps raise at the ceiling, pass below, and no-op when billing is off.
- 152 billing+credits tests green across the full C1+C2+C3 stack. Both import-linter configs pass.
- Two `test_delete_preview_*` failures in `test_service_v2.py` are pre-existing: they fail identically on the base branch (verified by running them on the pre-C3 tip) and don't touch billing/cap code.

Top of the stack. Please review, but hold the merge — the whole stack lands together (merge the base with `--rebase`).

### DIFF
--- a/ee/pocketpaw_ee/cloud/_core/errors.py
+++ b/ee/pocketpaw_ee/cloud/_core/errors.py
@@ -24,6 +24,13 @@ the subscription-cancel path when a workspace has no ``active`` subscription to
 cancel (only historical / already-cancelled rows, or never subscribed). 402 (the
 same money-error family as ``InsufficientCredits`` / ``QuotaExceeded``) so the
 client renders a "not subscribed" state rather than a 404-style missing resource.
+
+Changed 2026-07-08 (feat/billing-smb-caps): added ``PocketLimitError`` (402,
+``billing.pocket_limit``) and ``ConnectorLimitError`` (402,
+``billing.connector_limit``) — the pocket-create and connector-enable siblings of
+``SeatLimitError``. Same 402 money-adjacent family; distinct codes so the UI can
+prompt a plan upgrade. Both enforce at create/enable time only (never retroactive)
+and only when ``billing_enforced`` is on.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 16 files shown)


prakashUXtech:

PR #1669, opened 2026-07-08
Fan cloud notifications out to Slack and generic webhooks

prakashUXtech:
The first slice of external alerting (Wave 2), and the highest-value one: alerts now leave the app. Cloud notifications have been WebSocket-only (in-app); this adds a per-workspace fan-out to Slack and a generic webhook, on top of the existing in-app push.

## What changed
- New `NotificationDeliveryConfig` (per-workspace, unique on `workspace_id`, in `cloud/models/`): Slack incoming-webhook URL, a generic webhook URL, an `enabled` master switch, and per-kind `routes`.
- `notifications/delivery.py` -> `_deliver_external`, awaited right after the `emit(NotificationNew(...))` in `service.create`. Every cloud notification funnels through `create` (tasks, meetings, mentions), so this covers them all.
- It's **fire-and-forget / never-raise**, modeled on `emit`: a dead, slow, or malicious sink can't roll back the notification insert or bubble an exception out of `create`. Each POST is bounded by a 5s timeout and wrapped per-sink.
- `PUT` / `GET /notifications/delivery-config` to manage the config, gated by a new `notifications.manage` admin action.

## Security: SSRF guard on the webhook URLs
Workspace-admin-supplied URLs are an SSRF vector, so `is_safe_webhook_url` requires https, rejects literal private/loopback/link-local IPs and known metadata hostnames, and runs both at write time (a bad URL is rejected up front) and at delivery time (defense in depth). One known gap, flagged for follow-up: a hostname that *resolves* to a private IP isn't caught here (no DNS lookup in the hot path); full DNS-resolution hardening, like `audit.webhooks`, is the next step.

## Deliberately deferred (with reasons)
- **The OSS `_notify` / `_emit_alert` paths carry a FOLLOW-UP note only.** OSS core can't import the EE delivery module (import-linter forbids OSS->EE), so that fan-out needs a separate mechanism (event-bus indirection or an OSS-side impl), not this module. Forcing the import would break the build.
- **Email sink** (via the Gmail connector) is structured as an easy third sink, deferred so this PR stays tight.
- **Criteria 2 (default-on automation flips) and 3 (merged Automations screen)** are their own slices.
- **UI** (a form to paste the Slack webhook URL) is a small follow-up PR.

## Tests
- 19 new tests in `test_delivery.py` drive the real `notifications.service.create` with a stubbed httpx transport (spy on the POST, real client code runs): both sinks fire with the right payloads; a sink that raises does not propagate (the insert and emit still happen); no-config / disabled / wrong-workspace deliver nothing; per-kind routing narrows; upsert keeps one row per workspace; an unsafe URL is rejected and stores nothing.
- Both import-linter configs pass, including `OSS core may not import from EE` and the extended Notifications Beanie-writes contract. `ruff` clean.
- 7 failures in the broader notifications suite (`test_router_golden`, `test_dto`, `test_derived`) are pre-existing: they fail identically on `origin/dev` (verified), unrelated to this change.

Please review; hold the merge.

### DIFF
--- a/ee/pocketpaw_ee/cloud/models/__init__.py
+++ b/ee/pocketpaw_ee/cloud/models/__init__.py
@@ -1,5 +1,12 @@
 """Cloud document models — re-exports for Beanie init.
 
+Updated: 2026-07-08 (feat/external-alerting-delivery) — added
+``NotificationDeliveryConfig`` (the per-workspace external-delivery config: Slack
+incoming-webhook + generic HTTPS webhook + enabled switch + per-kind routing) to
+the imports, ``__all__``, and ``get_all_documents()`` so the
+``notification_delivery_configs`` collection is wired into ``init_beanie``. Only
+``ee.cloud.notifications`` (service writes, delivery reads) imports the doc class
+directly (import-linter "Notifications" contract).
 Updated: 2026-07-03 (feat/files-share-links FL-12b) — added ``ShareLink`` (the
 public, token-gated file share-link doc) to the lazy uploads loader,
 ``get_all_documents()`` and ``__all__`` so the ``file_share_links`` collection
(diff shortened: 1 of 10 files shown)


prakashUXtech:

PR #1673, opened 2026-07-09
Governed-rules router + per-workspace enforcement toggle

prakashUXtech:
The API and per-tenant switch behind the guardrail authoring UI (instinct-guardrail-ux C1 backend). The enforcement engine already ships and is tested — a UI-authored active rule already flows through `resolve_instinct` — so this is the thin HTTP surface over `rules.service` plus a per-workspace enforcement flip, with no new gate logic.

## What changed
- New `rules/router.py` — thin, workspace-scoped, admin-gated (`rules.manage`) endpoints over the shipped service:
  - `POST /rules` create · `GET /rules` list-active · `POST /rules/{id}/archive`
  - `GET` / `PUT /rules/enforcement` — the per-workspace enforcement toggle (declared before `/{id}` so `/enforcement` is not captured as a rule id). Mounted in `cloud/__init__.py`.
- Per-workspace enforcement: `instinct_enforce_discovered_rules` was a single global flag. A new `InstinctWorkspaceConfig` override lets an admin turn authored-rule enforcement on for one tenant without a code change. `instinct_dispatch` now resolves via `_enforcement_enabled(workspace_id)` — the tri-state override (true/false) wins, else the global flag.
- The override read is **fail-open**: a config/DB read error yields no enforcement, logs, and never blocks the gate — the same rail the existing rule-load uses. Enforcement can only ever add a block, so "no enforcement" is the safe direction when the override is unreadable.

## Deliberately deferred
- The frontend `GuardrailsPanel` (object/property/operator → CEL authoring) is the next slice.
- The 3-step LLM composer (Criterion 2) is a later, separate slice.

## Tests
- `test_rules_router.py` + `test_workspace_enforcement.py` (16 new): create/list/archive with tenant isolation + auth; the per-workspace override enforces when the global flag is off, falls back to the global default, and fails open on a read error. 30 pass across these plus the existing discovered-rule enforcement test (no regression). Both import-linter configs green (including OSS-cannot-import-EE); ruff clean.

Please review; hold the merge.

### DIFF
--- a/ee/pocketpaw_ee/cloud/__init__.py
+++ b/ee/pocketpaw_ee/cloud/__init__.py
@@ -249,6 +249,7 @@ def mount_cloud(app: FastAPI) -> None:
     from pocketpaw_ee.cloud.pockets.router import router as pockets_router
     from pocketpaw_ee.cloud.projects.router import router as projects_router
     from pocketpaw_ee.cloud.request_log.router import router as request_log_router
+    from pocketpaw_ee.cloud.rules.router import router as rules_router
     from pocketpaw_ee.cloud.sessions.router import router as sessions_router
     from pocketpaw_ee.cloud.skills.router import router as skills_router
     from pocketpaw_ee.cloud.workspace.router import router as workspace_router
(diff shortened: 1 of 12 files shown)


prakashUXtech:

prakashUXtech:
Folded into the thesis-gap integration branch, which merged to dev in #1672. Closing in favor of the combined land.

PR #1691, opened 2026-07-10
feat(fabric): trust-ladder resolver — pure survivorship over statements (FST-2)

prakashUXtech:
## What this does

Builds on #1690. The resolver — the brain of the source-truth feature — as a pure, deterministic function, exhaustively tested before it touches any write path. `resolve(statements, rules)` picks the winning value for a property: deprecated filtered out, pinned short-circuits, then the writer-class trust ladder (human > connector > mirror > agent > inferred, per-(object_type, property) overridable), then within the winning tier: preferred rank > normal, open validity > closed, recency on `observed_at`, statement-id tiebreak. Returns winner + ordered losers + `is_disputed` + `unresolvable` (the same-tier/same-rank/both-fresh/materially-different case the stewardship slice consumes later).

Nothing in production calls it yet — pure functions plus 37 unit tests, including a shuffle-determinism check across every ordering leg.

## Design notes

- Rank never crosses tiers: a low-trust writer marking its own statement preferred can never beat a higher-tier normal statement.
- Closed-validity statements stay candidates (reads never block, even on all-closed history) but are excluded from dispute checks — so a CHANGE (close old + append preferred) resolves cleanly instead of generating dispute noise. Simulated in tests ahead of the FST-5 verbs.
- Unresolvable requires same tier, same rank, both open, within a configurable recency epsilon (default 24h), materially different — a rank or validity difference is a ranking by definition.
- Trust rules are plain readable data (`TrustRules` + per-property overrides + `default_trust_rules()`), so an agent can explain why a value won.

## Verification

37 resolver tests; full fabric selection 242 passed (205 baseline + 37, zero regressions); ruff clean; import-linter contract kept.

Stacked on #1690; next slice wires this into the first merge site in shadow mode.

### DIFF
--- a/src/pocketpaw/fabric/resolver.py
+++ b/src/pocketpaw/fabric/resolver.py
@@ -1,0 +1,251 @@
+# fabric/resolver.py — Trust-ladder resolution over Statements (FST-2).
+# Created: 2026-07-10 (feat/fst-2-resolver) — the brain of the source-truth
+#   chain as a PURE function.
+# Updated: 2026-07-10 (FST-2 spec review) — rank + validity joined the
+#   within-tier ordering so FST-5's CHANGE verb (close the old statement's
+#   valid_to + append a new PREFERRED statement) is visible to resolution;
+#   is_disputed and unresolvable tightened accordingly (closed-validity
+#   losers are resolved history, not live disputes).
+#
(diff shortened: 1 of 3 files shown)


prakashUXtech:

PR #1694, opened 2026-07-10
feat(fabric): shadow statements at the projection + cloud-mirror merge sites (FST-4)

prakashUXtech:
## What this does

Builds on #1692. Completes shadow coverage: all three Fabric merge sites now emit statements + divergence telemetry in shadow mode. After this slice, every write path — connector ingest, journal-replay projection, and the Firestore mirror — is observed; the mixed-trust cohabitation (discovery-inferred next to connector facts) is fully visible before anything enforces.

## The two hard problems, solved

- **Journal projection (site 2)** turned out to be a pure in-memory fold, not a store write — so it stages shadow observations (pre-merge snapshot + patch + event identity) and flushes them asynchronously through the same store machinery FST-3 built (no duplicated logic). Replay idempotence rides a new `fabric_shadow_events` claim table: the event id is claimed with INSERT OR IGNORE *before* the statement pass, making shadow recording at-most-once per journal event — replaying the same journal twice cannot double-append. A store-less projection (the default) never even reads the mode flag.
- **Mirror provenance (site 3)**: the ee worker now writes `writer_class="mirror"` with the exact Firestore doc path as the source identity, and threads real source timestamps (cursor field, else snapshot update_time, else honest ingest-time — never fabricated). The second-source promotion gate compares writer *families* (connector+mirror = one machine-sync family) so a mirror self-refresh can't force-promote every property; the ladder still ranks connector above mirror at resolve time.

## Create-path decision (documented, test-proven)

Creates get no statement hook: promotion-time seeding already preserves the create-time claim. Proven by create-then-update tests on both the OSS and mirror paths — both claims present, conflict visible.

## Verification

267 fabric tests (256 baseline + 11), 21 cloud fabric_ingest tests (incl. 5 new), replay-matches-live and double-replay-dedupe tests, FST-3 guard suite untouched and green, adjacent consumer sweep 43 passed. Ruff + import-linter clean.

Stacked on #1692. Next slice: enforce mode — the resolver takes ownership of the cache, plus the CHANGE/CORRECT verbs and the discovery-vs-connector proof.

### DIFF
--- a/ee/pocketpaw_ee/cloud/fabric_ingest/service.py
+++ b/ee/pocketpaw_ee/cloud/fabric_ingest/service.py
@@ -11,6 +11,17 @@
 #   pinned via the mapping's new optional type_id). Behavior note inherited
 #   from the canonical loop: a mapped field absent from a document projects as
 #   None (the mirror reflects the source) rather than being skipped.
+# Updated: 2026-07-10 (FST-4 — SHADOW mode at merge site 3) — the mirror now
+#   threads its TRUE provenance through ingest_records' new keyword params:
+#   writer_class="mirror" (a Firestore mirror is not a primary connector —
+#   the trust ladder ranks connector > mirror), source_document_uri = the
+#   full Firestore doc path (per-record, via the existing _DOC_PATH_KEY
+#   sentinel), and observed_at = the best available SOURCE time per doc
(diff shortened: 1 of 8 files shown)


prakashUXtech:

PR #1696, opened 2026-07-10
feat(fabric): un-rankable conflicts to the Instinct stewardship queue + PIN/IGNORE (FST-6)

prakashUXtech:
## What this does

Builds on #1695. The conflict lifecycle: conflicts the trust policy can rank keep auto-resolving silently; the un-rankable ones (same tier, same rank, both fresh, materially different) become Instinct proposals a human arbitrates in the Tray — the same agents-propose-humans-decide gate the rest of the OS uses, now arbitrating facts.

- OSS steward verbs as siblings of change/correct: `pin_statement` (durable "this one wins" — the resolver short-circuits on pins, so future rival observations stay settled), `unpin_statement`, `ignore_statement` (deprecate a bogus claim). Mode-respecting cache updates; every action returns the new Resolution.
- OSS conflict surface: `detect_open_conflicts` recomputes from statements (no separate conflicts table — the statements are the state), keyed `(workspace, object, property)`.
- EE gate glue (`cloud/fabric_conflicts/`, mirroring the rule-proposals precedent): a post-ingest sweep stages one proposal per conflicted property with the competing values + full provenance as read-only choices; approve executes PIN on the chosen statement (choice validated against the staged set — an edit can't smuggle an arbitrary id); reject closes with zero statement changes. Volume guard warns past 5 open proposals per workspace.
- Reject-memory: a dismissed conflict is not re-filed by the next sweep unless a new rival observation changes the conflict signature — without this, the 5-minute sweep would re-nag every human decision forever.

## Wiring

The sweep rides the existing fabric-ingest scheduler tick plus fires after ingest batches (one exception-shielded tail hook — conflicts are born at merge sites, so after-ingest is the natural beat). Mode-gated: off = zero reads. Conflict proposals deliberately do not participate in discovery's supersede sweep — they aren't discovery artifacts and survive re-runs.

## Verification

34 new tests (11 verb, 9 conflict-surface, 14 gate e2e incl. dedupe, reject-memory, approve→PIN→cache-in-enforce, tenancy). Full fabric selection 308; the live-gate regression sweep: 383 cloud instinct + 131 OSS instinct tests green; shadow/enforce guard suites byte-for-byte untouched; ruff + import-linter clean.

Stacked on #1695. Next: freshness (the stale-authority demotion the enforce slice deliberately deferred).

### DIFF
--- a/ee/pocketpaw_ee/cloud/fabric_conflicts/__init__.py
+++ b/ee/pocketpaw_ee/cloud/fabric_conflicts/__init__.py
@@ -1,0 +1,40 @@
+# ee/cloud/fabric_conflicts/__init__.py — the conflict-stewardship Instinct
+# proposal type (a peer of ``_instinct_rule`` / ``_fabric_objects`` /
+# ``_pocket_create`` / ``_pocket_write`` / ``_code_change`` /
+# ``_external_action`` / ``_belt_plan`` / ``_artifact_change`` /
+# ``_admin_action``).
+# Created: 2026-07-10 (FST-6 — the conflict lifecycle) — the hybrid conflict UX:
+#   the source-truth resolver auto-resolves every conflict it can RANK; the
+#   un-rankable残り (``Resolution.unresolvable=True``) is swept into ONE Instinct
+#   proposal per conflicted property for a human steward to arbitrate. Approve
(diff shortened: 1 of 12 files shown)


prakashUXtech:

PR #1697, opened 2026-07-11
feat(fabric): freshness tri-state, within-family staleness demotion, provenance on reads (FST-7)

prakashUXtech:
## What this does

Builds on #1696. Facts now age visibly, and trust becomes readable:

- **TTL classes as data** (volatile 1d / default 30d / stable 365d, per-property overridable) + a pure `freshness()` tri-state: fresh / aging / stale. Datetime handling normalized to UTC at the comparison boundary (the naive-local vs naive-UTC wrinkle carried since FST-3, closed).
- **Within-family staleness demotion** in the resolver: when a tier's best candidate is stale, a fresh value from the SAME writer family (connector+mirror = one machine-sync family) takes the win — closing the case FST-5 flagged, where a stale connector seed beat a fresher mirror value forever. Demotion never crosses families: a stale connector fact still beats a fresh agent/inferred claim. `now=None` callers skip freshness entirely (legacy behavior byte-identical); a freshness-demoted loser is resolved history, not dispute noise.
- **Provenance on reads** (opt-in): `store.get_object_provenance(object_id)` reports, per statement-tracked property, the dispute/unresolvable flags, freshness, statement count, and the winning value's writer + source summary. The `fabric_query` MCP tool gains `include_provenance` (default false — response shape unchanged without it), so an agent asked "where did this figure come from / can it be trusted" can answer.
- The divergence line gains a trailing ` freshness=<tri-state>` token — additive, field order untouched; the three format-guard assertions were updated with it (the deliberate format evolution those guards exist to gate).

## Build note

The implementing agent stalled twice on infra (stream watchdog) after completing the trust/resolver core with 30 passing tests; the controller finished the read surface, MCP wiring, and format-guard updates by hand and reviewed the combined diff.

## Verification

341 fabric tests (308 baseline + 33: TTL/tri-state units incl. UTC-mixing, the named stale-seed-vs-fresh-mirror case, never-across-family, now=None back-compat, pin-ignores-freshness, read-surface e2e); cloud fabric dirs 37 green; prior-suite diff confined to the three format-guard assertions; ruff + format + import-linter clean.

Stacked on #1696. Final slice next: the shadow-divergence harness + docs.

### DIFF
--- a/ee/pocketpaw_ee/agent/mcp_servers/fabric.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/fabric.py
@@ -1,6 +1,12 @@
 # fabric.py — in-process MCP server exposing read-only Fabric ontology access
 # to the claude_agent_sdk cloud chat backend. Created: 2026-06-11
 # (feat/fabric-instinct-mcp-providers).
+# Updated: 2026-07-11 (FST-7 — provenance on reads) — fabric_query gained an
+#   opt-in ``include_provenance`` arg (default false; response shape unchanged
+#   without it): attaches per-property trust detail for statement-tracked
+#   properties via store.get_object_provenance — disputed/unresolvable flags,
+#   freshness tri-state, and the winning value's writer + source. Best-effort
+#   per object; a provenance failure never breaks the query.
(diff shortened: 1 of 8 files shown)


prakashUXtech:

PR #1704, opened 2026-07-12
Add create_html_site tool for the html track (HE-6)

DevRohit06:
## What this does

Adds `create_html_site` as the fourth create tool on the shared
`pocketpaw_sites_manager` MCP server, alongside `create_landing_site`
(ripple), `create_svelte_site`, and `create_dynamic_site`. It lets an
agent author a raw HTML/CSS/JS site and hand it straight to `publish`.

The tool persists a `{ relative_path: file_contents }` source map
verbatim — `engine="html"`, `ripple_spec=None`, `trusted=True` — and
stamps the pocket `type="site"`, `pattern="landing"`. That is the same
author-owns-the-source shape `create_svelte_site` uses, minus the
SvelteKit scaffold and the bundler. Validation is intentionally light:
the only required key is `index.html` (the edge serves it at the root),
and every value must be a content string, since an html site carries no
live-data binding siblings.

## Opt-in, not a default flip

The default marketing path is unchanged. The tool description steers the
agent here only on an explicit "plain HTML / no framework / just an
`index.html`" request, and `pocketpaw-create-paw-site` gains a scoped
opt-in section that points at it while keeping `create_landing_site`
(ripple) as the default. Skill count stays at three.

Making html the *default* — rewriting the marketing brain to author html
out of the box — is a separate change. It has to ship the refine-mode fix
in the same PR (otherwise refining an html site opens ripple mode against
a pocket with no rippleSpec) and it is gated behind the taste work, so it
is deliberately left out here.

## Scope note

This is the create half of the slice. The full create → publish →
live-URL demo also needs the assets-only Worker deploy path, which lands
separately. Create works now; the build-skip for html publishes already
merged.

## Tests

- `test_create_html_site.py` — source-map validation (`index.html`
  required, non-string values rejected), registration on the shared
  allowlist, and an end-to-end persist that reads the pocket back from
  Mongo to confirm `engine="html"`, the full source map, and no
  rippleSpec.
- `test_mcp_tool.py` — pinned tool count bumped to six with the new id
  asserted on the server allowlist.

`tests/ee/agent/test_sites_mcp_server/` is green (67 pass). Lint and
format are clean on the changed files; the two modules carry no new mypy
errors (the pre-existing `_validate_dynamic_spec` union-attr notes are
unchanged). One unrelated installer test fails on this box from a Windows
cp1252 encoding quirk that also fails on `dev`.

### DIFF
--- a/ee/pocketpaw_ee/agent/mcp_servers/sites.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/sites.py
@@ -22,6 +22,13 @@
 # ``create_dynamic_site`` tool also registers on this SAME server. Dynamic sites
 # are ripple-engine sites whose spec carries live-data bindings; publish carries
 # those through to the paw-sites generator.
+#
+# Updated 2026-07-12 (feat/sites-html-create-tool, HE-6): the ``create_html_site``
+# tool also registers on this SAME server. An html site is a raw {path: contents}
+# HTML/CSS/JS map with no framework; publish materializes it and skips the Node
+# build. Its id rides ``SITES_TOOL_IDS`` so the per-surface allowlist picks it up
+# automatically. Opt-in — the default marketing brain stays create_landing_site
+# (ripple); the default flip to html is HE-12.
 """Agent-side MCP surface for publishing a PocketPaw pocket as a Paw Site.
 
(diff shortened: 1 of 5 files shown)


DevRohit06:

PR #1705, opened 2026-07-12
Deploy an html site as an assets-only Worker (HE-4)

DevRohit06:
## What this does

Makes the workers-mode deploy engine-aware so a form-less html brochure
goes live with no server script at all — the last piece the html
create → publish → live-URL demo was waiting on.

An html site's generator output is a raw static tree at the project root
(`static_output_rel("html")` is `"."`), with no `_worker.js`. Its
`wrangler.jsonc` now drops `main` and `nodejs_compat` and just serves
`assets.directory: "."` — a legal no-`main` Worker that ships zero bytes
of JavaScript. ripple and svelte keep the exact SvelteKit-worker config
and the `_worker.js` `.assetsignore`. The engine parameter defaults to
`ripple`, so every existing caller is byte-for-byte unchanged; the
static publish path threads the pocket's engine into `deploy_workers`.

## One deviation from the task, on purpose

The task sketch said to drop the `.assetsignore` for html "since there
is no `_worker.js` to ignore." That misses a real leak: with
`assets.directory` set to the project root, `wrangler.jsonc` lives inside
the asset directory and wrangler would serve it as a public asset —
wrangler only auto-excludes its own config when the asset directory is a
subdirectory ([Cloudflare static-assets docs](https://developers.cloudflare.com/workers/static-assets/)).
So html writes a minimal `.assetsignore` that excludes the config files
instead. Same intent (nothing but real assets gets served), different
files.

## Depends on

Built on #1685 (dynamic workers deploy), which rewrote this same file and
merged to `dev` first along with its base #1683.

## Tests

- html emits an assets-only config: no `main`, no compat flags, no D1,
  `assets.directory` of `.`
- its `.assetsignore` excludes `wrangler.jsonc` and never lists a
  `_worker.js`
- html still passes `--config wrangler.jsonc`
- a regression guard pins the svelte/ripple config and its `.assetsignore`
  unchanged

Full `ee/sites` suite green. The four failing subprocess-timeout tests
are pre-existing and reproduce on a clean `dev` on this machine
(Windows process-group timing), unrelated to this change. Lint, format,
and mypy clean on the touched files.

### DIFF
--- a/ee/pocketpaw_ee/sites/service.py
+++ b/ee/pocketpaw_ee/sites/service.py
@@ -1323,7 +1323,9 @@ async def _deploy_site_doc(
     cloudflare: Any | None = None,
     bundle_reader: Callable[[str], bytes] = _default_bundle_reader,
     local_deploy: Callable[[str, str], str] | None = None,
-    workers_deploy: Callable[[str, str], Any] | None = None,
+    # HE-4: the workers deployer is engine-aware (``deploy_workers(site_id,
+    # project_dir, *, engine=..., d1_database_id=...)``), so the seam takes kwargs.
+    workers_deploy: Callable[..., Any] | None = None,
 ) -> _SiteDoc:
     """Generate, smoke-gate, deploy, and UPSERT the LIVE canonical Site doc.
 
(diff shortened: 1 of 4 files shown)


DevRohit06:

PR #1706, opened 2026-07-12
Remove the bundled ripple-recipes KB scope

DevRohit06:
## Why

The shipped `ripple-recipes` kb-go scope was a set of hand-authored
ripple pattern recipes — sales-pipeline dashboard, customer-support app,
landing page, how-to viewer — mirrored to `~/.knowledge-base/` on boot
and fed to the create specialist through a "recipe preflight" step that
anchored every new pocket on one of a handful of fixed compositions.

That anchoring is the design bias worth shedding. Canned recipes make
generated pockets converge on the same few layouts; design breadth
should come from live references, not a frozen scope baked into the
wheel. So this removes the scope rather than growing it.

## What's removed

- `src/pocketpaw/bundled_kb/` — the installer and the compiled scope
  (index + wiki + cache), plus its test
- the `auto_install_bundled_kb_scopes` setting and the boot-time install
  call in `dashboard_lifecycle`
- the recipe-preflight / recipe-library prompt blocks in the create
  specialist (`ripple/_pockets.py`). The MCP-variant steps renumber to
  STEP 0-3 (template preflight → understand the brief → pick structure →
  delegate)

## What's untouched

- The general `_get_kb_context` injection (workspace / agent / pocket KB
  scopes) — only the bundled ripple-recipes scope is gone
- Bundled templates and bundled skills — their docstrings just drop the
  stale "sibling to bundled_kb" references
- Built-in pocket templates still work; the template preflight is the
  remaining, higher-quality anchor

## Verification

Modules import, the setting is gone, and `pocketpaw.bundled_kb` is no
longer importable (removed a leftover `__pycache__` that otherwise kept
the dir alive as a namespace package). OSS-core suite is green — 407
passed; the 3 failures reproduce on a clean tree on this machine
(a Windows encoding quirk and two config env-leak tests), unrelated to
this change. Lint and format clean.


DevRohit06:

PR #1708, opened 2026-07-13
Stop the Svelte skill hijacking plain landing requests

DevRohit06:
## The bug

A plain "build a modern dentist landing page with a booking form,
services, and contact info" launched `pocketpaw-create-svelte-site` — the
hand-authored Svelte track — instead of the default ripple landing brain.

## Why

The svelte and ripple create skills carried the **same trigger examples**
in their descriptions: "build a dentist landing site", "a marketing page
for my bakery", "a landing page for my SaaS". The svelte one added a prose
caveat ("only when the Svelte engine is on"), but the agent's
natural-language skill selection matched on the shared examples and fired
svelte anyway.

The "Use Svelte pages" toggle is only authoritative through the surface
profile — when it's on, the profile pins the skill set to svelte. When
it's **off**, the default surface doesn't exclude the svelte skill, so the
description alone can pull the agent to it. That happens on any surface,
and especially in general chat where there's no sites-profile gating at
all.

## The fix

Rewrite the two descriptions so the split is unmistakable:

- **create-svelte-site** — use ONLY when `engine="svelte"` is explicitly
  selected. A plain landing request is now stated as the *wrong* signal,
  with a direct pointer to `create-paw-site`. The old shared examples are
  reframed as "if you see these and the Svelte engine is off, this is the
  wrong skill".
- **create-paw-site** — stated as the DEFAULT landing brain: use for any
  brand-new landing site UNLESS the Svelte engine is on.

Description-only — no code or routing tables touched. It's the portable
lever because it corrects natural-language selection on every surface,
not just `/sites`.

## Follow-up (not here)

The authoritative belt-and-suspenders — having the default `/sites`
surface profile expose every sites-create skill *except* svelte (without
breaking dynamic/landing/html routing) — belongs with the sites-profile
rework, since that restructures the profile anyway.

## Verification

YAML frontmatter parses for both skills; surface/routing tests green (45);
no test pins the changed prose.

### DIFF
--- a/src/pocketpaw/bundled_skills/_bundled/skills/pocketpaw-create-paw-site/SKILL.md
+++ b/src/pocketpaw/bundled_skills/_bundled/skills/pocketpaw-create-paw-site/SKILL.md
@@ -4,11 +4,13 @@ description: |
   Build a marketing landing page as a Paw Site — a real, standalone
   website composed by conversion role (navbar, hero, services, social
   proof, pricing, CTA, lead form, footer) and rendered statically to the
-  edge. Invoke when the user describes a BRAND-NEW marketing / landing
-  site: "build a dentist landing site", "make a marketing page for my
-  bakery", "a landing page for my SaaS", or when the create-site flow
-  routes a new-site request here. This is the marketing-first authoring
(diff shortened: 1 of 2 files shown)


DevRohit06:

PR #1714, opened 2026-07-14
Paw Bar standalone concierge: public chat seam + grounding guard

prakashUXtech:
## What this is

The backend seam for the standalone Paw Bar concierge: a public, anonymous chat endpoint that any embedded widget can call, running the site's own scoped agent (its soul and KB) behind a governed tool lockdown. This is Tier 1 (text) of the standalone-concierge plan. Live voice and the widget UI are separate follow-ups.

Design + build spec: `docs/design/drafts/2026-07-14-paw-bar-standalone-concierge.md` and the `-tier1-build.md` companion (in paw-workspace).

## How it works

A visitor's widget posts to `POST /paw-bar/chat` with the public embed key, an anonymous `customer_ref`, and a message. The handler runs the cheap gates first (widget exists, origin allowlist, rate limit, injection screen), authenticates the embed key, binds the widget to the key's workspace and pocket, then dispatches a scoped run over the same machinery the authenticated dashboard chat already uses.

The build reuses what ships instead of adding parallel primitives:

- The public embed key is the existing `Site.signed_key` + `allowed_origins`, the same credential the leads-capture path already resolves. `resolve_site_key` lifts that inline check into one reusable resolver. It is deliberately not a fork of the argon2 `api_keys` path, which is the wrong model for a world-visible key.
- The run rides `create_run` + `executor.submit` + `execute_run` + `transport` unchanged, through a `CONCIERGE`-scoped `RunSpec`. No new streaming loop.
- The front-gate reuses the shipped origin allowlist, rate limiter, and injection scanner.
- The widget gained one new column, `agent_id`, mirroring the `workspace_id` column that shipped with W4a.

## Security posture (worth your eyes at the gate)

This is a public, anonymous, prompt-injectable surface, so the tool lockdown is the load-bearing part.

- An adversarial review of the credential foundation found no live bypass. It did surface a cross-tenant mint gap (a caller could bind a concierge to another workspace's pocket), now fixed at the sole Site-writer with the same ownership check every other pocket path runs.
- A primary-source read of `claude_sdk` showed an allowlist alone does not lock down a public agent: built-in tools like `Bash`/`Write` are never stripped by the allow path, and a universal grant plus always-allowed servers survive it. `deny_mcp_tool_ids` is the only real lever, so the `CONCIERGE` surface hard-strips web, code/filesystem/subagent, and pocket-write tools, and locks the KB read to `pocket:<id>`.
- The one gap a static deny cannot close is composio connectors (dynamic per-workspace ids, surviving the always-allowed server). For the pilot, a concierge run refuses fail-closed if its pocket exposes any connector. The proper fix, an untrusted mode in `claude_sdk` that drops the universal grant and the always-allowed bypass for a `CONCIERGE` run, is written up as a blocker before any non-pilot use. It touches shared tool-gating, so it needs full-suite plus flag-mode validation.

## Tests

63 tests across the resolver, the widget store, and the concierge endpoint, covering the fail-closed paths: bad, empty, non-string, and revoked keys; disallowed origin; injection; sibling-pocket and cross-workspace widgets; cross-tenant mint; connector present; and connector-check error. import-linter clean on both configs.

## Not in this PR

Live voice (Tier 2, which adopts the deferred LiveKit Agents migration), the widget chat UI (separate repo), self-serve onboarding, and the connector GA fix. Deploy note: `agent_id` is an additive `CREATE TABLE` add, so confirm no deployed `paw_bar.db` exists or add an `ALTER` before cloud deploy.

### DIFF
--- a/ee/pocketpaw_ee/cloud/_core/context.py
+++ b/ee/pocketpaw_ee/cloud/_core/context.py
@@ -16,6 +16,14 @@ Updates:
     back to the standard JWT auth path when any check fails. Used by the
     foresight router; the dev-grade bypass tightens to a short-lived
     signed JWT in a follow-up PR.
+  - 2026-07-14 (Paw Bar concierge seam, T1): added ``ScopeKind.CONCIERGE``
+    — the scope a public, origin-bound Paw Bar embed key resolves into
+    (``site_keys.resolve_site_key``) — and a trailing ``pocket_id`` field on
+    ``RequestContext``. The concierge is bound to a Site's ``pocket_id`` (its
+    KB scope downstream is ``pocket:<pocket_id>``), which the workspace/user
+    axes don't carry, so the resolved envelope needs its own slot. The field
+    is optional (default ``None``) and trailing, so every existing keyword
+    construction is unaffected — no caller passes it yet.
 """
 
 from __future__ import annotations
@@ -46,6 +54,12 @@ class ScopeKind(StrEnum):
     GROUP = "group"
     DM = "dm"
     NONE = "none"
+    # A public, origin-bound Paw Bar embed key (Site.signed_key) resolves into
+    # this scope via ``auth.site_keys.resolve_site_key``. It is NOT a signed-in
+    # user session: the credential is world-visible (embedded in a foreign
+    # site's HTML), so the request's authority is the resolved Site's
+    # ``scopes`` + ``workspace`` + ``pocket_id``, never full workspace access.
+    CONCIERGE = "concierge"
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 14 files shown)


prakashUXtech:
Refinement on the connector posture, after tracing the composio runtime identity.

The fail-closed guard refuses on NATIVE connectors (via `list_pocket_connectors`, pocket or workspace scoped). Those execute with the tenant's credentials, so they are the real leak, and that is the right check.

Composio turned out not to be a tenant-data risk. It resolves its runtime identity through `current_user_id()`, which for a concierge run is the anonymous visitor's `customer_ref`. So composio acts as the visitor's own empty account, not the tenant's, and cannot reach the tenant's connected accounts. The residual composio surface is just discovery and OAuth-initiate meta-tools that the visitor can't complete as the tenant, which folds into the same GA lockdown.

Net: the pilot guard is correctly scoped, and the GA-blocker (an untrusted mode in claude_sdk) is a little narrower than the PR description implied.

Also pushed a small cleanup (`9314fd49`) removing a dead draft helper that a concurrent edit swept into the lockdown commit. No behavior change; the inline guard is the one that runs.

prakashUXtech:

PR #1715, opened 2026-07-15
Fix Paw Bar decision-loop tenancy: store split + cross-tenant approve

prakashUXtech:
## What this fixes

Two tenancy defects in the Paw Bar customer-decision loop, both found while grounding the glass-bar design and reproduced with failing tests before any fix.

### 1. Proposals were dropped or mis-scoped (store split)

When a visitor's event raised an Instinct proposal, the write went to the bare instinct store with the workspace resolved from the widget owner. But the widget already carries a real, server-stamped `workspace_id` (the old code comment claiming it does not was stale). On the public, unauthenticated ingest path there is no active workspace, so in workspace-scoped (cloud) mode the bare store raised `WorkspaceScopeRequired`, the best-effort loop swallowed it, and the proposal was dropped entirely. The owner never saw it. In shared mode it was stamped with the owner label the dashboard never filters by.

Now the write routes through the same per-workspace store the dashboard reads from, keyed by the real `workspace_id`, with the owner kept as the human assignee.

### 2. A cross-tenant approve was possible (missing assert)

The approve, reject, and bulk paths run a per-blob workspace assertion for every gated blob kind except the customer-reply blob, which had a delivery hook but no assert. Since the delivery hook routes the reply to the blob's own workspace, an approver in one workspace could approve a customer-reply action belonging to another and flip a decision row in that other tenant.

This adds `_assert_customer_reply_workspace`, mirroring the existing peer assert: it fails closed on a missing workspace claim and refuses a cross-workspace approve with a 403. It is wired into all four state-flip entry points: approve, reject, bulk-approve, bulk-reject.

## Verification

Both defects were reproduced with failing tests first, then fixed. The store-split test drives the real store factory in workspace-scoped mode (it deliberately does not mock the store, which is how the earlier test hid the split). The cross-tenant test exercises the 403 on all four wirings plus a same-workspace control that still succeeds. Regression: the decision-loop, tenancy, customer-reply, and paw_bar backend suites pass, and import-linter is clean on both configs. An independent adversarial review confirmed there is no fifth un-gated approve path (it walked every state-flip site in the tree, including the mission-control bulk facade and the separate approvals router) and no blob-shape bypass.

## Follow-ups (not in this PR)

- Bulk approve and reject flip a customer-reply's status but do not run the delivery hook, so a bulk-approved reply never reaches the customer. Fail-safe and pre-existing, worth a ticket if bulk delivery is wanted.
- A legacy widget with no `workspace_id` would fall back to the owner label and then always fail the new assert. This fails closed and is unreachable today (widget creation always stamps a real `workspace_id` and there are no deployments), but the fallback and the assert should be reconciled.
- Pre-existing and unrelated: `edit_proposal` calls the store with no workspace argument where one is required, a 500 in production that the tests hide with a no-arg mock. Flagging only.

### DIFF
--- a/ee/pocketpaw_ee/instinct/router.py
+++ b/ee/pocketpaw_ee/instinct/router.py
@@ -1,5 +1,15 @@
 # ee/instinct/router.py — FastAPI router for the Instinct decision pipeline API.
 # Created: 2026-03-28 — Propose, approve/reject, list pending, query audit.
+# Updated: 2026-07-15 (B0 H2 — _customer_reply tenancy gate) — closed a gap in the
+#   FOUR-path dispatch: the paw-bar ``_customer_reply`` blob (carried by
+#   ``decision_loop.propose_customer_decision``) had a delivery hook on the
+#   approve/reject paths but NO workspace assertion, while every other gated blob
+#   kind did. Added ``_customer_reply_blob`` + ``_assert_customer_reply_workspace``
+#   (peers of the ``_external_action`` pair) and wired the 403 into ALL FOUR
+#   locations (approve / bulk-approve / reject / bulk-reject) BEFORE any delivery.
+#   Without it, a ws-A admin could approve/reject a ws-B ``_customer_reply`` action
+#   and ``deliver_customer_decision`` would flip a decision row in ws-B (it routes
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1716, opened 2026-07-15
Paw Bar frame endpoint + CSP frame-ancestors origin model

prakashUXtech:
Stacks on #1714 (the concierge seam). Review this on top of it.

## What this adds

The iframe FRAME endpoint and the CSP-based origin model for the glass bar. It is the security foundation for embedding the bar as a sandboxed iframe instead of inline in the host page.

## How it works

`GET /paw-bar/frame?key=...` looks up the Site by its public embed key and server-renders the iframe document with a `Content-Security-Policy: frame-ancestors <the Site's allowed origins>` header. The browser then refuses to render the iframe inside any parent that is not on the Site's allowlist. This moves the per-embedder gate off the per-request `Origin` header (which stops identifying the embedder once the iframe is served from our own domain) and onto the document's CSP, which is the control Intercom and Stripe Elements use. An empty or all-unusable allowlist fails closed with a 403 rather than rendering an ungated frame.

The `/paw-bar/chat` origin check becomes dual-mode: a legacy inline-widget request is still gated against the Site's allowed origins (fail-closed), and an iframe-mode request is accepted when its `Origin` is exactly our configured frame origin (the embedder having already been gated by the CSP at frame-render). With no frame origin configured the behavior is byte-identical to before, so the frozen inline widget keeps working.

Each owner-controlled allowed-origin value is reduced to a strict host-source shape before it reaches the header, so it cannot inject a second CSP directive or split the header. The loader-supplied parent origin is validated against the same allowlist before it is used as the glass app's postMessage target.

## Honest residual

The CSP binds browsers only. The embed key is world-visible and a raw curl POST to the chat endpoint (spoofing any `Origin`) was always possible and is unchanged. CSP does not close that. The real controls remain the rate limit, the injection screen, and the zero-authority concierge scope. The dual-mode change opens no new browser-reachable hole: a browser cannot spoof its `Origin`, and cross-origin fetches are CORS-gated.

## Verification

53 tests, including header-injection, script-breakout, revoked key, and the dual-mode gate. import-linter clean on both configs. An independent adversarial review walked all six vectors (CSP injection and splitting, frame fail-open, the Host-header/proxy dual-mode case, targetOrigin leaks, the static mount, origin normalization) and found no blocker; its two info-level notes are folded in (a comment locking the sanitize ordering against a future CSP-injection regression, and None-safe allowlist guards).

## Deploy note

Behind a proxy or in any multi-origin deploy, set `PAWBAR_FRAME_ORIGIN` to the fixed public frame origin so the frame-origin check cannot be Host-header influenced. Single-origin and local deploys work on the default. The glass app assets (built separately) go in `PAWBAR_APP_DIR`.

### DIFF
--- a/ee/pocketpaw_ee/cloud/__init__.py
+++ b/ee/pocketpaw_ee/cloud/__init__.py
@@ -694,6 +694,25 @@ def mount_cloud(app: FastAPI) -> None:
     uploads_dir.mkdir(parents=True, exist_ok=True)
     app.mount("/uploads", StaticFiles(directory=str(uploads_dir)), name="uploads")
 
+    # Paw Bar glass app (A1) — the iframe served by GET /api/v1/paw-bar/frame loads
+    # pawbar.js + pawbar.css from THIS root-absolute mount. The dir is configurable
+    # (PAWBAR_APP_DIR) so the built Vite/Svelte dist can be dropped in at deploy/smoke
+    # time; it defaults under ~/.pocketpaw and is created empty so the mount always
+    # binds (an empty dir just 404s the asset until the real bundle is copied in).
+    # PAWBAR_APP_MOUNT is imported from the router so the mount path and the frame
+    # HTML's <script src> can never drift.
+    from pocketpaw_ee.paw_bar.router import PAWBAR_APP_MOUNT
+
+    pawbar_app_dir = Path(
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1719, opened 2026-07-15
Sandbox registry for the browser IDE

DevRohit06:
First slice of the Web Cursor browser IDE: the registry that every later piece authorizes against. Nothing calls it yet — it's the foundation the VM provisioning, terminal, editor, and git broker slices build on.

## What's here

A new cloud entity, `websandbox/`, on the standard 4-file shape, plus a `WebSandbox` Beanie document. It maps `(workspace_id, user_id, repo)` to a sandbox id, a lifecycle state, and a GitHub App installation pointer — one row per key, enforced by a unique compound index and an app-level upsert.

The point of the slice is `authorize_sandbox`. It looks a sandbox up under a tenant filter and grants access only when the row is in the caller's workspace and owned by the caller. Anything else — wrong workspace, wrong owner, or no such sandbox — records a `vm.cross_tenant_denied` audit event and then raises the same opaque `Forbidden`, so a probe can't tell a missing sandbox from someone else's. The audit write happens before the raise and can't fail the request (it's fire-and-forget), so a denial is always on the record.

## Tests

12 tests covering the round-trip, idempotent registration, owner-allowed, the three denial paths (cross-workspace, cross-user, unknown), that a denial writes the audit event and an allow does not, per-user list scoping, and that the domain object can't be built without tenancy fields.

```
12 passed in 1.50s
```

Verified `WebSandbox` is wired into `get_all_documents()` so the collection actually initializes.

## Notes

- Endpoints are gated by license + an authenticated context; per-action RBAC needs new action strings in the catalog and is deferred to a later slice (noted in the router header).
- `authorize_sandbox` keys on the Daytona `sandbox_id`; `get`/`update` key on the registry row id. Called out in case the session-WebSocket slice wants the row id instead.
- `installation_id` is stored plain for now (an installation id, never a token). The GitHub App slice encrypts it at rest.

### DIFF
--- a/ee/pocketpaw_ee/cloud/__init__.py
+++ b/ee/pocketpaw_ee/cloud/__init__.py
@@ -260,6 +260,7 @@ def mount_cloud(app: FastAPI) -> None:
     from pocketpaw_ee.cloud.rules.router import router as rules_router
     from pocketpaw_ee.cloud.sessions.router import router as sessions_router
     from pocketpaw_ee.cloud.skills.router import router as skills_router
+    from pocketpaw_ee.cloud.websandbox.router import router as websandbox_router
     from pocketpaw_ee.cloud.workspace.router import router as workspace_router
 
     app.include_router(auth_router, prefix="/api/v1")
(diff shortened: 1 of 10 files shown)


DevRohit06:

PR #1720, opened 2026-07-15
Scoped agent discovery + gallery presentation fields

prakashUXtech:
Backend groundwork for the Agent Gallery + Studio work. Two additive changes to the cloud agents domain — no behavior change for existing callers, no migration.

## Scoped discovery
`POST /agents/discover` gains a `scoped` flag (default `true`). When scoped, the default viewer union is strictly `{owner == me} ∪ {visibility == workspace}` within the caller's workspace — another member's `public` agent can no longer appear. Pass `scoped=false` to restore the previous cross-workspace public union. The explicit `visibility=private|workspace|public` modes are unchanged.

This is the guarantee the Gallery leans on: a user browsing their agents sees their own plus what their team has shared, never a stranger's public agent. (The public storefront is a separate, later surface.)

## Gallery/Studio presentation fields
Additive, defaulted fields so the Studio can author them and the Gallery can render them:

- `AgentConfig`: `welcome_message`, `conversation_starters`, `voice`, `appearance`
- `Agent`: `tags`

All default to empty (`""`, `[]`, `None`, `{}`), so existing agent docs load unchanged. Threaded through the doc↔domain↔wire path; `visibility`/`owner`/`workspace` semantics untouched; writes stay in `service.py`.

## Tests
New `tests/cloud/test_agent_discover_scope.py` (6 tests) against the real Beanie/mongomock query path. The load-bearing case seeds the viewer's private agent, a peer's workspace-shared agent, and two peer `public` agents (same workspace + cross-workspace), then asserts the scoped default returns exactly the first two. Round-trip coverage for the new fields. `217 passed` across the agent test modules; ruff clean.

## Scope
Backend only. The Gallery, Studio polish, and Publish UX land as separate paw-enterprise PRs that consume this contract. A follow-up hardening item (enforcing visibility on the direct read + group/pocket-attach paths) is tracked separately — pre-existing, out of scope here.

### DIFF
--- a/ee/pocketpaw_ee/cloud/agents/domain.py
+++ b/ee/pocketpaw_ee/cloud/agents/domain.py
@@ -9,11 +9,19 @@ Beanie sub-model would require touching every caller of
 Updated: 2026-06-28 (feat/aiam-agent-revoke, AW-4) — added ``Agent.disabled``
 mirroring the new top-level flag on the Beanie doc, so callers (and the wire
 dict) can read whether an agent has been soft-disabled / revoked.
+
+Updated: 2026-07-15 (feat/agent-scoped-discover-fields, ASG-1) — added the
+additive presentation fields that mirror the new Beanie fields:
+``AgentConfigSpec.welcome_message`` / ``conversation_starters`` / ``voice`` /
+``appearance`` and ``Agent.tags``. Lists stay frozen-friendly tuples (matching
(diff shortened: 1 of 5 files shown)


prakashUXtech:
Superseded by #1774 — the Agent Studio & Gallery work is consolidated into one integration branch off current dev, which is where it'll be finished and shipped. Closing in favour of that.

PR #1721, opened 2026-07-15
Provision a VM and clone a repo into it

DevRohit06:
Second slice of the browser IDE, stacked on the sandbox registry (#1719). This one boots a real VM and gets code into it — no auth yet, a public repo URL is enough to prove the runtime works end to end.

## What's here

`POST /websandbox/open` takes a public repo URL, registers a row, provisions a cold Daytona sandbox, waits for it to boot, clones the repo in with no credentials, and marks the row ready with the sandbox id bound. `GET /websandbox/{id}/tree` authorizes the caller against the registry and lists the tree. Orchestration lives in a new `provision.py` in the service layer; the registry doc is still only ever touched through `service.py`.

A background reaper sweeps sandboxes whose row has gone idle past a TTL, stops and deletes the VM, and marks the row reaped. If the delete fails it leaves the row for the next sweep rather than marking it reaped over a VM that's still alive. It's wired into the cloud lifespan under the existing scheduler gate, with `POCKETPAW_WEBSANDBOX_IDLE_TTL_SECONDS` (1800), `POCKETPAW_WEBSANDBOX_REAP_INTERVAL_SECONDS` (300), and `POCKETPAW_WEBSANDBOX_REAPER_ENABLED`.

One thing worth calling out from reading the Daytona docs: `auto_stop_interval` is in minutes, not seconds. The SDK default of 3600 means a VM sits idle for 60 hours before it self-stops. This slice passes 30 explicitly as a backstop under the reaper. The legacy daytona path still passes 3600 — worth a separate look.

## Tests

9 unit tests over a faked Daytona client (injected through a DI seam, so nothing in CI hits real infra): open provisions/clones/marks-ready, tree enforces tenancy, the reaper reclaims an idle VM and spares a fresh one, and a missing Daytona client surfaces a clean error instead of crashing. WC-1's 12 registry tests still pass.

```
21 passed in 3.28s
```

I also ran a one-off live smoke against real Daytona (kept out of CI): it provisioned a VM, cloned octocat/Hello-World, listed the tree with the real README and .git present, and deleted the VM in a finally block. Output confirms the delete.

## Notes

- A cross-tenant caller on `/tree` gets a not-found, not a forbidden — the row is resolved through the owner-scoped read first, so existence never leaks. `authorize_sandbox` stays as a second guard on the sandbox id.
- The clone lands in the sandbox home (`get_project_dir()` returns `/root`), so the tree shows home dotfiles alongside the repo. A later slice can move it to a dedicated subdir if we want a cleaner root.
- Tree listing is single-level for now.

Stacked on #1719 — review/merge that first. Merge this base with rebase, not squash, per the stacked-PR rule.

### DIFF
--- a/ee/pocketpaw_ee/cloud/__init__.py
+++ b/ee/pocketpaw_ee/cloud/__init__.py
@@ -935,6 +935,30 @@ def mount_cloud(app: FastAPI) -> None:
         async def _stop_fabric_ingest() -> None:
             await stop_fabric_ingest(app)
 
+    # Web Cursor idle-TTL reaper (WC-2, feat/websandbox-vm-provision). Every
+    # ~5 minutes (POCKETPAW_WEBSANDBOX_REAP_INTERVAL_SECONDS) it reclaims
+    # dropped browser-IDE sandboxes: rows in ``ready``/``opening`` whose
+    # ``updated_at`` is older than the idle TTL
+    # (POCKETPAW_WEBSANDBOX_IDLE_TTL_SECONDS, default 1800) get their Daytona VM
+    # stopped+deleted and the row marked ``reaped``. Same scheduler gate as the
+    # cycle/decisions/ingest loops so pytest runs never spawn a background loop
+    # that outlives the test; production sets POCKETPAW_CLOUD_SCHEDULER_ENABLED=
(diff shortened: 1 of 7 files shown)


DevRohit06:

PR #1722, opened 2026-07-15
Stream a real terminal from the VM to the browser

DevRohit06:
Third slice, stacked on the VM provisioning (#1721). The VM has a shell — this gets it into the browser over one authenticated WebSocket, which is what the old cloud terminal never managed (it imported a function that didn't exist, and the local path was a Unix-only singleton).

## What's here

A `/ws/websandbox/{row_id}` WebSocket endpoint and a socket-agnostic `PtyBridge`. The bridge holds the Daytona PTY handle directly (the shared `DaytonaClient.create_pty_session` throws the handle away, so there's no way to send keystrokes through it) and pumps VM output to a sink; the endpoint wires that sink to binary WebSocket frames, which is what xterm wants.

Auth reuses the existing single-use ws-ticket, because a browser can't put an Authorization header on a WebSocket upgrade. The connect path is fail-closed the whole way down: license, then ticket, then active workspace, then an owner-scoped registry read, then a ready check, then the `authorize_sandbox` oracle on the Daytona id. Any failure closes 1008 before the socket is even accepted, so a PTY only ever opens for a caller who owns that sandbox. Someone else's ticket for your `row_id` never gets past the registry read.

Client sends `{type:input}` and `{type:resize}`; the server streams output as binary and answers `{type:ping}` with a pong to survive idle proxies. Live traffic bumps the row's `updated_at` (throttled to once a minute) so the idle reaper from the last slice doesn't reclaim a session someone's actively using. On disconnect the `finally` kills the PTY so a closed tab never leaks a shell in the VM.

## Tests

8 unit tests: a valid ticket on an owned sandbox opens the PTY and round-trips input/resize/output; an invalid or missing ticket closes 1008 with no PTY; a valid ticket on a sandbox you don't own is denied before any PTY; disconnect kills the session; the activity throttle collapses a keystroke burst into one write. WC-1/WC-2 still green.

```
29 passed in 4.29s
```

Live smoke against real Daytona (out of CI): provisioned a VM, opened a PTY through the same bridge code the socket uses, sent `echo paw-smoke-ok`, and the real shell echoed it back. VM deleted in a finally block.

## Notes

- The ticket is consumed before the authz check, so a denied connect still spends it. That's intended — a ticket is one connection attempt.
- The prompt I worked from said `handle.send(bytes)`; the real SDK method is `handle.send_input`. Used the real one, confirmed live.
- Skipped the optional `sb.refresh_activity()` SDK call — the reaper reads the registry `updated_at`, and the VM's own 30-minute auto-stop is an independent backstop.

Stacked on #1721 (which is stacked on #1719). Merge the bases with rebase, not squash.

### DIFF
--- a/ee/pocketpaw_ee/cloud/__init__.py
+++ b/ee/pocketpaw_ee/cloud/__init__.py
@@ -704,6 +704,13 @@ def mount_cloud(app: FastAPI) -> None:
 
     app.add_api_websocket_route("/ws/cloud", websocket_endpoint)
 
+    # Web Cursor terminal WS (WC-3) — one PTY-over-WS shell per session, bound to
+    # the owning sandbox's Daytona VM. Mounted at root (not /api/v1) so the SPA
+    # connects to ws://host/ws/websandbox/<row_id>?token=<ws_ticket>.
+    from pocketpaw_ee.cloud.websandbox.ws import terminal_websocket_endpoint
+
+    app.add_api_websocket_route("/ws/websandbox/{row_id}", terminal_websocket_endpoint)
+
     # License endpoint (no auth)
     @app.get("/api/v1/license", tags=["License"])
     async def license_info():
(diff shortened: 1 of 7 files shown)


DevRohit06:

PR #1723, opened 2026-07-15
File read, write, and list over the session socket

DevRohit06:
Backend half of the editor slice, stacked on the terminal WebSocket (#1722). The editor UI is a separate PR — this is the transport it saves through: file operations multiplexed onto the same authenticated session socket, no separate REST write route.

## What's here

A socket-agnostic `FileRpc` helper (mirrors the terminal's `PtyBridge`) and three new frame types on `/ws/websandbox/{row_id}`: `file.list`, `file.read`, `file.write`. Each request carries a client `reqId` that the response echoes, because the socket is async and multiplexed with terminal output — file responses are typed JSON text frames, terminal output stays binary, so the two streams never collide. The socket is already owner-authorized on connect, so file ops don't re-authorize per frame; they do bump the same throttled activity heartbeat.

## The path jail

This is the security core. Every path is relative to the sandbox project dir. It's resolved by lexically normalizing `project_dir/rel` with posixpath (the VM is Linux; using posixpath keeps it correct even though CI runs on Windows) and asserting the result is the project dir or under it. Absolute paths and `..` escapes fail that check and get a `file.error` before any download or upload runs.

The jail is lexical, not realpath-based, and that's deliberate: resolving a symlink on a remote VM filesystem needs a shell round-trip, and it buys nothing here — the socket is already bound to the caller's own VM, so a symlink can only ever point at files that caller already reaches through the terminal. It's documented in the file header rather than left as an unspoken gap. Reads and writes are size-capped (`POCKETPAW_WEBSANDBOX_MAX_FILE_KB`, default 1024). A missing file on read returns an honest error, not empty success.

## Tests

33 unit tests over the helper and the frame dispatch with a faked Daytona client, plus the WC-3 terminal tests as regression since this edits `ws.py`. The must-pass security cases: `../../etc/passwd`, an absolute path, and a `..`-containing path all get rejected and never call download/upload.

```
60 passed in 4.87s
```

Live smoke against real Daytona (out of CI): wrote a file, read it back and matched, listed the dir and saw it, then tried a `../escape.txt` write and got the rejection. VM deleted in a finally block.

Stacked on #1722 → #1721 → #1719. Merge bases with rebase, not squash.

### DIFF
--- a/ee/pocketpaw_ee/cloud/websandbox/files.py
+++ b/ee/pocketpaw_ee/cloud/websandbox/files.py
@@ -1,0 +1,255 @@
+# files.py — Web Cursor file read/write/list RPC core (WC-4a).
+# Created 2026-07-15 (feat/websandbox-file-rpc).
+#
+# Socket-agnostic sibling of terminal.py's PtyBridge: this is the file-operation
+# half of the session WebSocket, multiplexed alongside the terminal on the SAME
+# authenticated socket. The browser editor round-trips directory listings, file
+# reads, and file writes through here; ws.py just parses one JSON frame and calls
+# ``FileRpc.dispatch``. Keeping the logic here (taking a DaytonaClient + a
+# sandbox_id, resolving the project dir lazily) makes the whole path-jail + size
+# guard unit-testable with a fake client and no socket.
+#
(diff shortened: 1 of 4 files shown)


DevRohit06:

PR #1724, opened 2026-07-15
Let the terminal socket authenticate by first-message frame

DevRohit06:
Small auth fix on the terminal WebSocket, needed before the editor frontend connects to it.

The terminal socket from WC-3 only authenticates via `?token=` in the URL. The frontend WS convention moved off that back in June (REVIEW-4) because a ticket in the query string leaks into access logs, browser history, and proxies — the chat socket already accepts a first-message `{type:"auth", ticket}` frame instead. This brings the terminal socket to parity so the new editor can connect the same leak-free way.

## What changed

`terminal_websocket_endpoint` now resolves the user by either path, then runs the same unchanged authz flow (workspace → owner-scoped sandbox read → ready check → `authorize_sandbox` → open PTY):

- `?token=` ws-ticket, consumed before accept (unchanged).
- When there's no `?token=`: accept first, read the first frame under a 5s timeout, parse the auth frame (`token` accepted as an alias for `ticket`), consume the ticket. Timeout, a malformed or non-auth frame, or a bad ticket closes 4001 and never reaches the message loop.

`accept()` is called exactly once on every path, tracked by an `accepted` flag so the shared authz section doesn't double-accept. Framing, file-RPC, and teardown are untouched — this is only the auth resolution.

## Tests

15 tests: the original `?token=` cases still pass, plus the new frame path — happy path, the `token` alias, malformed/non-auth/missing first frame all closing 4001, an invalid ticket closing 4001, and a valid ticket on an unowned sandbox closing 1008 after accept.

```
15 passed in 3.86s
```

One existing test was repurposed rather than kept: the old "no token closes 1008 before accept" asserted a contract that no longer exists — no token now routes to the frame path — so it's replaced by "no token and no auth frame closes 4001." All genuine `?token=` cases are unchanged.

Stacked on #1723. Merge bases with rebase, not squash.

### DIFF
--- a/ee/pocketpaw_ee/cloud/websandbox/ws.py
+++ b/ee/pocketpaw_ee/cloud/websandbox/ws.py
@@ -7,15 +7,35 @@
 # single-use ws_ticket auth, because browsers can't set auth headers on a WS
 # upgrade.
 #
+# 2026-07-15 (WC-4b): the ticket now resolves via EITHER of two auth paths, so
+# the browser can authenticate without leaking the ticket in the URL (URL query
+# strings leak into access logs, browser history, and proxies — the same
+# REVIEW-4 reasoning that moved the chat WS off URL tickets):
+#   * Path 1 — single-use ws_ticket in ``?token=``, consumed BEFORE accept
+#     (unchanged; the only path a cross-origin WS upgrade can carry in the URL).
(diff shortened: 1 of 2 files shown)


DevRohit06:

PR #1725, opened 2026-07-15
Enforce agent visibility on read and attach paths

prakashUXtech:
Makes "private means private" a real guarantee for agents. An audit (during the Agent Studio work) found that agent `visibility` was enforced on the publish/DM path but ignored on several read and attach paths — so any workspace member could pull another user's private agent, and admins could wire one in. None of this was introduced by the gallery work; it was pre-existing. This closes it.

## What was leaking
- **Reads:** `GET /agents/{id}`, `GET /agents`, and the three knowledge-read endpoints returned any agent by id with no visibility check — a non-owner could read a private agent's full config (system prompt, persona, scopes) and its knowledge base, even across workspaces.
- **Attach:** group and pocket `add_agent` checked workspace membership but not visibility — an admin could attach another user's private agent so the whole group/pocket runs it.

## The fix
One canonical predicate, applied at every tenant-facing seam:
- `can_read_agent` / `can_use_agent` in `agents/service.py` — owner always; else a same-workspace `workspace`-visible agent; else `public`. A private agent is owner-only. (Mirrors the existing DM-path predicate.)
- Reads now go through `get_for_viewer` / `ensure_can_read`, which raise `NotFound` (not `Forbidden`) so a leaked id never even confirms the agent exists. `list_agents` gained a `viewer_user_id` filter.
- Group + pocket attach call `ensure_can_use`.
- Mutation guards (`require_agent_owner_or_admin`) are unchanged — non-owners still can't edit.

Internal/privileged callers (planner, KB aggregation, the run pool) keep the unguarded `get` / unfiltered `list_agents` — every caller was checked so no legitimate system path breaks.

## Reproduce-first
`tests/cloud/agents/test_agent_visibility_enforcement.py` — 19 tests, written to fail first (**13 failed before the fix, 19 pass after**): the predicate matrix, deny-non-owner-private / allow-owner / allow-workspace-member / deny-cross-workspace for reads, `list_agents` exclusion, the three knowledge reads, and both attach paths — with positive controls so it doesn't over-tighten. Broad regression across agents/chat/pockets is green except 7 failures that are pre-existing on `dev` (confirmed by stashing this change — identical failures without it).

## Known residual (follow-up, not in this PR)
Two chat-preamble builders (`surface/handlers/agent.py`, `agents.py`) still stamp the **name/slug** of same-workspace private agents into an agent's context — low severity (no config or KB), a different module. Tracked as a small follow-up; happy to fold in if preferred.

## Note
Co-touches `agents/service.py` with pocketpaw#1720 (different functions), so a trivial 3-way at merge.

### DIFF
--- a/ee/pocketpaw_ee/cloud/agents/router.py
+++ b/ee/pocketpaw_ee/cloud/agents/router.py
@@ -18,6 +18,15 @@ soft-disable / revoke-everywhere flow. Both carry the SAME owner/admin guard
 (``require_agent_owner_or_admin``) + ``current_user_id`` as DELETE — symmetric
 tenant-scope protection, so a cross-workspace / non-owner caller cannot revoke
 or restore an agent it doesn't own.
+
+Updated 2026-07-15 (fix/agent-visibility-enforcement, ASG-7): the READ endpoints
+now enforce agent visibility. ``GET /agents/{id}`` routes through
+``agents_service.get_for_viewer`` (404 for another user's private agent);
+``GET /agents`` passes ``viewer_user_id`` so the list hides others' private
+agents; and the three knowledge READS (``/knowledge``, ``/knowledge/{id}``,
+``/knowledge/search``) call ``agents_service.ensure_can_read`` before hitting kb.
+Previously any licensed workspace member could read any agent's config / KB by
+id. ``GET /agents/{id}/scope`` was already owner-gated and is unchanged.
 """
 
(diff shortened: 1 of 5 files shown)


prakashUXtech:
Superseded by #1774 — the Agent Studio & Gallery work is consolidated into one integration branch off current dev, which is where it'll be finished and shipped. Closing in favour of that.

PR #1726, opened 2026-07-15
Show the cloned files: unify the sandbox on /home/daytona

DevRohit06:
Bugfix found by actually running it: you clone a repo, it succeeds, but the file tree is empty. Two causes, both fixed here, verified live against a real Daytona VM cloning a real repo.

## 1. The tree asked for the root and got rejected

The editor lists the repo root by sending an empty path. The file-RPC jail treated an empty path as an error, so the root listing always failed and the tree showed nothing. Empty (and `.`) now resolve to the jail root for `list`; `read`/`write` still require a real file path.

## 2. The clone, the terminal, and the tree disagreed on the directory

The clone and the tree used `get_project_dir()`, which returns `/root` on this Daytona image. But the terminal PTY opens in `/home/daytona`. So the repo landed in `/root`, the shell started in an empty `/home/daytona`, and the tree listed a directory that didn't match either in a consistent way. Checking the VM directly showed the disconnect.

A single pinned constant — `WEBSANDBOX_WORKDIR = /home/daytona` — now drives the clone target, the jail root, and the PTY cwd, so all three point at the same place. `/home/daytona` is the sandbox user's home and starts empty, so the repo lands there cleanly.

## Proof

Live run against a real VM cloning a real project repo:

```
cloned into /home/daytona
FileRpc.list_dir('') -> 21 entries: ['.git', 'README.md', 'package.json', 'docs', ...]
terminal: root@...:/home/daytona# pwd && ls  ->  /home/daytona + the repo files
PASS: clone + tree('') + terminal all agree on /home/daytona
```

69 websandbox unit tests pass, including a new regression that `list_dir('')` returns the root listing and that the jail root defaults to the workspace dir.

Stacked on #1724. Merge bases with rebase, not squash.

### DIFF
--- a/ee/pocketpaw_ee/cloud/websandbox/constants.py
+++ b/ee/pocketpaw_ee/cloud/websandbox/constants.py
@@ -1,0 +1,13 @@
+# constants.py — shared Web Cursor sandbox constants.
+# Created 2026-07-15: single source of truth for the in-VM workspace directory.
+#
+# Everything the user touches lives under ONE directory in the VM: the repo is
+# cloned here, the file tree/RPC is jailed here, and the terminal opens here.
+# We deliberately use ``/home/daytona`` (the sandbox user's home) rather than the
+# SDK's ``get_project_dir()`` — which on the current Daytona image returns
+# ``/root``, mismatching the PTY's default cwd (``/home/daytona``). Pinning one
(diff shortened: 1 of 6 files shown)


DevRohit06:

PR #1727, opened 2026-07-15
Web Cursor: cloud IDE backend (registry, VM, terminal, files)

DevRohit06:
Consolidates the Web Cursor backend stack into one branch (the captain asked for a single `feat/code`; that name collides with an existing `feat/code/cloud` ref, so this is `feat/code-cloud`). Everything a browser IDE session needs on the server side: a tenant-isolated cloud VM you own, a streaming terminal, and file read/write persisted to the VM — over one authenticated, fail-closed WebSocket. Supersedes the six stacked PRs (#1719, #1721, #1722, #1723, #1724, #1726), now closed in favor of this branch.

## What's in it

- **Sandbox registry** — maps (workspace, user, repo) to a VM with fail-closed cross-tenant authorization and an audit trail.
- **Cold VM provision + clone** — opens a Daytona VM, clones a public repo into it, tracks lifecycle, reaps idle VMs. Auto-stop is set in minutes (the SDK default of 3600 would idle a VM for 60 hours).
- **Terminal over WebSocket** — a real PTY streamed to the browser, authenticated by a single-use ticket sent as a first-message frame (no ticket in the URL).
- **File RPC** — list/read/write over the same socket, jailed to the workspace dir with traversal rejection.
- **/home/daytona unification** — the clone target, the file-tree root, and the terminal cwd all point at one directory, so a cloned repo actually shows up in the tree and the shell. Verified live against a real VM cloning a real repo.

## Tests

The full websandbox suite passes (registry, provision, terminal, file RPC, workdir), plus live smokes against real Daytona for provisioning, the terminal, file persistence, and the workdir fix.

Pairs with the frontend editor shell in paw-enterprise #623.

PR #1729, opened 2026-07-15
Backend AI edit agent and auto feature branch

DevRohit06:
The backend half of Cmd-K editing, stacked on the cloud IDE backend (#1727). The per-hunk review UI is a separate frontend slice — this generates the proposal and never applies anything.

## What's here

Two things:

**Auto feature branch.** When a repo opens into a VM, the provision flow now creates and checks out `paw/edit-<hex>` right after the clone, so AI edits never touch the checked-out default. The branch is stored on the registry row and surfaced on the response. Proven end to end against real Daytona — the smoke confirms the branch is checked out inside the VM.

**Edit endpoint.** `POST /websandbox/{id}/edit` takes a path, an instruction, and an optional selection. It authorizes the caller, jails the path to the workspace root, reads the file, gathers a little ripgrep context, and asks a model for a rewrite — then returns the original and proposed content for the frontend to review per hunk and save through the existing file RPC. It's generate-only: it never writes to the VM. The agent runs entirely backend-side (reads and greps over DaytonaClient network calls) — no in-VM agent, no UDS bridge.

## Tests

6 unit tests over a faked model client and faked Daytona (proposed content flows through, cross-tenant denied before any model call, path traversal rejected, a model failure surfaces a clean error not a crash), plus a new provision test proving the auto-branch is created and bound. Full websandbox suite green.

```
16 passed (edit + provision) · 67 passed (rest of websandbox) · import-linter clean
```

## One thing needs your input: the model seam

The edit agent calls the model the same way the decisions/explain narrator does — direct `AsyncAnthropic` with `anthropic_api_key`. But this deployment has no direct Anthropic key; it routes Claude through the `claude_code` OAuth provider and a LiteLLM proxy. I probed both the proxy's Anthropic passthrough and its OpenAI-compatible route (`deepseek-v4-flash`) and both return `no available server` right now — so the live model half of the smoke couldn't run (the code fails closed with a clean error, exactly as designed).

So: how should Cmd-K call a model here — through the same `claude_code`/agent path chat uses, or a specific LiteLLM model your proxy actually serves? The seam is a one-line swap behind `POCKETPAW_WEBSANDBOX_EDIT_MODEL` once we know the target. Everything up to the model call already runs against the real registry + real VM.

Stacked on #1727.

### DIFF
--- a/ee/pocketpaw_ee/cloud/models/web_sandbox.py
+++ b/ee/pocketpaw_ee/cloud/models/web_sandbox.py
@@ -19,6 +19,12 @@ in blob storage (an ``EEUploadService`` FileRecord id). The VM is scratch and
 gets reaped; this durable pointer lets a user's uncommitted work + workspace
 state be restored into a fresh VM. Null until the first snapshot is taken. No
 new index needed — it's read only via the owner-scoped registry row.
+
+Changed 2026-07-15 (WC-5a, feat/websandbox-edit-agent): added ``branch`` — the
+auto-created ``paw/edit-<hex>`` feature branch the repo is checked out onto in
+the VM at open time, so AI edits never touch the checked-out default branch.
+Null until the provisioner creates it (right after the clone). No new index
+needed — it's read only via the owner-scoped registry row.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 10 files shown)


DevRohit06:
Consolidated into #1730 (feat/code-mode) — same commits, one branch.

PR #1731, opened 2026-07-16
Paw Bar action registry: visitor cart, gated actions, concierge tools

prakashUXtech:
## What this adds

The Paw Bar action registry: the visitor commerce loop plus the per-verb tools the concierge agent uses to run it. Stacked on feat/paw-bar-frame.

A widget spec can now declare `actions` (verbs with a policy and typed args), a `catalog`, and a `checkout_url`. Two public endpoints let a visitor run those actions, and the concierge agent gets one tool per declared verb wired to the same code path.

## SS-2 alignment: gated verbs never execute

Following the staffed-sites rule, agent-facing action tools never run tenant-scoped effects. Only visitor-scoped state auto-fires:

- `auto` verbs (add_to_cart, checkout) touch only the visitor's own cart. This is enforced structurally: the spec validator rejects `policy: auto` on any verb other than the two built-in cart verbs.
- `gated` verbs (book a table, request a quote) never run. They raise an Instinct proposal for a human to approve, reusing the existing decision-loop delivery so the approved outcome reaches the visitor on the same poll endpoint.
- Checkout is a handoff link, rendered with an opaque cart reference. The agent never runs payment.

## Surface

- `POST /paw-bar/action` {key, w, customer_ref, verb, args} returns {ok, result, cart}
- `GET /paw-bar/cart` ?key&w&customer_ref returns {items, total_cents, currency, checkout_url}

Both carry the same armor as concierge chat, factored into one shared front-gate: fail-closed key resolution, dual-mode origin, rate limit, then the widget-to-workspace/pocket binding check. Args are structured, so there is no free-text injection screen; the shared executor validates every arg against the declared schema, rejects unknown keys, caps strings at 256 chars, and clamps qty to 1..99.

- `PATCH /paw-bar/widgets/{id}`: admin plus owner-token, workspace-scoped partial update of agent_id and the other mutable fields (a cross-tenant id gets 404). The spec stays on update_spec. The paw-enterprise agent-binding UI drives this.

## Tool injection

When a widget declares actions, the concierge run exposes one tool per verb (pawbar_add_to_cart, and so on) through a per-run, per-widget in-process MCP server built from a ContextVar that run_core binds. Each handler re-loads the live widget and calls the same execute_action, so the executor stays the single authority. The concierge profile widens its allow-list by exactly this widget's verb ids and nothing else; a widget with no actions is byte-identical deny-all. The preamble lists the available tools and the pawbar-card fence format only when actions are declared.

## Follow-ups

- No cart TTL in v1. Carts persist per (widget_id, customer_ref) with no expiry; a sweep is a follow-up.
- Warm-client toolset: the per-widget tool set is built at client-build time. Under warm-subprocess reuse a stale set could show the wrong verbs, but the handler re-validates against the live widget, so no undeclared or unauthorized effect can fire. Flagged for a later pass.

## Testing

`uv run pytest tests/cloud/test_paw_bar_actions.py tests/cloud/test_paw_bar_concierge_chat.py tests/cloud/test_paw_bar_backend.py -o addopts="" -q` gives 82 passed. The new file covers spec validation, the executor's auto and gated paths, the endpoint armor, the agent PATCH including the cross-tenant 404, tool injection on and off, and the session_key interlock. Both import-linter configs stay green (OSS-EE boundary plus the EE Beanie chokepoint).

security-review applies here: these are public, unauthenticated-by-user endpoints. The cart read uses the visitor's own customer_ref as its scope, the same trust model as the existing decision poll.

## Review hardening (follow-up commit)

Two-stage review plus security review turned up one functional gap and a set of cheap hardenings:

- Catalog to the agent: the concierge preamble carried the action verbs but not the products, so the agent declined ("I don't have a take-home list"). The widget catalog now rides surface_meta (capped at 50) and the preamble renders a compact id/name/price block, so the agent names real products and emits pawbar-card fences with real ids.
- Owner-facing proposal safety: the gated-verb proposal text (summary, title, recommendation) now sanitizes the visitor-controlled parts (strip control chars, collapse whitespace, cap the summary at 280) and demarcates them as untrusted input. This is backend defense in depth; the Tray frontend should also escape on render (separate paw-enterprise follow-up).
- Proposal spam: a dedicated lower cap (6 per minute) for gated, proposal-generating actions, separate from the overall widget rate, so a rotating customer_ref cannot flood the owner's tray.
- Read amplification: GET /paw-bar/cart records a read marker so enumeration reads count toward the rate limiter like writes.
- customer_ref bound: validated server-side at the front gate (^[A-Za-z0-9_-]{8,128}$), fail-closed 400.

## Deferred follow-ups (not in this PR)

- Server-issued or HMAC-signed customer_ref (the root fix for the handle trust model). Low practical risk today because the client mints a 128-bit crypto-random ref, so enumeration is impractical.
- Cart and decision-row TTL / GC. Carts persist per (widget_id, customer_ref) with no expiry.
- Ops note: production MUST set PAWBAR_FRAME_ORIGIN. When unset, the origin gate falls back to the request's own origin, which is fine for single-origin or local deploys but should be pinned in a proxied production deploy.

### DIFF
--- a/ee/pocketpaw_ee/agent/mcp_servers/pawbar.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/pawbar.py
@@ -1,0 +1,174 @@
+# ee/agent/mcp_servers/pawbar.py — in-process MCP server: Paw Bar action tools.
+# Created: 2026-07-16 (Paw Bar action registry, C1) — exposes ONE tool per verb a
+#   CONCIERGE widget declares (e.g. ``pawbar_add_to_cart``) so the public concierge
+#   agent can drive the visitor commerce loop through the SAME shared executor the
+#   POST /paw-bar/action endpoint uses. The tool set is PER-RUN and PER-WIDGET: it
+#   is built from the ``current_pawbar_run()`` ContextVar that ``run_core`` binds
+#   for a concierge run whose widget declares actions. When that context is absent
+#   (every non-concierge run, or a concierge widget with no actions) the builder
+#   returns None — no tools register, so the concierge tool surface stays deny-all
+#   exactly as before this slice.
+#
+#   SS-2 safety: each handler RE-LOADS the live widget by id (workspace-scoped) and
+#   calls ``execute_action``, which re-validates the verb is declared, coerces the
+#   args, and enforces the auto/gated policy. So even if a warm subprocess reuses a
+#   stale tool set, no undeclared/unauthorized effect can fire — the executor, not
+#   the tool surface, is the authority. A gated verb still only raises an Instinct
+#   proposal; the agent never executes a tenant-scoped effect.
+#
+#   Registered via the ``pocketpaw.mcp_servers`` entry point (``pawbar`` →
+#   ``CloudPawBarActionsMcpProvider``); its tool ids join the SDK allowlist through
+#   the provider's ``tool_ids()`` and are kept past the concierge restrictive
+#   allow-list because ``_concierge_profile`` names exactly this widget's verb ids.
+
+from __future__ import annotations
+
+import json
+import logging
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+SERVER_NAME = "pawbar_actions"
+
+# type-name (as declared in the spec's flat arg map) → the Python type the SDK
+# ``@tool`` schema uses for that arg.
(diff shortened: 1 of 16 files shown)


prakashUXtech:

PR #1739, opened 2026-07-16
feat(paw-bar): concierge kill switch and greeting settings

prakashUXtech:
## What

Adds an owner on/off kill switch and a greeting to the Paw Bar concierge, enforced fail-closed. Stacked on top of the paw-bar action stack (`feat/paw-bar-actions`); the diff is only this slice.

## Changes

**Site doc fields** (`ee/pocketpaw_ee/cloud/models/site.py`)
- `concierge_enabled: bool = True` and `concierge_greeting: str = ""`. Additive with backward-compatible defaults, so every existing site reads as enabled with no greeting and no migration runs.

**Kill switch, fail-closed, re-checked per request**
- When `concierge_enabled` is False the three public entry points refuse with `403 concierge_disabled`:
  - `GET /paw-bar/frame` checks it inline right after key auth, mirroring the existing empty-allowlist 403.
  - `POST /paw-bar/chat` and `POST /paw-bar/action` / `GET /paw-bar/cart` get the same 403 through `resolve_site_key`, the shared resolver both paths already run.
- The flag is read on every request (the resolver does a fresh `find_one`, nothing is cached on a warm client), so toggling it off silences the bar immediately.
- This is separate from `revoked`, which cuts the embed key at 401 for anti-enumeration. `lookup_site_by_key` stays 401-only; the kill switch is a distinct 403 on a valid key. A disabled concierge is a legitimate owner state, not a probe signal.

**Greeting into the frame config**
- `concierge_greeting` rides into the `window.__PAWBAR__` bootstrap payload as `greeting` (empty string when unset; the glass app defaults). The glass app reads it in the D4 slice.

**Owner read/update surface**
- `GET` and `PATCH /paw-bar/admin/site/{site_id}/settings`, admin-authed (`require_scope("admin")`) and workspace-scoped via `current_workspace_id`, reading and writing only the two concierge fields. A cross-tenant or malformed site id 404s before anything is read or written. PATCH is partial (`model_fields_set`).

### Why this surface

The two fields live on the **Site** doc, not the widget, so the natural owner key is the Site. The C1 widget PATCH keys on a widget id and would need a widget-to-site hop to reach these. The new endpoint reuses the same lightweight admin-auth pattern the widget CRUD already uses (`require_scope("admin")` + `current_workspace_id`), keeps the owner surface next to the enforcement it drives, and stays off the heavier sites control plane (plan-feature + action-scope gates). It touches only the two fields, so it is a small new surface rather than an extension of an unrelated update path.

This folds in staffed-sites SS-6's kill switch and greeting fields.

## Tests

New `tests/cloud/test_paw_bar_concierge_settings.py` covers: the shared resolver refuses a disabled site and still resolves an enabled one; frame, chat, and action each 403 when disabled while a default site renders; the greeting appears in the frame config; the settings GET/PATCH round-trips, is partial, and 404s cross-tenant; and toggling off silences the frame on the very next request.

```
uv run pytest tests/cloud/test_paw_bar_frame.py tests/cloud/test_paw_bar_concierge_chat.py tests/cloud/test_paw_bar_actions.py tests/cloud/test_paw_bar_concierge_settings.py -o addopts="" -q
86 passed in 6.25s
```

### DIFF
--- a/ee/pocketpaw_ee/cloud/auth/site_keys.py
+++ b/ee/pocketpaw_ee/cloud/auth/site_keys.py
@@ -47,6 +47,18 @@
 # is unchanged: the key is world-visible, so a raw ``curl`` POST was always
 # possible — CSP binds BROWSERS only; the real controls stay the rate-limit +
 # injection screen + the zero-authority CONCIERGE scope.
+#
+# Updated 2026-07-16 (Paw Bar concierge kill switch, D1 / SS-6): ``resolve_site_key``
+# now fails closed with a 403 when the resolved Site has ``concierge_enabled=False``
+# — the owner's on/off switch for the concierge (chat / action / cart all go through
+# this resolver). It is checked AFTER ``lookup_site_by_key`` (which stays 401-only to
+# preserve its anti-enumeration contract) and BEFORE the origin gate, at the point
+# where the Site is resolved, and is re-read on every request (a fresh ``find_one``,
+# never cached) so a toggle takes effect immediately. This is DISTINCT from
+# ``revoked``: ``revoked`` cuts the KEY (401, indistinguishable from a bad key);
+# ``concierge_enabled=False`` refuses the resolved concierge (403 ``concierge_disabled``).
+# The FRAME endpoint enforces the same switch inline (it calls ``lookup_site_by_key``
+# directly, not this resolver).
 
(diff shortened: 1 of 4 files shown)


prakashUXtech:

PR #1740, opened 2026-07-16
Owner aggregation reads for the per-site concierge dashboard

prakashUXtech:
Stacked on `feat/concierge-dashboard-settings` (D1). The diff here is only the D2 slice.

## What this adds

Four owner-facing, admin-authed, workspace-scoped reads for the paw-enterprise concierge dashboard, all under `/paw-bar/admin/site/{site_id}`:

- `GET .../overview` returns `{widget:{id,spec,agent_id}, enabled, greeting, counts:{conversations, pending_decisions, handoffs}}`
- `GET .../conversations?limit&cursor` returns `{items:[{customer_ref, last_message_at, preview}], cursor?, unsupported}`
- `GET .../decisions` returns `{items:[{id, verb_or_kind, summary, status, created_at}]}`
- `GET .../handoffs` returns `{items:[{contact, question, transcript_ref, created_at}]}`

These are owner reads, not the public visitor surface. The decisions, conversations, and handoffs filters are the leak surface, so `/security-review` applies.

## Tenancy and the isolation guarantee

Every read runs two workspace gates. First it loads the Site scoped to the caller's active workspace, so a cross-tenant or malformed id is a 404 with no existence leak. Then it resolves that site's paw-bar widget from `Site.pocket_id`, also workspace-scoped. From there each read binds to that one widget or pocket, never pocket-wide and never workspace-wide, so a sibling site or a second widget in the same workspace can never appear. There is an explicit test for that.

## Where decisions come from (a deliberate change from the task's suggested source)

The task suggested reading the Instinct store filtered by the `widget_id` in the proposal blob. Reading the code, that path is fragile: paw-bar stamps the Instinct row's in-row `workspace_id` with the widget owner (`decision_loop.resolve_workspace_id` returns `widget.owner`, for example `user:maya`), not the physical dashboard workspace, so a workspace-scoped Instinct read comes back empty; and the Instinct proposal's physical file depends on request context (per-workspace on the run path, shared on the ingest path).

So decisions read the paw_bar `DecisionStatus` table instead. The decision loop parks a `DecisionStatus` row alongside every Instinct proposal, keyed on the Instinct action id, so it is the 1:1 mirror of those proposals but keyed on a real `widget_id` column and served from a single store regardless of deployment mode. That turns cross-site isolation into a plain `WHERE widget_id = ?` and removes the physical-file hazard. The returned `id` is the Instinct action id so the dashboard can still deep-link into The Tray.

## Conversations feasibility answer

Listable. Concierge dispatches persist as `ChatRunDoc` with `context_type="concierge"` and `scope_id=<pocket_id>`, and the model carries a compound `(workspace, context_type, scope_id, createdAt)` index that backs an efficient per-site query. A Site is 1:1 with `(workspace, pocket_id)`, so a pocket-scoped read is site-scoped. Runs are grouped by `customer_ref` into one conversation each over a bounded scan (never a full-collection read), so `unsupported` stays `false`.

## Handoffs

`_paw_handoffs` is a reserved Fabric object type (SS-6) with no producer yet, so this defines the read shape and returns an empty list in v1. When a capture path ships it writes objects of that type carrying the widget id, and the same widget-id property filter keeps them isolated per site. The capture path is deferred and not built here.

## Tests

New `tests/cloud/test_paw_bar_admin_aggregation.py` covers: admin auth required (403 under `enforce_scope`), cross-tenant and malformed site id 404, overview shape and counts, decisions filtered to the widget with verb and kind mapping, handoffs empty-but-well-shaped plus the mapped shape when present, conversations grouped by customer, and the cross-site isolation test that a sibling site and widget in the same workspace is absent from decisions, conversations, handoffs, and the overview counts.

Run: `uv run pytest tests/cloud/test_paw_bar_admin_aggregation.py tests/cloud/test_paw_bar_backend.py -o addopts="" -q` gives 46 passed. The wider paw-bar set (admin aggregation, backend, concierge settings, decision loop) is 74 passed.

## Security review follow-up (commit e398d2d3)

Isolation held under review (no cross-tenant leak). Two hardening fixes landed on top:

- Role gate. The four reads used `require_scope("admin")`, which admits any authenticated dashboard user, so a workspace member could read another owner's visitor conversations, decisions, and handoffs. Registered a `paw_bar.read` RBAC action (ADMIN, mirroring `audit.read`) and gated all four reads on `require_action("paw_bar.read", workspace_dep=current_workspace_id)`. The role check resolves the workspace from the session, not a path or query value, so tenancy stays session-derived and matches the data scope. Owner and admin get 200, member gets 403.
- Empty pocket_id guard. `_resolve_site_and_widget` now returns no widget when `Site.pocket_id` is blank, since the store drops the pocket filter on a falsy value and would otherwise resolve a sibling site's widget.

Added tests: member 403 on all four reads, admin and owner 200, the empty-pocket_id path resolves widget=None with a sibling widget present, and a cross-workspace test where a second workspace's widget, decisions, runs, and handoffs never appear in workspace-1's reads. `test_paw_bar_admin_aggregation` is now 17 tests; the targeted run with the backend suite is 50 passed.

### DIFF
--- a/ee/pocketpaw_ee/guards/actions.py
+++ b/ee/pocketpaw_ee/guards/actions.py
@@ -3,6 +3,14 @@
 # machine-readable `code` emitted on denial. Tests iterate ACTIONS to
 # guarantee every guarded operation is covered.
 #
+# Updated: 2026-07-16 (D2 concierge dashboard reads) — added ``paw_bar.read``
+# (ADMIN) so the per-site Concierge dashboard reads (ee.pocketpaw_ee.paw_bar
+# .router: overview / conversations / decisions / handoffs) gate on the caller's
+# workspace ROLE instead of the coarse ``require_scope("admin")`` (which admits
+# any authenticated dashboard user). ADMIN, mirroring ``audit.read``: a site's
+# concierge conversations, decisions, and handoffs carry visitor PII and owner
+# decision context, so they must not be visible to every workspace member.
+#
 # Updated: 2026-07-11 (feat/external-alerting-c2c3) — registered
 # ``automations.read`` (MEMBER) and ``automations.manage`` (ADMIN) for the
 # always-on automation status surface (ee.cloud.automations_status.router): view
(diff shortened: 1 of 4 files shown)


prakashUXtech:

PR #1751, opened 2026-07-20
Derive the native-toolchain requirement from the manifest

DevRohit06:
## The problem

The runtime-requirements probe marked every project that had a `package.json` as needing a native toolchain. The reasoning in the code was "any non-trivial tree pulls prebuilt native binaries" and it named esbuild and rollup as the examples.

Those two are exactly the wrong examples. Both ship a prebuilt binary *and* a WASM fallback, and our own WebContainers gate run on 2026-07-18 proved they work in the tab: `npm install` exit 0 over 320 packages, then a running Vite dev server on the rollup toolchain. The packages cited as proof that a project needs a real VM were the ones already shown running without one.

`Capabilities.nativeToolchain` on the client means the other thing — "compile and run native code: gcc, node-gyp, native modules." So the blanket `true` claimed, of every project, a need most of them don't have.

That wasn't just imprecise. The WebContainers adapter honestly declares `nativeToolchain: false`, so the two together made that runtime **unselectable for every repo** — no env var or operator config could reach it. The registry could only ever pick Daytona.

## The fix

`nativeToolchain` is now read from the manifest:

- a table of packages that actually compile (`node-gyp` addons, `better-sqlite3`, `sharp`) or ship machine code with no WASM path (`electron`, `playwright`, `sass-embedded`)
- npm's own `"gypfile": true` marker
- `scripts` that shell out to a compiler (`node-gyp`, `cmake`, `make`, `gcc`, matched at a command boundary, not as a substring)

`optionalDependencies` is deliberately not scanned. That's where rollup and esbuild put their per-platform prebuilds, and reading it would bring the old behaviour back under a new name.

The fail-up policy is untouched: an absent, unreadable, oversized, or unparseable manifest still routes to most-capable, and there's a test pinning that so a later change can't quietly invert it.

## Review

An adversarial review of the first commit found three real gaps, all fixed in the second commit:

- **`@playwright/test`** — the package `npm init playwright` installs — wasn't in the table, so Playwright projects routed to a runtime that can't launch a browser. This was the one defect in the stranding direction. Closed with an `@playwright/` prefix, plus `sass-embedded` and `electron`.
- **`better-sqlite3`** was emitting `rawSockets: true`. It opens no socket. It was in the socket table as a leftover from when that flag was the only way to force a VM. Now that `nativeToolchain` covers it honestly, the wrong reason is gone.
- The script scanner matched substrings, so `make` fired inside `cmake` and `prebuild` fired on the npm `prebuild` lifecycle-script name. Both over-provision, which re-hid the in-tab runtime. Replaced with a boundary regex; `cmake .` now reports "runs cmake".

## Tests

`tests/cloud/test_websandbox_requirements.py` — 63 passed. New coverage for each package class, the script false-positives, the sqlite mislabel, and the fail-up guarantee. mypy and ruff clean.

### DIFF
--- a/ee/pocketpaw_ee/cloud/websandbox/requirements.py
+++ b/ee/pocketpaw_ee/cloud/websandbox/requirements.py
@@ -88,12 +88,19 @@ _MAX_MANIFEST_BYTES = 2 * 1024 * 1024
 # project that talks to Postgres or Redis directly can only run in a real VM.
 #
 # Each entry carries the CAUSE it raises the flag for, because the reason string
-# is the product here. An earlier version emitted "better-sqlite3 in dependencies
-# -> rawSockets", which is simply false — better-sqlite3 is an embedded file
-# database that opens no sockets. It still needs a real machine (it is a native
-# FFI binding an in-tab runtime cannot load), so the VERDICT was right and the
-# EXPLANATION was wrong. That is the worst combination: it survives review and
-# then sends the next debugger hunting a network problem that does not exist.
+# is the product here. This table must contain ONLY packages that open a real
+# network socket — nothing else, however much it might also need a VM.
+#
+# sqlite3 and better-sqlite3 used to live here (as "native FFI, honest reason"),
+# because before ``nativeToolchain`` was derived, rawSockets was the ONLY lever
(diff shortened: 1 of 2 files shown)


DevRohit06:

PR #1752, opened 2026-07-21
Broker the WebContainers licence key the way BrowserPod's is

DevRohit06:
Code Mode's in-tab runtime had an adapter but no key.

StackBlitz licenses WebContainer per deployment: a boot is permitted without a key on localhost, and needs a licensed key on any other origin. So the frontend adapter, which deliberately carried no key (a key in a bundle is a key that has leaked), could only ever work on a developer's machine.

The generic broker from #1750 was already the right shape. It just had one runtime registered, so this is a registration plus a module that resolves the key, and no new endpoint:

    GET /websandbox/runtimes/webcontainer/credentials

## What's here

- `webcontainer.py` resolves `WEBCONTAINER_API_KEY` and answers `available: false` when it isn't set, so the client falls through to Daytona instead of erroring.
- `vendor_keys.py` holds the resolution BrowserPod had already worked out (environment first, then `.env`, whitespace counts as unset). Every line of that reasoning is about the shape of the problem rather than about BrowserPod, and there are two callers now. Same precedence, same test seam, one implementation.
- `.env.example` documents both in-tab keys. Neither had an entry anywhere an operator would look, and "nothing visibly breaks, Code Mode just uses the VM" is not a behaviour anyone guesses from a blank line.

## One thing that reads differently for this runtime

`available: false` means "this deploy holds no licensed key", not "this runtime cannot run". Whether a keyless boot is allowed depends on the caller's origin, which the backend can't see, so the client decides. The frontend half is qbtrix/paw-enterprise#650.

## Verification

`tests/cloud/test_websandbox_webcontainer.py` covers the configured, unconfigured, blank, trimmed, `.env`-fallback and rotation paths, plus one that goes through the dispatcher by runtime id. That last one matters: an unregistered runtime and an unconfigured one both answer `available: false` on purpose, so a missing dict entry is invisible at the wire and only a test that dispatches by id can tell them apart.

317 websandbox tests pass. `ruff check` and `ruff format --check` clean.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -129,3 +129,22 @@
 #                                                # "python:3.12-slim"). Unset or "standard"
 #                                                # uses the pre-built Paw dev image with
 #                                                # Python, Node.js, GCC, Docker, and UV.
+
+# ── Code Mode in-tab runtimes (browser-side, no VM) ───────
+# Keys for runtimes that execute INSIDE the user's browser tab. Both are handed
+# to the browser per-request over /websandbox/runtimes/{id}/credentials — never
+# built into the frontend bundle — so they can be rotated and revoked here
+# without redeploying the client. An unset key is not an error: the runtime
+# reports itself unavailable and Code Mode uses the Daytona VM instead.
+#
+# The frontend must ALSO be told to use an in-tab runtime (PAW_CODE_RUNTIMES)
+# and be served cross-origin isolated (PAW_COI) — see paw-enterprise's
+# deploy/README-build.md § 2b.
+#
+# StackBlitz WebContainers. Licensed per deployment: keyless boots are permitted
+# on localhost, so local dev needs nothing here; any other origin requires this.
+# WEBCONTAINER_API_KEY=
+#
+# BrowserPod. Experimental and absent from every default runtime order — it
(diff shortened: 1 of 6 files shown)


DevRohit06:
Closing: the premise was wrong.

WebContainer isn't licensed with an API key. StackBlitz issues an OAuth **client id**, used with `auth.init({clientId, scope})` — OAuth 2.0 with PKCE, which is the flow designed for clients that cannot hold a secret. The client id is a public identifier, so there's nothing here to keep out of the bundle and nothing to broker.

This PR would have added a backend endpoint and a network round trip to hand the frontend a value that is public by design, and worse, it would have documented that value as a credential. Anyone reading it later would have drawn wrong conclusions about what else could be handled the same way.

The client id now travels the same runtime-config chain as `apiUrl` and `codeRuntimes` (`PAW_WEBCONTAINER_CLIENT_ID` → `window.__PAW_CONFIG__`), which is where operator settings belong. All of it lives in qbtrix/paw-enterprise#650.

Nothing here is salvaged. `vendor_keys.py` only existed to serve a second caller that no longer exists, and the `.env.example` entry it added documented the wrong variable. BrowserPod's key is a genuine secret and keeps using the existing broker unchanged.

PR #1753, opened 2026-07-21
Ask questions about your code, on either runtime

DevRohit06:
First slice of Code Mode's agentic layer: a stateless endpoint that answers a question about code the caller sends it.

## Why it looks like this

The endpoint this will replace (`websandbox/edit.py`) gathers context by calling `DaytonaClient.download_file` and running ripgrep inside the VM. That is the reason the in-tab WebContainer runtime ships `features.cmdK: false` — a runtime that lives in the browser has no server-side row for any of that to address.

So the seam inverts. The client reads through `CodeFileSession`, which already has all seven verbs implemented on both runtimes, and sends what it read. The server opens nothing: no `DaytonaClient` import, no row id, no `authorize_sandbox`, no path jail — there is no path to escape when you are not opening files.

Tenancy is still enforced, but for a different reason than its siblings: the route spends money, so it is workspace-scoped and license-gated for metering and abuse. There is no resource to own.

## The contract is the point

`CodeAgent — runtime-agnostic: never reaches a sandbox` forbids `codeagent` from importing `daytona` or `websandbox`. Someone who "just fetches the file server-side" here would take the feature away from the in-tab runtime again, and every test would stay green. A reviewer will not reliably catch that on a diff that reads like a helpful refactor.

Mutation-tested: adding the daytona import breaks the contract, removing it restores 32 kept.

## Two things found on the way

A safety decline arrives as a **200 with an empty body**. Reading `content[0]` first raises `IndexError`, which would have surfaced a refusal as "the model call failed" and sent someone to debug an outage that was not one. Checked before the read, with a test.

The model seam this inherits defaulted to `claude-sonnet-4-7`, **which is not a model that exists** — the Sonnet line goes 4-5, 4-6, 5. An unconfigured deployment could only ever fail on it. That is worth checking before anything else on the outstanding "no available server" report, since it may be the whole story rather than a broken proxy route. Default is now `claude-opus-4-8`, pinned by a regression test. `POCKETPAW_WEBSANDBOX_EDIT_MODEL` is still read, so one operator variable can point both paths at a working route.

## Design notes

- **Read-only by construction.** Ask mode sends no tools, so the model has no mechanism to change a file regardless of what a prompt talks it into wanting. Tested.
- **The context budget drops whole files, never clips them.** A half-truncated file reads to the model as a complete one and invites a confident answer about code that was cut off. Dropped paths are reported to the caller *and* named to the model, so it can say which file it would need instead of guessing.
- **Context rides only on the final user turn.** Re-attaching it to earlier turns would resend every file on every follow-up, and those copies go stale the moment a buffer is edited.
- **The system prompt is server-owned** — `role` accepts only `user` and `assistant`, so a caller cannot inject instructions.

## Verification

- 20 new tests, all through a `client=` DI seam — no API key, no network. This is what lets the slice merge while the live-model check is still outstanding.
- 379 passing across `websandbox` / `codeproject` / `codeconnect` / `codegit` / `codeagent`.
- `ruff check` + `ruff format --check` clean. 32 import-linter contracts kept.
- Router mounts and serves `POST /api/v1/codeagent/turn` (needs a backend restart to appear).
- Two collection errors under `tests/cloud/extraction/` are pre-existing — confirmed against a stashed clean tree.

## Not in this PR

The frontend half — the `Cmd-L` binding, the context chip, and rendering into the docked chat rail — lands separately in paw-enterprise, since the two repos cannot share a PR. The tool loop that lets the agent pull a file nobody named is the next slice, and it reuses this endpoint's shape.

`websandbox/edit.py` is deleted when Cmd-K moves onto this loop, not before.

### DIFF
--- a/ee/pocketpaw_ee/cloud/__init__.py
+++ b/ee/pocketpaw_ee/cloud/__init__.py
@@ -238,6 +238,7 @@ def mount_cloud(app: FastAPI) -> None:
     from pocketpaw_ee.cloud.billing.webhooks import router as billing_webhooks_router
     from pocketpaw_ee.cloud.chat.router import router as chat_router
     from pocketpaw_ee.cloud.chat.runs.router import router as runs_router
+    from pocketpaw_ee.cloud.codeagent.router import router as codeagent_router
     from pocketpaw_ee.cloud.codeconnect.router import router as codeconnect_router
     from pocketpaw_ee.cloud.codegit.router import router as codegit_router
     from pocketpaw_ee.cloud.codeproject.router import router as codeproject_router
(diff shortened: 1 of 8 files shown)


DevRohit06:

PR #1754, opened 2026-07-21
Let the agent ask for files it was never given

DevRohit06:
Stacked on **#1753** — review that first; this diff is only the retrieval half.

`run_turn` becomes one **step** of a turn rather than the whole turn: it either answers, or comes back with the files the model wants read.

## The loop lives on the client, and that is the design

The tools are `CodeFileSession`'s own read verbs. Only the client can execute them — it has them over a WebSocket to a Daytona VM *and* directly against the in-tab WebContainer fs. A server-side loop would have to reach into the browser to run them, which is the exact shape this module exists to avoid: a project running in a tab has no server-side row for a backend to address.

So the tool **names mirror those verbs exactly**. That is the mechanism, not a naming convention — the client dispatches by name, and a rename here breaks retrieval on one runtime and not the other. Pinned by a test.

## A mutating call can never reach the client

Only `listDir` / `readFile` / `search` are offered, and a call naming anything else is **dropped rather than forwarded**. The client executes whatever it is handed, so if a `writeFile` could ride out of here — a stale tool set, a hallucinated name — then "Ask cannot edit" would be a polite request to the browser rather than a fact. Filtering server-side is the enforcement; the client refuses again locally as belt and braces.

## Stateless still holds, and it costs something

The server keeps no record of what it asked for, so the client hands back each call's `name` and `input` beside its `output`, and `_replay_tool_exchanges` rebuilds both halves for the model.

They are emitted as **one `tool_use`/`tool_result` pair per exchange** rather than batching a round into a single assistant turn. The model reads either shape, but one pair per exchange means a client that drops or reorders a result cannot produce a turn containing an unanswered `tool_use` — which the API rejects outright.

## Running out of budget withdraws the tools

When the budget is spent the tools are **omitted from the request entirely**, not accompanied by "please stop looking now". Asking leaves the model free to call again; taking the tools away leaves it no option but to answer with what it has, which beats failing a question it could mostly answer. Results are capped server-side too — trusting a client-supplied length is no ceiling at all, and one `readFile` of a bundled asset would otherwise blow the context in a single round.

## Additive to #1753

The human conversation shape is untouched. Tool traffic rides in a separate per-question list, and earlier questions replay as their **final answers only** — replaying their tool traffic would re-send every file ever read, for the rest of the conversation, to say what the answer already says.

## Verification

- **18 new tests** in their own file (`test_codeagent_tools.py` pins the loop; `test_codeagent.py` still pins the one-shot contract, so a failure names which broke).
- **397 passing** across the Code Mode suites. `ruff check` + `format --check` clean. **32 import-linter contracts kept** — including the runtime-agnostic one, which still forbids this module from importing `daytona` or `websandbox`.

## Not yet

Still nothing live: the model route is unconfirmed and no WebContainer has booted through the real app. The both-runtimes claim is architectural here and tested through fakes; proving it is CA-2's live done-when.

### DIFF
--- a/ee/pocketpaw_ee/cloud/__init__.py
+++ b/ee/pocketpaw_ee/cloud/__init__.py
@@ -238,6 +238,7 @@ def mount_cloud(app: FastAPI) -> None:
     from pocketpaw_ee.cloud.billing.webhooks import router as billing_webhooks_router
     from pocketpaw_ee.cloud.chat.router import router as chat_router
     from pocketpaw_ee.cloud.chat.runs.router import router as runs_router
+    from pocketpaw_ee.cloud.codeagent.router import router as codeagent_router
     from pocketpaw_ee.cloud.codeconnect.router import router as codeconnect_router
     from pocketpaw_ee.cloud.codegit.router import router as codegit_router
     from pocketpaw_ee.cloud.codeproject.router import router as codeproject_router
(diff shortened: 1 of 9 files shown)


DevRohit06:

PR #1755, opened 2026-07-21
Cmd-K works in the browser tab, and the VM-bound edit path is gone

DevRohit06:
Third in the Code Mode agent stack. Stacked on #1754 — review that one first.

## The gap this closes

Cmd-K shipped disabled on WebContainers, and the reason was structural rather than an oversight. `POST /websandbox/{row_id}/edit` took a sandbox row id and read the target file off the VM's disk to build context. A WebContainer project runs in the user's tab and has no backend row, so there was nothing for that code to read — the adapter declared `features: { cmdK: false }` and that was the honest answer.

## What changed

The turn now runs under a `mode`.

- **Ask** keeps the three read verbs, unchanged.
- **Edit** gets those plus `writeFile`.

The prompt, the offered tools, and the filter that enforces them all come from one row in `_MODES`, so a mode can't be half-added — you can't end up running Edit's tools under Ask's "you are in READ-ONLY mode" instructions.

`writeFile` doesn't write. The client stages it and the user accepts or rejects it region by region, which is the review gate Cmd-K has always had. The tool description states that plainly, because a model that thinks the write landed will report it in the past tense and the user has no reason to doubt it.

## One deliberate deviation

The task text listed all four mutating verbs. `createEntry`, `deleteEntry` and `moveEntry` are **not** here.

Per-hunk review only covers one file's content. There is no review UI for "the agent deleted a file", so shipping those three would add an ungated destructive path in the same change that promises the review UX is unchanged. They land when their gate does. A test pins the current set, so adding one without a gate has to fail there first.

## The delete

`websandbox/edit.py`, its route, its DTOs and its test file are gone rather than left beside the replacement.

## Verification

```
374 passed  (websandbox / codeproject / codeagent / daytona)
 56 passed  (the three codeagent suites)
ruff check + format: clean
Contracts: 32 kept, 0 broken
```

The two collection errors under `tests/cloud/extraction/` are pre-existing — confirmed against a stashed clean tree earlier in this stack.

## What is not proven

Nothing has run against a live model or a live runtime. The permission split is enforced and tested through the DI seam; the "Cmd-K in a WebContainer" half of the done-when needs the client PR plus an actual boot, and that is CV-1's job.


DevRohit06:

PR #1756, opened 2026-07-21
Prompt to a running project: React, Vue, Svelte or Next

DevRohit06:
CS-1 + CS-2 of the scaffold track. Based on `feat/ai-agentic-code`.

A prompt becomes a running dev server. `/codescaffold/plan` reads the prompt and picks a framework; `/codescaffold/compose` returns the project as a source map; `/websandbox/{row_id}/scaffold` ships it into a VM, installs, and serves.

## Why not `git clone`

I researched this before building it, and the numbers are measured rather than assumed:

| Framework | Standalone template repo? |
|---|---|
| React | ✗ — inside `vitejs/vite`, **12.9 MB** to obtain a 20 KB template |
| Vue | ✗ — `create-vue` composes fragments; there is no flat template |
| Next | ✗ — inside `vercel/next.js`, **51 MB** |
| Svelte | `sveltejs/kit-template-default` exists, but was last pushed 2025-11-21 |

Every one of them **does** ship its templates as real files in a small versioned npm package. `create-vite` is 81 KB and carries sixteen. It's literally the bytes `npm create vite` writes, it's pinnable where a branch isn't, and it arrives as a tarball — the shape the runtime materializer already consumes.

## Integrity is verified, not assumed

A pinned *version* alone still trusts whatever the network returns, and these files get installed and executed inside a user's sandbox. Every starter carries the registry's sha512 and it's checked before a byte is read out of the archive. A mismatch is fatal — there's no safe way past one. Tarballs are cached on disk, and re-verified on read, because a cache is a place bytes can rot.

Versions are pinned to the newest release **at least seven days old**, per the workspace supply-chain rule — which is why `create-next-app` is on 15.5.20 rather than 16.2.11.

## Three things the packaging forced

Each fails silently if missed:

1. **npm strips `.gitignore`** from published tarballs, so every package smuggles it under a different alias — `_gitignore` in create-vite, a bare `gitignore` in create-next-app. Without the rename, every scaffolded project commits `node_modules` on its first commit.
2. **`create-next-app` ships no `package.json`** — its CLI generates one, resolving `next` and `react` versions at run time. Pure extraction gives you a Next project with no dependencies and no `dev` script. The catalog supplies it, visibly, and never overrides a real one if a future version starts shipping its own.
3. **Templates carry binaries** (a favicon, a hero image) that can't ride in a JSON string map. They travel base64 rather than being silently dropped — a missing favicon is a mystery to whoever hits it.

## Bring-up reports steps, not a boolean

A failed `npm install` shows npm's own error text, not a spinner. Every stage carries its exit code and output tail; successful steps carry nothing, because nobody reads a clean install log. Three details that would each have failed *silently*: `--host 0.0.0.0` (Vite binds loopback, Daytona's preview arrives from outside), `nohup … &` (`execute_command` waits for exit; a dev server doesn't), and a log file (the only record of why the server died later).

The dev port comes from the starter — Next serves on 3000, Vite on 5173. A preview pane pointed at the wrong one shows nothing with no indication why.

## What this bought

None of these starters needs `workerd`. The Cloudflare/D1 template could only ever run in a VM; a Vite or Next dev server runs in an in-tab WebContainer. The pivot un-blocks the in-tab runtime, which the old template structurally ruled out.

## What was given up

**The recipes.** `auth` and `stripe` used to arrive working; a Vite starter is a blank app, and the agent has to write those features itself (CS-5). That's the real trade. Recoverable at `d233253f` if the depth is ever wanted back.

## Verification

```
 45 passed  tests/cloud/test_codescaffold.py  (incl. 4 live npm fetches)
 26 passed  tests/cloud/test_websandbox_scaffold.py
473 passed  websandbox / codeproject / codeagent / codescaffold / daytona
ruff check + format: clean
Contracts: 33 kept, 0 broken
```

All four starters were fetched, verified and extracted for real — each yields a `package.json` with a working `dev` script and a restored `.gitignore`. The live tests are behind `PAW_TEST_LIVE_NPM=1`; they're the only thing that can catch a pinned package whose subdirectory moved, which is what this design is most exposed to.

## Not proven

**No VM has run this.** `bring_up` is driven by a fake Daytona client — the sequencing and failure reporting are proven, that `npm install` succeeds in the sandbox image is not. And there's still no UI: nothing calls these endpoints, and the landing page still refuses a prompt. That's CS-4.


DevRohit06:

PR #1757, opened 2026-07-21
ship: deploy-engine contract and Dokku driver

prakashUXtech:
First slice of `/ship`, the managed-deploy surface. Dokku is the v1 engine, driven over SSH, but the load-bearing artifact here is the **contract** — a provider-agnostic port with nine verbs. Swapping Dokku for Dokploy or a Go engine later should be a driver replacement, not a product rewrite, and the contract test suite is what makes that true.

## What's here

- **`ship_engine/port.py`** — the `ShipEngine` protocol: `provision_box`, `deploy_app`, `add_domain`, `db_create`, `backup`, `rollback`, `logs`, `metrics`, `destroy`. Frozen DTOs per verb, typed errors (`VerbNotSupported`, `CommandFailed` carrying exit code and a redacted stderr tail). No result DTO carries secrets.
- **`ship_engine/dokku.py`** — `DokkuDriver` over asyncssh, implementing the eight box-level verbs. Every SSH exec funnels through one `_run` chokepoint with per-verb timeouts and redaction applied before anything is logged or raised. `provision_box` raises `VerbNotSupported`; provisioning belongs to the provider layer.
- **`ship_engine/transcripts/`** — 18 recorded Dokku CLI fixtures plus a replay transport, so the driver is fully exercisable with no network and no box.

## Testing

32 tests. The contract suite is parametrized over an engine registry rather than hardcoded to Dokku — a future driver registers one case and runs the same assertions unchanged. Tests assert on parsed content (app URLs, byte sizes, disk percentages, exact command sequences), not just "no exception".

`uv run pytest tests/ship_engine/` → 32 passed. Re-ran with sockets blocked at interpreter level to prove the no-network claim: still 32 passed. ruff, ruff format, and mypy clean on the new modules.

## Security notes

The first review pass caught a real one: env var **values** were shell-quoted but **keys** were interpolated raw, so a hostile key could have executed arbitrary commands as root on the box. Fixed by quoting the whole pair, plus key validation at the DTO boundary (POSIX name shape) so bad keys fail before reaching a driver. Both paths have tests.

Two related fixes: `AsyncSSHTransport` no longer passes `known_hosts=None` to asyncssh (an explicit None *disables* host-key verification — the kwarg is now omitted unless configured, so asyncssh's default applies), and `AppSpec.env` is `repr=False` so config secrets can't leak through a dataclass repr.

## Choices worth flagging

Rollback is a pinned-image redeploy rather than a Dokku release rollback, and backup is `mongo:export` to box-local disk — both documented at the call site, both v1 scoping. `asyncssh` is lazy-imported so the contract and transcript paths never load the SSH stack. The lock refresh pulled cryptography 49.0.0 and typing-extensions 26.3.0 along with it; every new pin is 24+ days old, well past the 7-day minimum-release-age policy.

`AsyncSSHTransport` itself is only covered at the import level — the real SSH path gets exercised in the dogfood run later in this arc.


prakashUXtech:

prakashUXtech:

PR #1758, opened 2026-07-21
Let a scaffolded project reach the in-tab runtime

DevRohit06:
Stacked on `feat/codescaffold` (#1756).

CS-3 started as client work and turned up a contradiction in the backend it depends on. Three statements, all in the tree:

| Where | Claim |
|---|---|
| `codescaffold/domain.py` | every starter needs `nativeToolchain: True` |
| a comment beside it | "the WebContainers adapter honestly declares it can run those" |
| `adapters/webcontainer.ts` | `nativeToolchain: false` |

The adapter is right — WebContainer executes WASM and has no x86 loader. The arithmetic of the other two is that **no scaffolded project could ever select the in-tab runtime**. The pivot's headline benefit was unreachable, silently, and it had been asserted in a comment rather than checked.

## What changed

`needs_native` is per-starter now, and the flag means what the capability says: does this project need to run native code.

Which exposed the real blocker. `create-vite@9.1.1` pins **Vite 8**, and Vite 8 replaced rollup+esbuild with **rolldown** (a hard dependency) plus lightningcss — both native Rust. Rolldown's wasi fallback crashes in a WebContainer, on this exact React template:

> `worker (tid = 43) sent an error! Invalid atomic access index` — raised from emnapi
> — stackblitz/webcontainer-core#2105, and #2107 for the same thing

lightningcss is worse: its optional deps list platform binaries and no wasm build at all.

Our own gate run had already proved both halves of this and it was sitting in the adapter's capability comment the whole time. Run 1, on rolldown, failed at `npm run dev`. Run 2, Vite on **rollup**, passed end to end — boot 6.5s, `npm install` exit 0 across 320 packages, running dev server, live preview.

So the three Vite starters are pinned to **create-vite@8.3.0**, which pins Vite ^7.3.1. The templates are structurally identical between the two majors — same `template-*-ts` subdirs, same `_` dotfile prefix — so this costs one minor version of Vite and buys the entire in-tab path. There's a test that fails if someone bumps it back, because that's invisible from the version string alone.

Next stays on `nativeToolchain: true`. That one is a claim held in the safe direction, not a settled question: Next 15's SWC has a wasm fallback it selects itself, but Tailwind v4's oxide engine publishes none, and nobody here has watched either resolve inside a container. Requirements fail up when we don't know. React/Vue/Svelte open in a tab, Next gets a VM, and the reason is a string the user can read.

## Also: `GET /codescaffold/starters`

The boot path has a project row and no prompt, so it can't ask `/plan` what a project needs. The landing page wants the four choices before anyone types. Neither can use the existing routes.

It omits `package`, `integrity` and `subdir` deliberately — those are how we fetch a starter, not what it is, and a client that had them could fetch a template itself, routing around the sha512 check that makes the pin worth having. There's a test asserting none of them reach the wire.

## Verified

51 tests in `test_codescaffold.py`, 4 of which really download from npm (`PAW_TEST_LIVE_NPM=1`). 109 across the Code Mode suites. ruff clean. 33 contracts kept, 0 broken. All four starters fetched, integrity-verified and extracted at the new pin — each yields a `package.json` with a working `dev` script, a restored `.gitignore`, and no rolldown.

## Docs

None to update: `/codescaffold` is an ee/cloud route and `docs/api-reference.md` documents neither it nor `/websandbox`. The C4 model has no entry for `codescaffold`, `codeagent` or `codeenv` yet — that's CV-1's touch-time pass, tracked in the tasks doc.


DevRohit06:

PR #1759, opened 2026-07-22
ship: Hetzner provisioning and box lifecycle

prakashUXtech:
> Stacked on #1757 (the engine contract). Review that one first; this PR's diff is only the provisioning layer.

Second slice of `/ship`. #1757 gave us a driver that can talk to a Dokku box; this one **creates the box**. Hetzner via the official SDK, cloud-init installs a pinned Dokku plus nixpacks, and a `ShipBox` document tracks the lifecycle.

## What's here

- **`ShipBox` model** — one row per box: provider, server id, IP, status (`provisioning` → `ready` | `degraded` | `destroyed`), captured monthly price, and the box's SSH private key encrypted at rest. Registered with Beanie; only the ship service and the provisioning job touch the document, matching the one-service-owns-one-doc boundary the credit and litellm-key entities use.
- **`ship_engine/hcloud.py`** — creates the server behind an injectable client seam, uploads the provisioning key, ensures a firewall (22/80/443 in), and captures the server type's monthly price at create time.
- **`ship_engine/cloudinit.py`** — the user-data that installs Docker, a pinned Dokku, and nixpacks, and authorizes our key for both root and the `dokku` user. Versions are pinned so the control surface stays reproducible and #1757's transcript assumptions hold.
- **`ship_engine/keygen.py`** — a fresh ed25519 keypair per box, so one box's credential is one box's blast radius.
- **`cloud/ship/`** — the store seam (owns the document, hides the encryption envelope), the lifecycle orchestrator, the arq job, and the web-side enqueue.

## How provisioning behaves

Poll-and-probe, not phone-home: create the server, then poll until it answers `dokku version` over SSH. Bounded attempts, so a box that never comes up lands in `degraded` instead of hanging.

**The retry story matters more than the happy path.** The server id is persisted the moment the create returns, before readiness. A retry sees it and skips the create entirely — so a job that dies mid-provision never orphans a second paid server. There's a test that asserts the provisioner is called zero times on the second run.

Provider failures never escape as raw payloads; they're mapped to a fixed safe string before landing in `status_reason`, since provider errors can carry token fragments.

## Testing

22 tests, no network and no real box — the hcloud client and the readiness probe are both injected fakes and `sleep` is a no-op, so the poll loop runs instantly.

Covered: provisioning → ready, provider failure → degraded, readiness exhaustion → degraded (with the server id retained), a connection error mid-boot tolerated then ready, the retry-never-double-creates guarantee, private key encrypted at rest and absent from the serialized document, and cross-tenant reads returning nothing.

- `uv run pytest tests/cloud/ship/` → 22 passed
- `uv run pytest tests/cloud/chat/ tests/ship_engine/` → 143 passed (the worker and engine consumers of this change)
- `lint-imports` on both the core and `ee/pyproject.toml` configs → 32 contracts kept
- ruff and ruff format clean

## Notes for review

The SSH private key is stored Fernet-encrypted on the box document rather than in the connector state store — that store is keyed on connector name and owned by the connector lifecycle, while this credential is 1:1 with the box and dies with it. Same deployment key (`CLOUD_ENCRYPTION_KEY`) the meetings and connector secrets already use.

`hcloud` 2.23.0 is the one new dependency, aged well past the 7-day minimum-release-age policy. The provisioning job rides the existing worker with its own timeout, since a box boot plus Dokku install runs longer than a chat run.

Needs `POCKETPAW_HCLOUD_TOKEN` and `CLOUD_ENCRYPTION_KEY` set to provision for real; the live run happens in the dogfood slice later in this arc.

### DIFF
--- a/ee/pocketpaw_ee/cloud/chat/runs/worker.py
+++ b/ee/pocketpaw_ee/cloud/chat/runs/worker.py
@@ -71,6 +71,7 @@ from pocketpaw_ee.cloud.jobs.domain import job_timeout_seconds
 from pocketpaw_ee.cloud.jobs.worker import execute_workspace_job
 from pocketpaw_ee.cloud.metering.sweeper import sweep_unbilled_runs
 from pocketpaw_ee.cloud.shared.db import close_cloud_db, init_cloud_db
+from pocketpaw_ee.cloud.ship.job import provision_box_job
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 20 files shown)


prakashUXtech:

prakashUXtech:

PR #1760, opened 2026-07-22
ship: REST surface, app and deploy entities, deploy pipeline

prakashUXtech:
> Stacked on #1759 (provisioning), which is stacked on #1757 (engine contract). Review those first; this PR's diff is only the HTTP surface.

Third slice of `/ship`. #1757 gave us a driver, #1759 gave us boxes — this one puts a workspace-scoped REST API in front of both, so the UI in paw-enterprise#655 has a backend to talk to.

## What's here

- **`cloud/ship/{domain,dto,service,router}.py`** on the ee/cloud 4-file shape, reusing the SHIP-2 modules rather than duplicating them. `store.py` stays the only module that touches the Beanie documents.
- **Two new documents** — `ShipApp` (name, box, build path, image/git ref, env var *names*, prod flag, URLs, routed domains) and `ShipDeploy` (the per-attempt log the deploy job advances).
- **The routes**: create and list boxes, box metrics, create and list apps, deploy, deploy history, add and list domains, create a database, read logs, plus the two DELETEs.
- **`engine.py`** — a session seam that decrypts the box key to a `0600` temp file, hands a `DokkuDriver` to the caller, and shreds the key in a `finally`. Every engine call in the slice goes through it.
- **Deploy pipeline** — an arq job on the shared worker with its own timeout (pulling an image and swapping containers outlasts a chat run), persisting `queued → building → releasing → live | failed` and emitting on transitions.

## Destroy does nothing yet, by design

Both DELETEs park a proposal and return `{status: "pending_approval", proposal_id}`. The engine's `destroy` verb is never called from the request path — there's a `# SHIP-4:` marker at each seam where the real Instinct proposal gets wired, and only the approve path will ever be allowed to execute. Repeat DELETEs reuse the existing proposal id instead of minting a second one.

## Testing

76 tests, no network — the engine is faked through the SHIP-1 transcript harness.

**Tenancy is the headline.** A parametrized sweep drives every id-bearing route as a second workspace and asserts `404` specifically — not 200, not 500. A cross-tenant deploy job id is a no-op, and a malformed id reads as "not found" rather than throwing out of bson.

Also covered: the frozen DTO shapes the UI consumes, deploy status transitions end to end, parked destroys leaving the entity intact, and an unreachable box answering `409` with a failed attempt rather than a `500`.

- `uv run pytest tests/cloud/ship/` → 76 passed
- `uv run pytest tests/cloud/chat/ tests/ship_engine/` → 143 passed
- `lint-imports` on both configs → 32 contracts kept
- ruff and ruff format clean

## Security notes

Every value that reaches a shell argument is constrained at the DTO boundary — app name, domain, and now image and git ref, which were free strings until review caught the asymmetry with their neighbours. The driver shell-quotes all of them regardless, so this is defence in depth rather than the only barrier; there are tests for both the hostile inputs and the legitimate ones (`ghcr.io/owner/repo@sha256:...` still works).

The API never accepts env *values* — only variable names. A body carrying `["API_KEY=hunter2"]` is refused rather than persisted, since that field is never treated as a secret.

## Notes for review

The import-linter contract for this entity names `store.py` as the doc owner rather than `service.py`, because the Fernet envelope on the SSH key lives there. The contract comment is explicit that `service.py`, `engine.py`, and the job modules stay off the documents by review rather than by contract — they annotate with the doc types under `TYPE_CHECKING`, which import-linter counts as a real import.

Box defaults are `cx22` in `fsn1` when a create body omits them.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -72,6 +72,12 @@ workspace's source→Fabric mappings, now with a "connector" source_kind that
 dispatches through the OSS FABRIC_INGESTORS registry — gcalendar first) and
 POST /fabric/ingest/run (run one mapping immediately; misconfiguration reports
 status="error" in the body, never a 5xx).
+
+Updated: 2026-07-22 (SHIP-3, feat/ship-3-cloud-entity) — documented the
+Ship — Managed Deploys section: the workspace-scoped /ship surface for
+provisioning a box, registering and deploying an app, routing a domain,
+creating a linked database, and reading logs + box health. The two DELETE
+routes PARK a teardown for human approval and never destroy anything.
 -->
 
 # Cloud REST API Reference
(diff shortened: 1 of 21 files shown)


prakashUXtech:

prakashUXtech:

PR #1765, opened 2026-07-22
ship: the Instinct gate and the agent verb surface

prakashUXtech:
> Stacked on #1760 → #1759 → #1757. Review those first; this PR's diff is the agent surface and the approval gate.

Fourth slice of `/ship`, and the one that decides whether any of this is safe to hand an agent. Until now the DELETE routes parked a placeholder id and a comment promised a real gate. This wires it.

`/ship` becomes the **fifth gated Instinct proposal kind**, alongside pocket writes, Belt code changes, external actions, and admin actions. Same blob contract, same propose/execute split, same discipline — deliberately, because the approve-path dispatch treats every kind uniformly and the existing ones have already paid for their lessons in production.

## The gate

`destroy` (box or app) and a **production** deploy never execute from a tool call or a request. They file an Instinct Action carrying a `_ship_action` blob and return a proposal id. `ship/executor.py` is the only code in the subsystem that may reach the engine's destroy verb, and it runs a six-step guard before it does:

1. **Schema check** — a stale blob approved after a deploy fails loud rather than running misinterpreted.
2. **Params hash re-check** — a human approved a *specific* teardown; an edit between propose and approve is refused.
3. **Idempotency** — a blob carrying an outcome, or an already-terminal Action, never re-fires.
4. **RBAC re-check at execute time** — the proposer must still hold `ship.manage` *right now*, resolved off a freshly loaded user doc. Demoted or removed while the action sat in the tray → fails closed.
5. **Run** through the box session seam, which resolves the SSH credential fresh. No secret ever touches the blob.
6. **Outcome, terminal, chain close** — exactly once.

Concurrency is the part worth reading closely. `admin_proposals` documented a real defect where two concurrent invocations on one approved action both passed the guard and double-fired. This inherits the fix: a per-action lock around the read-check-write window, with the outcome written inside it so a racing second caller's idempotency check sees it.

The executor never raises. An engine failure becomes a `failed` outcome; even a programming error inside the verb runner is caught, because breaking the approve response would be worse than the failure itself.

## The agent surface

`pocketpaw_ship`, an in-process MCP server with ten tools — list/provision boxes, create/deploy apps, add domains, attach databases, read logs and metrics, and request a teardown. Every tool calls the service layer, never the driver or the store.

The agent's surface is **narrower than the HTTP one**, on purpose:

| Verb | Operator over HTTP | Agent over MCP |
|------|--------------------|----------------|
| reads, provision, create app, domain, db | runs | runs |
| deploy to a non-prod app | runs | runs |
| deploy to a **prod-flagged** app | runs | **proposes** |
| destroy a box or an app | proposes | proposes |

An operator calling the API with their own credentials isn't the same actor as an agent acting on their behalf. Binding the new `ship` connector to a pocket auto-surfaces a bundled `ship` skill into that room, which teaches the verb surface and one rule above all: never report a teardown as done when the tool said `proposed`.

The connector declares no HTTP actions — it exists to hold the per-workspace Hetzner token (BYO-credential, same posture as the meetings connectors) and to carry the skill contribution.

## Testing

94 tests in `tests/cloud/ship/`, 17 of them new and aimed squarely at the gate:

- A destroy request runs **no engine command** — the engine fake raises on any command outside its recorded map and `apps:destroy` isn't in it, so reaching the engine fails the test instead of passing quietly.
- Two concurrent approvals execute the verb **exactly once**.
- Re-approval and terminal actions don't re-fire.
- An edited params blob, a stale schema, and a missing workspace are each refused.
- A demoted proposer fails closed.
- The never-raises contract holds even when the verb runner throws.

`uv run pytest tests/cloud/ship/` → 94 passed. `lint-imports` on both configs → 32 contracts kept. ruff clean.

I also ran the full `tests/ee/` suite before and after this branch: 61 failures, byte-identical lists. They're pre-existing on `dev` (sites, foresight, and friends) — none are from this change.

## Notes for review

`ship.read` (MEMBER) and `ship.manage` (ADMIN) are new RBAC actions. ADMIN mirrors `connector.manage`: irreversible work on shared infrastructure, but not the owner-only tier that governs money.

A prod deploy proposal reuses the destroy proposal *view* — the wire shape is identical — but deliberately does **not** emit `ShipDestroyProposed`, which would tell every listener a teardown is pending when nothing is being torn down. A dedicated event lands with the console work that renders it.

Two shared test fixtures moved from the router suite into `conftest.py` so both suites get them by discovery rather than importing fixtures across modules.


prakashUXtech:

prakashUXtech:

PR #1766, opened 2026-07-22
ship: /ship v1 backend — engine, provisioning, REST surface, Instinct gate

prakashUXtech:
Rolls up the four `/ship` backend slices into one branch so the surface can be tested as a whole instead of four PRs deep in a stack. Same commits, no rebase — `dev` hasn't moved since we branched, so this tip already contained the full chain.

Supersedes #1757, #1759, #1760, #1765 for merge purposes. Those stay open for per-slice review; merging this one lands all four.

## What lands

| Slice | What it adds |
|-------|--------------|
| SHIP-1 | `ShipEngine` port (9 verbs) + `DokkuDriver` over SSH + 18 recorded transcripts |
| SHIP-2 | Hetzner provisioning, cloud-init, per-box ed25519 keys, `ShipBox` lifecycle |
| SHIP-3 | `/api/v1/ship` REST surface, `ShipApp`/`ShipDeploy`, the background deploy pipeline |
| SHIP-4 | The Instinct gate (`_ship_action`), the `pocketpaw_ship` MCP surface, connector + skill |

73 files, ~7.9k lines. The frontend half is paw-enterprise#655, which consumes these DTO shapes verbatim and merges independently.

## Verification on this branch

- `uv run pytest tests/cloud/ship/ tests/ship_engine/` → **126 passed**
- `lint-imports` on both configs → **33 contracts kept**
- ruff clean across the new modules
- Full `tests/ee/` diffed before/after the stack: 61 failures, byte-identical lists — all pre-existing on `dev`, none from this work

Assembly checks that unit tests don't cover:

- the arq worker registers `provision_box_job` and `deploy_app_job` alongside the existing two
- all 10 `/ship` route paths resolve, mounted at `/api/v1`

## Testing it independently

Everything above runs box-free — the engine is faked through recorded Dokku transcripts, so CI needs no server and no Hetzner account.

To exercise it for real you need two env vars:

- `CLOUD_ENCRYPTION_KEY` — Fernet key for the at-rest SSH key. Generate: `python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"`
- `POCKETPAW_HCLOUD_TOKEN` — a Hetzner API token, per workspace

Then: `POST /api/v1/ship/boxes` → poll `GET /api/v1/ship/boxes` until `ready` (the box boots and installs Dokku via cloud-init) → create an app with an image → deploy → add a domain. Provisioning a `cx22` in `fsn1` bills monthly, so tear it down when you're done — which itself files an approval rather than destroying anything.

The scripted end-to-end version of that walk is SHIP-6, still to come.

## What's deliberately not here

- **SHIP-6** — the dogfood recipe that deploys the Paw-OS stack itself, plus the live run. Needs a Hetzner token.
- **SHIP-7** — the cost recorder and the flag-gated wallet hook.
- Provider teardown on `destroy_box`: the approved verb removes the box's apps and marks the row `destroyed`, but releasing the server at Hetzner lands with SHIP-6's provider work. Worth knowing before you approve a teardown and expect the bill to stop.

## Merge note

If you'd rather keep per-slice history, merge the four stacked PRs bottom-up with `--rebase` instead and close this. If you just want it on `dev`, merge this one and close the rest.


prakashUXtech:

PR #1769, opened 2026-07-22
Asked for an app on /code, got a pocket

DevRohit06:
With a React project open on /code, "Let's build an employee management app, with components, nice design etc" made the agent create a pocket and start authoring a ripple ui-spec. It answered a request for an app with a dashboard.

## Why the existing guards missed it

CD-3 had already scoped this surface: ripple off, filesystem built-ins denied, MCP surface narrowed to `code_mode`. The preamble says "do not create a pocket" in plain words.

The scope could not hold. `claude_sdk` unions three families back in *after* a mode's allow-list is applied: `POCKET_CREATION_GRANT`, `WIDGET_TOOL_IDS`, and every tool on an `ALWAYS_ALLOWED_MCP_SERVERS` server. That is deliberate, and right for general chat, where creating a pocket is the core capability and has to work from anywhere. On /code it hands the agent the exact tools the surface exists to avoid. Dumping the allowlist the reported session ran with: `pocket_specialist__create`, `plan_pocket`, `add_widget`, `update_widget` and all three widget tools were sitting there.

The prose lost for a separate reason. Every noun in that request (app, components, design) matches the bundled create-pocket skill's description, and the skill loads as a local plugin, so `skill_names` never withheld it.

## What changed

Deny, because deny runs before the grant is unioned in and is the only lever that reaches those families. The pocket, planner and widget tools go, along with the read verbs and `Skill` itself.

/code also gets its own system prompt, which wires `SurfaceProfile.system_message_override` (declared back in June, never consumed). Denying the tools alone would have left the agent holding a pocket-shaped behavioral stack it could no longer act on: the artifact rule wants `Write` and `deliver_artifact`, the delegation rule wants the specialist. CD-3 already wrote down what that costs, the agent tries, takes hard errors, burns turns.

The bigger problem is that forbidding something does not replace it. Told only what *not* to build, the agent still has the dashboard as its one concrete plan. The new prompt says what the surface is instead: the project is not on this machine, `code_mode` is the only door, "components" means React components, and nothing counts as done until the tool confirms it.

An override swaps the deliverable stack and nothing else. The runtime identity rule and the Composio rules stay, since they describe the environment rather than the work, and the Composio ones are already gated on Composio being enabled. /code's system prompt drops from 42,781 chars to 7,186.

## Effective allowlist

Before:

```
code_mode, pocket_specialist__create, pocket_specialist__edit,
pocket_planner__plan_pocket, pocket__add_widget, pocket__update_widget,
pocket__get_pocket, pocket__list_pockets, widgets__get_widget_spec,
widgets__get_inline_widget_help, widgets__start_flow,
atlas__*, composio, Skill, WebSearch, WebFetch
```

After:

```
code_mode, atlas__*, composio, WebSearch, WebFetch
```

## Testing

The repro drives the real `_build_options` rather than asserting on the profile, because the bypasses that caused this live in the computation. A membership check on `deny_mcp_tool_ids` goes green while the agent still holds the tools, which is exactly how the original slipped through CD-3's suite.

A second test pins the deny literals against the real `WIDGET_TOOL_IDS` and `POCKET_CREATION_GRANT`, so a tool added to either family later reopens the hole loudly instead of silently.

`build_behavior_instructions` runs on every cloud chat turn, so there is a regression guard on the untouched path too, covering both a default ripple-on profile and the legacy no-profile case.

I enumerated every test file that touches the changed code paths (`build_behavior_instructions`, `resolve_profile`, `SurfaceProfile`, the deny/allow sets, `_build_options`, `surface_registry`) and ran that set against the clean base commit and against this branch:

| | Files | Result |
|---|---|---|
| Base `a08876ce` | 32 | 2 failed, 399 passed |
| This branch | 33 | 2 failed, 409 passed |

Same two failures either side (`test_run_core.py::test_execute_run_cancellation_preserves_original_exception` and `::test_execute_run_threads_surface_meta_into_ctx`), so both predate this work. The extra 10 passing are the new tests. `tests/cloud/workspace` and `test_auth_cookie_chain.py` were compared the same way and are also identical (2 failed, 189 passed on both), those failures being environmental: LiteLLM 503, mongomock, fastapi cookie fixtures.

## Note

One CD-3 assertion was inverted on purpose. It expected `Skill` to survive on this surface, on the same "it does not reach a filesystem" reasoning that keeps WebSearch. That reasoning holds for the filesystem question and misses this one.

Two lint errors remain in `ee/pocketpaw_ee/agent/mcp_servers/code.py` (E501). They predate this branch and I left them alone to keep the diff on topic.

### DIFF
--- a/ee/pocketpaw_ee/cloud/chat/agent_service.py
+++ b/ee/pocketpaw_ee/cloud/chat/agent_service.py
@@ -1353,13 +1353,42 @@ def build_behavior_instructions(ctx: ScopeContext, *, backend_name: str | None =
     # ``surface_profile.ripple_mode="off"``) suppresses the ripple block.
     ripple_off = ctx.resolved_profile is not None and ctx.resolved_profile.ripple_mode == "off"
 
+    # A surface may replace the DELIVERABLE stack wholesale with its own system
+    # prompt. ``None`` (every surface but /code today, and the legacy
+    # ``resolved_profile is None`` path) keeps the assembly below untouched.
+    #
+    # What an override replaces: the ripple LAW, the pocket-delegation rule, the
+    # per-backend pocket prompts, and the artifact-delivery rule — everything
+    # that tells the agent what to BUILD. On a surface with a different
+    # deliverable each of those is an instruction to use a tool the profile has
+    # denied, and CD-3 already recorded what that costs ("the agent attempts
+    # them, takes hard errors, and burns turns"). ``ripple_mode="off"`` only ever
+    # covered the first two; /code showed the rest still landing.
+    #
+    # What it does NOT replace, and why the two survivors are not an oversight:
+    # ``_RUNTIME_IDENTITY_RULE`` is true on every surface (you are PocketPaw in a
+    # GUI chat; slash commands do not exist), and the Composio rules are gated on
+    # Composio actually being enabled, so prompt and tool list agree by
+    # construction. Both describe the ENVIRONMENT the agent is in. The override
+    # describes the WORK. Folding the environment rules into each surface's
+    # prompt would duplicate them per surface and let them drift.
+    override = ctx.resolved_profile.system_message_override if ctx.resolved_profile else None
     parts: list[str] = []
     parts.append(_RUNTIME_IDENTITY_RULE)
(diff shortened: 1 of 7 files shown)


DevRohit06:

PR #1770, opened 2026-07-22
Sites import: zip upload, from-url queue, and native form capture

prakashUXtech:
Adds the control-plane path for bringing an existing website into Paw Sites: a zip import that deploys through the html engine, a from-url endpoint that queues a crawl, and a native form-capture route so imported sites' contact forms keep working after the move.

## Endpoints

**`POST /api/v1/sites/import`** — multipart zip upload (25MB cap). Unpacks fully in memory with a zip-slip guard (absolute paths, `..` traversal, backslashes all rejected), an entry-count cap (2000) and an incremental uncompressed-size cap (50MB, so a decompression bomb is rejected before it inflates). Text and binary files are split by content sniff: text rides the html engine's `source` map, binaries ride a base64 `assets` sideband. The flow mints the html pocket, mints the draft Site doc (same draft-first pattern as create), publishes through the existing html/static deploy path, and persists an `import_report` on the Site doc — `{pages, asset_count, asset_bytes, forms, scripts, warnings}` — which now also surfaces on `SiteResponse`.

**`POST /api/v1/sites/import/from-url`** — `{url}` → `202 {site_id, pocket_id, status: "queued"}`. Validates URL shape only, mints the draft site with a crawler-pending report, and stops. The crawler is the next stacked slice; `import_service.crawl_site_from_url` is its seam and raises `NotImplementedError` with a note that it must add SSRF guards before fetching anything.

**`POST /api/v1/capture/form`** — the native-form capture sibling. Imported sites' forms get rewired to a plain urlencoded POST here with hidden fields `paw_site_id`, `paw_key`, `paw_page`, `paw_redirect` (plus optional `paw_form_type`, default `lead`). Same hardening ladder as the JSON capture: site exists → origin pin → constant-time key compare → payload size cap → the shared capture pipeline (honeypot, rate limit, injection screen, mapping). Responds 303 to the validated Origin plus the redirect path.

Both import endpoints are tenant-scoped writes (`fabric.write`) under the router's sites plan gate, matching every sibling mutation. Cross-tenant reads stay 404/empty.

## Cross-repo seams (heads up for the paw-sites side)

1. **Generator `assets` key.** The html build payload now carries `input.assets` (`{path: base64}`) when an import has binaries — only then, so existing payloads are byte-identical. The paw-sites slice adding decode-and-write for that key is in flight; a generator that predates it deploys the text tree only, and the import report carries a warning saying so.
2. **Capture form path.** The spec said `{captureApiBase}/v1/capture/form`, but `captureApiBase` already ends in `/api/v1`, which would have doubled the segment. Final contract as built: **`{captureApiBase}/capture/form`**, i.e. `/api/v1/capture/form`. The paw-sites form rewiring must target that.

## Open-redirect guard

`paw_redirect` must be a relative path: no absolute URLs, no protocol-relative `//host`, no backslashes, no embedded scheme, no CR/LF (header injection), length-capped. The 303 Location is built as validated-Origin + relative path, so the browser can only land back on the pinned site.

## Security self-review notes

Ran the upload/capture surfaces against the security checklist. Findings worth stating:

- From-url does no fetching at all in this slice, so SSRF is structurally impossible here; the stub documents the guards the crawler slice must add (private ranges, loopback, metadata IPs, redirect and size caps).
- Binary assets are not persisted on the pocket (Mongo doc-size risk at 25MB+ base64), so a later re-publish from the builder loses them until the pocket asset sideband lands. The import report warns about this; it is not silent.
- A dropped form submission (honeypot/rate/injection) still 303s — visitors never see which heuristic dropped them.
- `paw_*` control fields are stripped before the payload reaches the lead pipeline; file parts are skipped entirely.

## How tested

`uv run pytest tests/ee/sites/ tests/cloud/leads/ tests/sites_capture/` — 468 passed, 2 skipped (both pre-existing env-gated integration skips). New coverage: zip happy path (generator handed engine=html + source + assets, report persisted and readable), single-root flattening, zip-slip and absolute-path rejection, entry/uncompressed/upload caps, missing index.html, not-a-zip, cross-tenant invisibility, from-url 202 + queued report + invalid shapes, crawler seam, assets-sideband payload contract (present when passed, absent otherwise), and the form capture matrix (valid key records + 303, bad key 401, wrong origin 403, unknown site 404, four flavors of bad redirect 400).

### DIFF
--- a/ee/pocketpaw_ee/cloud/leads/router.py
+++ b/ee/pocketpaw_ee/cloud/leads/router.py
@@ -22,6 +22,20 @@
 # ``rate_key``. The caller-controlled ``body.submitter_ref`` is no longer the
 # limiter key (it was randomizable to dodge the cap); it rides through only as an
 # opaque, non-PII label on the stored Lead.
+#
+# Updated 2026-07-22 (SI-4 — feat/sites-import-endpoint): added the NATIVE-FORM
+# capture sibling, POST /capture/form (final URL: {captureApiBase}/capture/form,
+# i.e. /api/v1/capture/form — captureApiBase already carries /api/v1, so the
+# spec's "/v1/capture/form" would have doubled the segment; this is the
+# router-consistent resolution and the cross-repo contract paw-sites' import
+# rewiring must target). IMPORTED sites rewire their <form>s to a plain
+# application/x-www-form-urlencoded POST here, with hidden fields ``paw_site_id``,
+# ``paw_key`` (the per-site signed key), ``paw_page``, ``paw_redirect`` and
+# optionally ``paw_form_type``. The endpoint runs the SAME hardening ladder as the
+# JSON capture (site exists → origin pin → constant-time signed key → payload size
+# cap → leads_service.capture), then 303-redirects to ``paw_redirect`` — which MUST
+# be a relative path (open-redirect guard: absolute / protocol-relative /
+# backslash / CR-LF all 400), resolved against the validated request Origin.
 
 from __future__ import annotations
 
@@ -29,7 +43,7 @@ import hashlib
 import json
 import secrets
 
-from fastapi import APIRouter, Depends, HTTPException, Query, Request
+from fastapi import APIRouter, Depends, HTTPException, Query, Request, Response
 
 from pocketpaw.sites_capture.ingest import origin_allowed
 from pocketpaw.sites_capture.models import MAX_PAYLOAD_BYTES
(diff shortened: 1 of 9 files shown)


prakashUXtech:
Security-review follow-ups landed (see the new commit):

- Crafted or abnormal zip members now 422 per the endpoint's own contract instead of 500ing: bad CRC and lying size headers (BadZipFile), password-protected members (RuntimeError — an ordinary user mistake), and exotic compression methods (NotImplementedError) are all caught at the per-entry read and mapped to sites.import_zip_invalid. Regression test per case, including flipping the encryption bit in both the local header and the central directory.
- Entry names with control characters (NUL, newline, DEL) are rejected as unsafe — they end up as to-be-written file paths on the generator side of the seam.
- The upload-cap docstring no longer claims the body is never buffered past cap+1: Starlette spools the full multipart body before the handler runs, so raw-ingress bounding belongs to the fronting proxy's body limit.

Full targeted run after: 34 passed. Two review notes that ride elsewhere: the import_report fields are attacker-controlled and must render escaped (the paw-enterprise report panel was verified to do exactly that in its review), and the origin scheme tighten on the capture 303 is optional follow-up hardening.

prakashUXtech:
Landed on dev via #1771 — the endpoint slice and the crawler were squash-merged as one control-plane commit. This PR's commits are in dev; closing to avoid a duplicate merge.

PR #1771, opened 2026-07-23
Same-site crawler for import-from-url (SI-5)

prakashUXtech:
Stacked on #1770 (SI-4, `feat/sites-import-endpoint`) — review that one first. This slice fills the `crawl_site_from_url` seam SI-4 left open, so `POST /sites/import/from-url` actually imports a site now instead of parking it in "queued" forever.

## What it does

New module `ee/pocketpaw_ee/sites/url_crawler.py`: a same-site BFS crawler with an SSRF-hardened fetcher. `import_from_url` still returns 202 queued immediately, then runs the crawl as a detached background task (the same scheduler pattern the native-artifact pre-warm uses) with a 120s wall-clock backstop. The harvest goes through the exact pipeline the zip path built: text/binary split, publish with the base64 assets sideband, `import_report` persisted on the Site doc — now with crawl stats (pages, assets, bytes, robots skips, cross-origin ref count). The harvested source is also persisted on the pocket via a new `pockets.service.set_imported_source` (entity isolation: the Pocket write stays in the pockets service), so a later re-publish reads real content.

Every failure mode (unreachable seed, seed blocked by robots, byte budget blown, deploy error, timeout) lands as `status: "failed"` with a fixed safe message on the report. No stack traces, no raw upstream text.

## SSRF posture

This endpoint asks our server to fetch an attacker-chosen URL, so the fetcher treats everything as hostile:

- URL floors at validation time (the endpoint 422s these before minting anything): http/https only, no credentials in the URL, no ports beyond 80/443, and a literal-IP host is checked against the deny ranges immediately.
- DNS is resolved by the crawler itself and every resolved address must be public. The connection is then pinned to the validated IP: the request goes to `scheme://ip/...` with the original Host header, and SNI/cert verification runs against the real hostname. Re-resolution between check and connect (rebinding/TOCTOU) can't swap in a private address. A mixed public+private DNS answer is rejected outright.
- Deny ranges cover loopback, RFC1918, link-local (incl. 169.254.169.254), CGNAT 100.64/10, unspecified/reserved/multicast, IPv6 ULA and link-local, and v4 addresses embedded in v6 forms (v4-mapped, 6to4, Teredo, NAT64).
- Redirects are followed manually, max 5, with the full URL + DNS + IP check re-run per hop. A redirect to `file://` or an internal host dies mid-chain, and the internal host is never contacted (test asserts zero requests).
- Responses stream against a 10MB per-fetch cap and the 50MB total budget, aborted the moment a cap is crossed. Per-fetch timeout 10s. No cookies (jar cleared per response), no auth, `trust_env=False` so an env proxy can't bypass the pin. Honest User-Agent, robots.txt honored (failure to fetch robots degrades to a warning, not a block).
- Resource floors: pages <= 50, depth <= 3, assets <= 200, tracked-URL sets bounded so a link-soup page can't balloon memory, politeness delay between fetches.

## Known limitations (deliberate v1 calls, noted in code + report warnings)

- Same-site matching is exact-host: `www.` and apex are treated as different hosts. The report says so.
- Absolute same-origin URLs are rewritten root-relative with a plain string replace; prose that happens to contain the site's own URL gets rewritten too.
- Query strings are dropped when normalizing page/asset URLs.
- The background task is process-local. A web-process restart drops an in-flight crawl and the report stays "queued". The workspace-jobs (ARQ) machinery was considered and rejected for this slice: its writeback merges job state into the pocket's rippleSpec, and an imported html pocket has no spec, so a successful crawl could be marked failed by the writeback gate. Durable queueing is a follow-up.

## How it's tested

35 new tests in `tests/ee/sites/test_import_crawler.py`, all network mocked (httpx MockTransport + fake resolver, zero live sockets): forbidden literal-IP/port/scheme/credential seeds, private-resolving hostname making zero requests, mixed DNS answers, redirect-to-private and redirect-to-file mid-chain, IP pinning asserted at the wire level, BFS depth/page caps, robots disallow with the skip counter, cross-host refs counted but never fetched, encoded path traversal refused a FileMap slot, byte-budget abort, plus the wired happy path (2-page site with css + img through the faked publish, report + pocket source checked) and the failed-report modes. SI-4's `test_import.py` updated: the NotImplementedError seam test became a "crawl gets scheduled" test.

```
uv run pytest tests/ee/sites/test_import_crawler.py tests/ee/sites/test_import.py -q
52 passed in 2.57s

uv run pytest tests/ee/sites -q
471 passed, 2 skipped in 27.53s

uv run lint-imports && uv run lint-imports --config ee/pyproject.toml
Contracts: 1 kept, 0 broken / Contracts: 31 kept, 0 broken
```

### DIFF
--- a/ee/pocketpaw_ee/cloud/leads/router.py
+++ b/ee/pocketpaw_ee/cloud/leads/router.py
@@ -22,6 +22,20 @@
 # ``rate_key``. The caller-controlled ``body.submitter_ref`` is no longer the
 # limiter key (it was randomizable to dodge the cap); it rides through only as an
 # opaque, non-PII label on the stored Lead.
+#
+# Updated 2026-07-22 (SI-4 — feat/sites-import-endpoint): added the NATIVE-FORM
+# capture sibling, POST /capture/form (final URL: {captureApiBase}/capture/form,
+# i.e. /api/v1/capture/form — captureApiBase already carries /api/v1, so the
+# spec's "/v1/capture/form" would have doubled the segment; this is the
+# router-consistent resolution and the cross-repo contract paw-sites' import
+# rewiring must target). IMPORTED sites rewire their <form>s to a plain
+# application/x-www-form-urlencoded POST here, with hidden fields ``paw_site_id``,
+# ``paw_key`` (the per-site signed key), ``paw_page``, ``paw_redirect`` and
+# optionally ``paw_form_type``. The endpoint runs the SAME hardening ladder as the
+# JSON capture (site exists → origin pin → constant-time signed key → payload size
+# cap → leads_service.capture), then 303-redirects to ``paw_redirect`` — which MUST
+# be a relative path (open-redirect guard: absolute / protocol-relative /
+# backslash / CR-LF all 400), resolved against the validated request Origin.
 
 from __future__ import annotations
 
@@ -29,7 +43,7 @@ import hashlib
 import json
 import secrets
 
-from fastapi import APIRouter, Depends, HTTPException, Query, Request
+from fastapi import APIRouter, Depends, HTTPException, Query, Request, Response
 
 from pocketpaw.sites_capture.ingest import origin_allowed
 from pocketpaw.sites_capture.models import MAX_PAYLOAD_BYTES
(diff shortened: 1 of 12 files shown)


prakashUXtech:
SSRF-review follow-up landed (40c59e51):

- Redirect scope guard: every hop was already SSRF-revalidated, but the final URL was never re-checked against the seed host. Non-seed fetches are now scope-locked — a same-site page or asset that 302s off-host is rejected (sites.import_crawl_offsite_redirect) and skipped with a warning, so a redirect can't pull foreign content into the tenant's deployed site.
- Seed apex-to-www re-seed: the seed is allowed to redirect off-host, and the crawl re-homes to the final host so the whole site imports instead of one page plus cross-origin soup — the common real case.
- The same-origin URL rewrite now lists both schemes, so an http:// ref on an https page rewrites cleanly instead of mangling to http:/path.

Two regression tests added (off-host asset redirect not imported; apex-to-www re-seed crawls both pages). Full run: 57 passed. The remaining review notes (wall-clock covers publish too, 16MB BSON vs 50MB budget, from-url concurrency) are LOW/follow-up and tracked in the tasks doc.

PR #1772, opened 2026-07-23
Actually rewire imported forms so their leads reach the workspace

prakashUXtech:
Imported sites were shipping their forms un-rewired. import_service used an interim Python derivation that hardcoded rewired=False and deployed the raw upload, so an imported form still posted to its original backend (formspree, etc.) and the site's leads never reached the workspace. The rewire pipeline existed in the paw-sites generator, but nothing on the backend called it — the served-artifact e2e passed because it exercised the pipeline directly, not the seam the backend runs.

Both the zip and crawl paths now run the unpacked/harvested files through the generator's import command (generator_client.run_import) before publish:

- The draft Site doc is minted first, so its signed_key bakes into each form and matches the key the capture endpoint verifies.
- The rewired source is what we persist on the pocket and deploy — not the raw upload.
- The authoritative report (real per-form rewired verdicts, original actions, page titles, asset tallies) is what we store and surface.
- A rewire failure fails closed: it raises rather than falling back to deploying the un-rewired site.

Verified against the real built generator command end to end (not just the mocked seam): a formspree form comes back posting to /capture/form with the original action preserved and rewired=true. 58 import + crawler tests green including a fail-closed test; full sites suite 476 green.

Depends on the paw-sites import-command PR (qbtrix/paw-sites#37) landing + a generator re-vendor for prod. Codes to that command's JSON contract.

### DIFF
--- a/ee/pocketpaw_ee/sites/generator_client.py
+++ b/ee/pocketpaw_ee/sites/generator_client.py
@@ -1020,6 +1020,69 @@ class _SubprocessRunner:
         return await self.build_static(project_dir, gate=True)
 
 
+async def run_import(
+    files: dict[str, str],
+    *,
+    site_id: str,
+    capture_api_base: str,
+    capture_signed_key: str,
+    _exec: Any = None,
+) -> dict[str, Any]:
+    """Run imported files through the paw-sites ``import`` generator subcommand and
+    return ``{source, assets, report}`` — the rewired html source map, the base64
+    binary sideband, and the authoritative import report.
+
+    This is what makes an imported ``<form>`` actually post to the capture API: the
+    generator runs buildImportPlan + rewireForms over ``files`` (path -> base64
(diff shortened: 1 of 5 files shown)


prakashUXtech:
### Pre-deploy code review

Ran a multi-agent review (5 dimensions, adversarially verified) before deploy. Two real issues, both fixed forward:

1. **Lead leak (HIGH):** the zip path minted the pocket with the raw upload and deployed the rewired source but never persisted the rewired source back to the pocket, so a re-publish/edit would redeploy forms still posting to the origin backend. Fixed in pocketpaw/pocketpaw#1773.
2. **Import cap mismatch (MED):** the generator's input cap (50 MiB on the base64+JSON wire) was below the encoded size of a 50 MiB raw import, so mid-size imports failed opaquely. Fixed in qbtrix/paw-sites#38.

Everything else (rewire correctness, fail-closed, signed-key handling, base64 round-trip, tenancy/injection) came back clean.

PR #1775, opened 2026-07-23
Paw Bar ship integration: concierge, action loop, and owner dashboard

prakashUXtech:
One branch that carries the whole Paw Bar arc on top of current dev, so we can work and verify from a single tree until ship. Not for squash review: the individual PRs stay the review units.

Contains, in merge order: the concierge seam stack (#1714, #1716, #1731, #1739, #1740 via its head branch), the decision-loop tenancy fix (#1715), and the concierge soul-write policy (#1728).

Merge conflicts resolved against dev's last 73 commits: additive header blocks in the Site model and sites service, and a same-anchor function collision in sites/service.py (dev's create_draft_site + our mint_foreign_site, both kept). Full paw-bar sweep on the merged tree: 166 passed.

Draft on purpose. The stack merges bottom-up with rebase when we ship; this branch is the working integration until then.


prakashUXtech:

PR #1776, opened 2026-07-23
Auto-provision a dedicated concierge agent per site

prakashUXtech:
## What

Every site's concierge is now answered by an agent that exists **for that site**, never a shared or universal one. This adds idempotent auto-provisioning on top of the existing manual-bind invariant (an unbound widget still 409s on chat).

New module `ee/pocketpaw_ee/paw_bar/agent_provisioning.py`:

- `ensure_site_agent(site, widget)` creates one dedicated agent through the agents service (never a direct Beanie write) and binds it to the widget via the store's `update_fields` path. A widget already bound to a live agent keeps that bind, so manual bindings are never overwritten. The slug is deterministic on the site id, so a retry after a partial provision adopts the same agent instead of minting a duplicate.
- The agent is created in the site's workspace, owned by the site owner, with a soul persona ("You are the concierge for <site>...") and the "The Site Concierge" archetype.
- A private `_seed_connectors` no-op marks where the connector-distribution wave will wire per-site connectors.

## Triggers (both failure-soft)

1. **Widget create** with no `agent_id`, when the pocket resolves to a published Site: provision and bind after insert, return the widget with `agent_id` set. A widget on a non-site pocket stays a plain unbound widget.
2. **Concierge enable**: flipping the kill switch on provisions the site's widget when it is still unbound.

A provisioning error logs and leaves the widget unbound rather than 500-ing the request. Chat then still returns 409 and the dashboard can offer a manual create.

## ASG-1 identity fields

`welcome_message`, `conversation_starters`, and a free-form agent `tags` field are **not present** on the Agent model on this branch, so their seeding degrades to a graceful no-op. The derivation logic (starters from catalog + gated-action labels, capped at 4) is unit-tested so it is ready the moment those fields land. The frame config gains a `starters` field (capped 4, empty default) threaded from the bound agent for the E3 bar slice.

## Sites service

Added a thin public `canonical_site_for_pocket(workspace_id, pocket_id)` wrapper over the private `_canonical_site_doc` so the provisioner resolves a pocket to its live Site through the same dedupe-aware logic without reaching into a private helper. Read-only, no write-path change.

## Tests

New `tests/cloud/test_paw_bar_agent_provisioning.py` (17 tests): naming/persona/starter derivation, the create trigger (auto-create + bind + AgentCreated emitted), idempotency, manual-bind respected, non-site skip, the enable trigger, the unbound-chat 409 regression, the ASG-fields-absent skip path, and starters riding the frame config.

Green: new suite (17) plus `test_paw_bar_backend`, `test_paw_bar_frame`, `test_paw_bar_concierge_settings` (64), and `test_paw_bar_ingest`, `test_paw_bar_concierge_chat`, `test_paw_bar_admin_aggregation` (83) as regression cover. Import-linter ee config: 31 kept, 0 broken.

### DIFF
--- a/ee/pocketpaw_ee/paw_bar/agent_provisioning.py
+++ b/ee/pocketpaw_ee/paw_bar/agent_provisioning.py
@@ -1,0 +1,316 @@
+# ee/pocketpaw_ee/paw_bar/agent_provisioning.py — auto-provision a DEDICATED
+# concierge agent per Paw Site.
+#
+# Created 2026-07-23 (feat/site-dedicated-agent): every site's concierge is
+# answered by an agent that exists FOR that site — never a shared/universal
+# agent. ``ensure_site_agent(site, widget)`` is idempotent: a widget already
+# bound to a LIVE agent is returned unchanged (manual binds are never
+# overwritten); otherwise ONE dedicated agent is created (via the agents service
+# — never a direct Beanie write) and bound to the widget through the paw-bar
+# store's ``update_fields`` path. Two triggers funnel here: widget-create (no
+# agent_id + the pocket resolves to a Site) and concierge-enable (the site's
+# widget is still unbound). Both are FAILURE-SOFT — a provisioning failure logs
+# and leaves the widget unbound rather than 500-ing the caller (chat still 409s;
+# the dashboard offers a manual create).
+#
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1778, opened 2026-07-23
Main agent drives /code with per-call file tools

DevRohit06:
## What this does

The /code surface used to give the main chat agent one coarse tool, `code_mode(task)`, that handed a whole coding task to a browser sub-agent. That sub-agent was removed — so the tool now delegates to nothing, and Code Mode is down.

This replaces `code_mode` with the four file tools the main agent reasons *with*, driving its own tool loop the way a coding agent does:

- `readFile(path)` — read a file
- `search(query)` — search the project
- `listDir(path)` — list a directory
- `writeFile(path, content)` — **propose** a full-file rewrite

The files still live in the user's browser, not on the backend, so each tool parks on the existing browser-delegate channel and pushes one `code_delegate` frame carrying `{corrId, tool, input}`. The browser runs the matching file-session verb and resolves `{output, isError}`. Reasoning moved to the main agent; execution stays in the tab.

## The write gate

`writeFile` never writes. It delegates the proposed contents to the browser, which stages them for the user's per-hunk review — the user accepting a hunk is the only thing that writes, and it happens long after the tool returns. So the tool, the system prompt, and the /code preamble all report a `writeFile` as a change **proposed for review**, never as a file that was written. The honesty rule is load-bearing: a model that says "I created the file" after one `writeFile` has told the user something false.

## Changes

- `delegates.py`: `delegate_to_browser(task, mode)` → `delegate_call_to_browser(tool, input)`; the frame carries a tool call instead of a task
- `code.py`: one `code_mode` tool → four file tools with a uniform outcome relay
- `surface_registry.py`: the /code allow-list scopes to the four tool ids
- `system_prompts.py` + `handlers/code.py`: rewritten for the file-tool vocabulary and the staging-honesty rule
- tests updated across the delegate channel, the MCP server, and the surface profile

## Testing

`pytest` green on the delegate channel, the code MCP server, the surface handlers, and the system-prompt tests (79 tests). `ruff` clean.

## Coordination

This is one half of a lockstep pair — it changes the `code_delegate` wire shape, and the browser executor changes with it. Ships with the paw-enterprise change (linked below). Base is `feat/code-mode`; Code Mode is non-functional end to end until both land.

### DIFF
--- a/ee/pocketpaw_ee/agent/mcp_servers/code.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/code.py
@@ -1,57 +1,64 @@
-# code.py — in-process MCP server exposing the ONE tool the main chat agent
-# uses to reach the user's code: ``code_mode``. Created: 2026-07-22
-# (feat/code-mode-tool, CD-2).
+# code.py — in-process MCP server exposing the file tools the main chat agent
+# uses to reach the user's code. Created: 2026-07-22 (feat/code-mode-tool, CD-2).
+# Reshaped 2026-07-24 (feat/code-mode-file-tools).
 #
-# What this is for. The /code surface is being inverted so the MAIN PocketPaw
-# agent drives the conversation and Code Mode becomes a sub-agent it delegates
-# to. This file is the seam. The main agent gets exactly one coarse tool rather
-# than four proxied file verbs, so the sub-agent keeps its own read/write loop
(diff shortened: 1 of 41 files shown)


DevRohit06:
Frontend half of the lockstep pair: qbtrix/paw-enterprise#668 — the browser executor that runs each delegated file tool call and keeps the write gate in the tab. Ship together.

PR #1797, opened 2026-07-27
Give the concierge something to know, and both sides of the conversation

prakashUXtech:
Closes the two open decisions the per-site concierge arc was parked on.
Stacked on `feat/site-dedicated-agent` (#1776), so review this diff alone.

## 1. The concierge starts knowing the business

A dedicated concierge was provisioned with a soul, a persona and a KB
scope, and nothing ever put anything in that scope. Ask one about
opening hours and it correctly answered "I don't know" about a business
whose own homepage says 8am. Demos only looked grounded because the
widget spec happened to carry a catalog.

A concierge run reads exactly one scope, `pocket:<pocket_id>`, and never
`agent:` or `workspace:` (a public caller must not reach the tenant's
whole KB). So the fix is to get the site's own pages into that scope,
which is what the new `sites/kb_ingest.py` does.

Content is read from the **pocket**, not the built artifact and not the
live URL. The pocket is durable, so this works at publish time, at
provisioning time for a site published months ago, and on a manual
re-sync. It also needs no network, so there is no SSRF surface: fetching
a site's own `url` would mean fetching a customer-controlled hostname
server side, the exact class of request the import crawler had to be
hardened against.

Three lanes, one extractor each. HTML keeps title, body copy and image
alt text and drops script and style. Svelte drops code blocks and
template expressions. Ripple has no HTML yet, so the spec is walked for
copy, skipping structural keys and rendering price-like numbers with
their key so a price stays searchable.

Triggers: a live publish and a dedicated-agent bind, both backgrounded
because ingest compiles articles and neither caller can wait. A preview
publish deliberately does not fire, so a draft never rewrites the live
KB. Owners get `GET .../knowledge` and `POST .../knowledge/sync`, the
latter awaited because someone clicked a button. The sync gates on a new
`paw_bar.manage` action rather than the existing read, since it is a
mutation that spends compute.

Re-syncs are idempotent: kb-go derives an article id from the source
string, so deterministic sources mean a re-sync versions an article
instead of duplicating it. The site records the ids it produced and a
later sync prunes the ones it stopped producing, which is how a renamed
page leaves the KB. It never clears the scope, because owner-uploaded
files live there too.

**Not covered, named rather than implied:** a site minted for a foreign
origin has no content in its pocket, so it syncs zero documents and
reports `no_content`. Grounding those means crawling the customer's own
origin through the hardened crawler, which is its own slice.

## 2. Transcripts have both sides

An owner opening a transcript saw only the agent's replies. The cause
was structural: every authed surface persists the user turn as a Message
document, but a concierge visitor is anonymous and has no thread, so
there was nowhere for that text to live. `ChatRunDoc.user_text` is that
missing half.

Because a visitor types free text, a stored line can carry a name, an
email or an order number. That is new personal data, so the owner gets a
switch: `concierge_store_transcripts`, defaulting on, independent of the
kill switch in both directions. With it off the concierge keeps working
and keeps storing its replies; only the visitor's words are dropped. The
agent receives the full message either way, so the switch governs
storage and never the answer. Stored copies are capped at 4000
characters. Turning it off stops collection from the next message on and
does not purge what is already stored.

Transcripts now render the question stamped at ask time before the
answer stamped at reply time. Either half can be missing and the other
still shows, so a retention-off site produces exactly the assistant-only
shape it produced before. A run that failed before answering keeps its
question, and the conversations list previews that question rather than
a blank row.

## Worth a look during review

- **A lost-update bug I introduced and fixed.** The sync's bookkeeping
  first used a whole-document save. It runs minutes after the publish
  that scheduled it, holding a Site snapshotted at that moment, so it
  silently rolled back a domain connected in between. Caught by an
  existing sites test; now a targeted `$set` of three fields, with a
  regression test pinning it.
- **The stack was already failing CI lint.** `dev` is clean, but both
  `integration/paw-bar-ship` and `feat/site-dedicated-agent` fail `ruff
  check` and `ruff format --check`, so #1775 and #1776 are red on their
  own. First commit here fixes it; those two still need it (or this
  branch) before they go green.
- The resolver refactor: `resolve_site_key_with_site` now holds the
  implementation and `resolve_site_key` wraps it. Same gates, same
  order, same statuses, and the Site is returned only after all of them
  pass. Tests cover the withholding on a bad key and a bad origin.

## Tests

993 passing across the paw-bar, sites, runs, kb and RBAC suites. `ruff
check`, `ruff format --check`, both import-linter configs and `atlas
build --check` are all green.

One pre-existing failure is untouched and unrelated:
`tests/cloud/runs/test_run_core.py::test_execute_run_threads_surface_meta_into_ctx`,
confirmed red on `feat/site-dedicated-agent` before this branch existed.

Docs: new `docs/concepts/concierge-knowledge.mdx`, registered in the
sidebar.

### DIFF
--- a/docs/concepts/concierge-knowledge.mdx
+++ b/docs/concepts/concierge-knowledge.mdx
@@ -1,0 +1,109 @@
+---
+title: "Concierge Knowledge and Transcripts"
+description: "What a site's concierge knows, where that knowledge comes from, and what it records about the visitors it talks to."
+section: Core Concepts
+ogType: article
+keywords: ["concierge", "paw bar", "site knowledge", "kb sync", "visitor transcripts", "retention"]
+tags: ["paw-bar", "sites", "knowledge", "privacy"]
+---
+
+# Concierge Knowledge and Transcripts
+
+A Paw Bar concierge is an agent attached to a published site. It answers visitors'
+questions about that specific business, so two things decide whether it is any good:
+what it is allowed to know, and what it writes down about the people it talks to.
+
+## What a concierge is allowed to know
+
+A concierge run is locked to **one** knowledge scope: `pocket:<pocket_id>`, the
+pocket its site was published from. It cannot read the `agent:` scope or the
+workspace-wide scope. That is deliberate. The concierge answers anonymous callers
+from the public internet, and a public caller must never be able to reach the rest
+of the tenant's knowledge base by asking the right question.
+
+The upshot is that a concierge's usefulness is decided entirely by what is in that
+one scope.
+
+## Where the knowledge comes from
+
+The site's own pages are synced into that scope automatically.
+
+The content is read from the **pocket**, not from the built artifact and not by
+fetching the live site. The pocket is durable, so a sync works long after a publish
+has finished; and reading it needs no network, so there is no server-side fetch of a
+customer-controlled hostname to harden against.
+
+Each authoring engine is read the way it stores content:
+
+| Engine | Source | What is extracted |
+|---|---|---|
+| `html` | the source map | page title, body copy, image alt text; scripts and styles dropped |
+| `svelte` | the source map | prose from the markup; script and style blocks and template expressions dropped |
+| `ripple` | the rippleSpec | copy from the spec, skipping structural keys; price-like numbers are rendered with their key so they stay searchable |
+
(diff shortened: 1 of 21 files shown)


prakashUXtech:

PR #1798, opened 2026-07-27
growth: workspace-scoped prospect store

prakashUXtech:
First slice of the growth module: a workspace-scoped prospect store in the standard 4-file shape, plus a dedicated `growth` arq queue seam for the outbound jobs that later slices will add.

## What's here
- `ee/pocketpaw_ee/cloud/growth/` — Prospect domain (frozen, tenancy required at construction), DTOs, service (sole Beanie writer, tenant filter on every read, fail-closed when no active workspace), CRUD routes under `/api/v1/growth/prospects` with tier/status/source filters
- `upsert_by_domain(workspace_id, data)` — create-or-update keyed on (workspace, normalised domain), backed by a unique index. This is the seam the ingestion slices call; `source` is preserved on update so capture provenance survives re-imports
- `growth/worker.py` — own arq queue so outbound bursts can't starve interactive chat runs (no-op heartbeat until the first real job lands; arq requires at least one function)
- 16 tests: CRUD, enum validation, duplicate-domain 409, parametrized cross-tenant 404s, upsert convergence
- api-reference: new Growth section

## Heads-up for review
- `ee/pyproject.toml` also drops two stale codeagent modules from the sandbox-guard import contract — they were deleted in the 2026-07-23 turn-agent refactor and import-linter hard-fails the whole ee config on nonexistent modules, so the ee lint gate could never pass without it. Happy to split it out if you'd rather.
- Stacked series: g2 (ingestion) will branch off this head.

## Verification
`pytest tests/cloud/growth/` → 16 passed. `lint-imports` core → 1 kept, 0 broken; ee config → 34 kept, 0 broken (includes the new Growth write-isolation contract). Pre-existing rbac/api-contract reds on dev confirmed identical with this diff reverted.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -66,6 +66,11 @@ rebuild — dynamic-source split + input-keyspace confinement) and GET
 /sites/by-pocket/{id}/native-artifact (serve the armed build's body_html + css
 for shadow render — per-GET arm-build cost, path-traversal-guarded CSS reader).
 
+Updated: 2026-07-27 (feat/growth-g1) — documented the Growth — Prospects
+section: workspace-scoped prospect store under /growth/prospects (create /
+get / list with tier|status|source filters / update), domain-deduped per
+workspace, cross-tenant ids 404. First slice of the /growth outbound engine.
+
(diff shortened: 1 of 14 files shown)


prakashUXtech:

PR #1799, opened 2026-07-27
growth: bulk prospect ingestion

prakashUXtech:
Second slice of the growth module, stacked on #1798.

Adds `POST /api/v1/growth/prospects/bulk` — batch prospect ingestion through the existing `upsert_by_domain` seam. Rows validate individually: a bad row records an indexed error entry and the rest proceed, so a Clay export with one malformed record doesn't lose the other 499. Re-posting the same payload is safe (upserts are idempotent — second run reports everything as updated). 500-row cap at the DTO boundary.

Also ships the `growth-ingest` crew skill in the workspace repo documenting the two ingest paths (Clay search + partner-directory scrape) and the payload contract.

## Verification
`pytest tests/cloud/growth/` → 21 passed (5 new: double-post idempotency, mixed-validity indexed errors, create/update split, 501-row 422, cross-tenant isolation). Both import-linter configs green (1 kept / 34 kept, 0 broken). ruff check + format clean.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -71,6 +71,10 @@ section: workspace-scoped prospect store under /growth/prospects (create /
 get / list with tier|status|source filters / update), domain-deduped per
 workspace, cross-tenant ids 404. First slice of the /growth outbound engine.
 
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1800, opened 2026-07-27
growth: outreach drafts with an enforced lifecycle

prakashUXtech:
Third slice, stacked on #1799.

Drafts are the unit the send gate (next slice) will propose and dispatch: per-channel outreach copy attached to a prospect, with a status machine enforced in the service — draft→proposed→approved→sent→replied, any non-terminal→rejected, everything else 422s with `draft.illegal_transition` and mutates nothing. Transition is deliberately dumb (no side effects); the Instinct wiring lands in the next slice on top of it.

Creating a first draft flips the prospect to `drafted` (never regressing later statuses). Subject is email-only, bodies must be non-blank — both at the DTO boundary. Ships the growth-draft crew skill (qualify → tier → per-channel copy via copywriting/humanizer) in the workspace repo, and the api-reference Drafts section.

## Verification
`pytest tests/cloud/growth/` → 51 passed (22 new: 3-channel drafting, legal chain, 13 parametrized illegal transitions, tenancy 404s, status flip + no-regress). Both import-linter configs green (Draft doc added to the Growth write-isolation contract). Route-table smoke confirms existing URLs unchanged after the router-prefix widening.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -75,6 +75,13 @@ Updated: 2026-07-27 (feat/growth-g2) — added POST /growth/prospects/bulk to
 the Growth — Prospects section: batch ingestion (max 500 rows) via the
 upsert-by-domain seam, per-row errors, idempotent re-runs.
 
+Updated: 2026-07-27 (feat/growth-g3) — added the Growth — Drafts section:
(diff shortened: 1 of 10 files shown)


prakashUXtech:

PR #1801, opened 2026-07-27
growth: LinkedIn manual send queue

prakashUXtech:
Stacked on #1800. The deliberately-manual slice: LinkedIn sends stay human (automation there gets accounts restricted), so the platform's job is to make the manual part effortless.

- `GET /growth/linkedin/queue` — queued LinkedIn drafts joined with the prospect's profile link, tier, and research brief, newest first
- `?format=md` — paste-ready markdown: one section per prospect with the connect note (with a live character counter against LinkedIn's 300 limit), the after-accept message, and draft ids
- `POST /growth/linkedin/{id}/mark-sent` — channel-guarded (422 `draft.wrong_channel`), then delegates to the existing lifecycle `transition`, so there's exactly one status enforcer and approved→sent stays the only legal path

No LinkedIn API, no new documents — two workspace-scoped queries in the service.

## Verification
`pytest tests/cloud/growth/` → 60 passed (9 new: queue membership incl. cross-tenant both directions, join fields + ordering, mark-sent legal/illegal paths, md export assertions, empty queue, unknown format). Both import-linter configs green. ruff clean.

PR #1803, opened 2026-07-27
growth: human-approval gate for outbound sends

prakashUXtech:
Stacked on #1801. The slice that makes 'nothing goes out without a human saying yes' structural rather than conventional — plus fixes for two cross-tenant/forgery holes a security pass found in the shared gate machinery.

## The gate
`POST /growth/drafts/{id}/propose` files an Instinct action carrying the rendered preview; approving it (single or bulk) flips the draft and enqueues dispatch. `approved` and `sent` are now gate-owned statuses — the public status route refuses them with 403 `draft.gate_required` even though the lifecycle allows them, so the only route onto those edges runs through an approved proposal. The dispatch job is a logging stub; the email and WhatsApp senders land in the next slices.

The executor re-checks at execute time that the proposer still holds `growth.manage`, failing closed on every error path — a since-demoted admin's pending proposals don't send.

## Security fixes (found by review, confirmed before fixing)
Two ways approval *content* could diverge from what the human actually saw. Both are inherited by every gated kind, so both are fixed generically rather than for growth alone:

- **Approve-with-edits skipped the tenancy check.** The workspace assert ran on the pre-edit action while the executor read the post-edit blob, so an approver could rewrite it to target another workspace's draft. Now `_pin_gated_blobs` restores tenancy, proposer, target, schema and chain ids from the *stored* blob on any edit, drops reserved keys the action never carried, and the tenancy sweep re-runs post-edit. Content stays editable — the mandates flow legitimately edits a `_belt_plan` task list while approving — but identity is pinned.
- **The generic propose route accepted forged gated blobs.** Any member could file an innocuous-looking Tray card whose parameters hid a `_growth_send` payload naming an admin as proposer. Reserved keys are now rejected there (422), and the executor derives the proposer from the action's own audit trail instead of the blob, failing closed when the two disagree.
- Growth RBAC was registered but wired to no routes — now enforced per-route (`growth.read` / `growth.write` / `growth.manage` on propose).

Also collapsed twelve per-kind workspace asserts into one chokepoint so a new gated kind can't be wired into three dispatch paths and miss the fourth.

## Verification
Each regression test confirmed failing with its fix reverted, then restored (F1 → 4, F2a → 5, F2b → 2, F3 → 2). Sweep: 412 passed across growth, ship, instinct, workspace-isolation and the RBAC matrix; blast-radius run across every gate/belt/mandates/mission-control suite → 959 passed. Both import-linter configs green (35 ee contracts). Pre-existing unrelated reds in tests/ee/sites confirmed identical on the clean base.


prakashUXtech:

PR #1804, opened 2026-07-27
growth: send approved emails through Mailtrap

prakashUXtech:
Stacked on #1803. The first real dispatcher — approved email drafts now actually leave the building.

The job loads the draft, refuses anything that isn't `approved` without touching the provider, sends via a per-workspace Mailtrap connector, writes a MessageLog audit row, then flips the draft through the existing `gate_transition` seam. No second status path.

## Decisions worth a look
- **The connector declares zero actions.** Credentials live in connector state like ship's, but a generic REST send action would be reachable from the agent/connector-execute surface — and that's a road around the approval gate. The token is read at send time and only resolves for *enabled* rows, so disabling the connector revokes sending with no separate kill switch.
- **`GROWTH_SENDING_DOMAIN` fails closed** — unset means nothing sends. The from-address is checked against it, and a value matching the deployment's own host is refused: cold outreach on the apex gambles the domain that carries password resets and invoices.
- **Record, then flip.** The audit row is written before the status change, so a message that physically left always has a record even if the flip fails.
- **Failures stay retryable** — provider error writes a failed MessageLog row and leaves the draft `approved`; nothing raises. The worker runs `max_tries=1` on purpose so outbound work is never auto-retried into a double-send.
- Every error string passes through a scrubber that strips the token; the hygiene test goes red if it's removed.

## Verification
215 passed across growth, ship and instinct (22 new). Coverage includes: exact send payload and headers, parametrized non-approved statuses making no provider call, provider rejection and transport failure leaving the draft retryable, fail-closed on unconfigured/disabled connector, sending-domain guards, and a credential-leak sweep across logs, the audit row, the draft row and the response DTO — including a 401 body that echoes the token back. Network-free via MockTransport. Both import-linter configs green; ruff and mypy clean on the new modules.

Pre-existing connector-suite failures confirmed identical with this diff stashed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -225,6 +225,19 @@ The web dashboard (`frontend/`) is vanilla JS/CSS/HTML served via FastAPI+Jinja2
   env-configurable), marking queued/running `ChatRunDoc`s older than 10 minutes as
   `interrupted` so runs abandoned by a backend restart surface a retry affordance
   instead of leaving clients subscribed forever.
+- **Growth outbound config (`GROWTH_SENDING_DOMAIN`)**: the secondary domain the
+  `/growth` engine sends cold outreach from — **required**, with no default. The
+  dispatch worker fails closed when it is unset (nothing goes out), validates the
+  from-address against it at send time, and refuses a value equal to the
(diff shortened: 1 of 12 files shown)


prakashUXtech:

PR #1805, opened 2026-07-27
growth: follow-up sweep that proposes, never sends

prakashUXtech:
Stacked on #1803 (sibling of #1804 — both branch off the gate slice).

A daily cron on the growth queue finds sends that went quiet past `GROWTH_FOLLOWUP_DELAY_DAYS` (default 4) and files a follow-up draft back through the **existing** propose path. It stops at `proposed` — the nudge lands in the Instinct tray and waits for a human like any other send. Max 2 follow-ups per thread (`GROWTH_FOLLOWUP_MAX`), then the prospect goes `dead` and we stop touching them.

## Decisions worth a look
- **The sweep inherits a human proposer.** A cron has no user, but the executor re-checks the proposer's *current* `growth.manage` role at dispatch. A "system" proposer would produce follow-ups that look approvable in the tray and then fail closed at send — a silent dead end. So the sweep reads the proposer off the thread's last send and reuses it; a thread with no resolvable proposer is skipped, because a proposal nobody can execute is worse than no proposal.
- **"No existing follow-up" means no *open* one.** Read literally as "none ever", the cap of 2 would be unreachable — the first nudge would block every later one. Open-status scoping is also what makes the sweep idempotent, and a human-rejected follow-up doesn't burn a cap slot.
- **The delay measures from the last touch**, not the first, or an already-nudged thread comes due again immediately.
- If proposing fails after the draft exists, the draft is retracted to `rejected` — otherwise it sits in `draft` and stalls that thread on every future pass.

## Verification
213 passed across growth, ship and instinct (20 new). Covers the five acceptance cases, cron registration on the growth queue, and a three-sweep cycle ending in `dead`. Time is frozen by injection, no sleeps. Mutation-checked: deleting the delay check or the open-follow-up guard turns 4 tests red, so they aren't passing vacuously. Both import-linter configs green; ruff clean, no new mypy errors.

**One coordination note for review:** the send timestamp currently resolves from the draft's own `sent_at`/`updatedAt`. #1804 lands a separate MessageLog collection — after both merge, `_resolved_sent_at` (one clearly-marked function) should point at it.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -93,6 +93,14 @@ Security review follow-up: documented per-route growth RBAC
 and the fact that a _growth_send blob can only be minted by this route —
 the generic POST /instinct/actions refuses reserved gated parameter keys.
 
+Updated: 2026-07-27 (feat/growth-g7) — added the Growth — Follow-ups section:
+the daily `growth.followup_sweep` arq cron on the `growth` queue turns a send
+that went quiet into a `variant: "follow_up"` draft filed back through the
+same `_growth_send` gate (proposed, never approved or sent), capped at
+GROWTH_FOLLOWUP_MAX per prospect+channel after which the prospect is retired
+to `dead`. Documented both env knobs (GROWTH_FOLLOWUP_DELAY_DAYS,
+GROWTH_FOLLOWUP_MAX).
+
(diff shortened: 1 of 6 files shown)


prakashUXtech:

PR #1806, opened 2026-07-27
growth: WhatsApp sends behind a hard opt-in guard

prakashUXtech:
Stacked on #1803 (sibling of #1804 and #1805 — all three branch off the gate slice).

Approved WhatsApp drafts now send via MSG91, but only to a prospect with a recorded opt-in. Meta's policy isn't advisory here: business-initiated templates to non-opted-in numbers wreck the WABA quality rating and get numbers banned — for the whole tenant, not the one draft. So the guard is a service-layer invariant, not a UI convention.

**A prospect without opt-in produces zero provider interaction.** The check fires before credentials are even resolved (whatsapp.py step 3, ahead of the client build), writes a `blocked`/`not_opted_in` compliance row, raises a typed error, and leaves the draft `approved`. Success walks approved→sent through the existing `gate_transition` seam.

- Inbound webhook fails closed on bad, missing, replayed **and** unconfigured signatures — nothing is read from an unverified body — and always returns a constant `{"ok": true}` so it can't be used to probe whether a number is in our database.
- `GROWTH_WHATSAPP_MAX_PER_HOUR` (default 20) refuses the N+1th send with no provider call; refused attempts don't consume the window, and `0` refuses everything rather than quietly meaning unlimited.
- Authkey resolves per workspace through connector state with no env fallback; the credentials repr redacts it.

## Verification
237 passed across growth, ship and instinct. Network-free by construction — the fake is injected at the client-build seam, so the class that imports httpx is never constructed on any test path. The critical test asserts the fake was **never invoked** for a non-opted-in prospect. Both import-linter configs green; ruff clean.

## Known gaps for review
- **Not operator-usable as shipped:** there's no `connectors/msg91.yaml`, and the state store won't create rows, so the config has to be inserted by hand until a follow-up adds the connector definition. Dispatch degrades cleanly to `blocked`/`not_configured` meanwhile.
- **The MSG91 request envelope is unverified against a live account** — their public docs don't publish the JSON body shape. Worth one live smoke before production; the failure mode is a clean error row, not a silent send.
- The hourly cap counts log rows rather than using an atomic counter, so concurrent workers could overshoot slightly.
- Webhook tenancy is number-derived; resolving it from the receiving integrated number needs the connector row to be queryable by config value.
- `WhatsAppSendLog` needs reconciling with #1804's MessageLog once both land.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -82,6 +82,15 @@ GET /growth/drafts with prospect|channel|status filters, POST
 draft→proposed→approved→sent, sent→replied, non-terminal→rejected; illegal
 moves 422 draft.illegal_transition.
 
+Updated: 2026-07-27 (feat/growth-g6) — documented the Growth — WhatsApp
+dispatch section: the growth.dispatch job's channel="whatsapp" branch sends via
+MSG91 behind a HARD prospect.opted_in guard (not opted in ⇒ no provider call at
+all, typed error, blocked send-log row, draft left approved), the guard order,
+the per-attempt WhatsAppSendLog compliance record, connector-state credential
+resolution (no env fallback for the authkey), the fail-closed inbound webhook
+POST /growth/webhooks/msg91, and the GROWTH_WHATSAPP_MAX_PER_HOUR /
+GROWTH_MSG91_WEBHOOK_SECRET environment variables.
+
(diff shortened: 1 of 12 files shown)


prakashUXtech:

PR #1807, opened 2026-07-27
/growth: the outbound engine, gated end to end

prakashUXtech:
The whole `/growth` wave, consolidated. The eight slice PRs (#1798–#1806) and
the three stacked ones (#1808, #1809, #1812) are closed — every commit is here,
and this is the single review surface.

`/growth` is the outbound engine as a paw-os verb surface: find people worth
reaching, write to them, and never send anything a human didn't approve.

```
  prospect → draft → propose → INSTINCT TRAY → approve → send
                                     ↑                     ↓
                              a human, always      email · whatsapp
                                                   linkedin (manual)
                                                          ↓
                                          no reply → daily sweep proposes
                                                     a follow-up
```

## The gate

Nothing sends without an approved `_growth_send` Instinct proposal. There is no
direct-send code path anywhere — not in the service, not in the router, not in
the agent's tool surface. `connectors/mailtrap.yaml` ships `actions: []` for the
same reason: every route around the gate stays shut, including friendly-looking
ones.

WhatsApp sends are refused outright for a prospect without `opted_in`, in the
service before credentials are even resolved.

## The agent's reach

An in-process MCP server (`pocketpaw_growth`, 13 tools) lets the chat rail work
a prospect end to end and stop at **proposed**. No tool sends. No write tool
takes a `status` argument. `gate_transition` is not imported, and a test asserts
it stays that way — against the tool *list and schemas*, so a future tool that
opens the door fails the suite rather than passing review.

The discovery tools follow the same shape: the agent can define and dry-run a
hunt, but no tool accepts a `cadence`. Scheduling is a standing spend
commitment, so a human makes it.

## Discovery

An **ICP** is a standing description of who you want, in a sentence, plus a
cadence. Discovery is not a scraper and not a vendor integration — it is the
agent's own research on a schedule, filed through the existing
`upsert_by_domain` path with `source='discovery'`, the `icp_id`, and the URLs
the research actually read.

**`recordable_emails` is the one door** that turns evidence into a storable
address. Anything guessed, merely claimed, or missing its source URL is dropped,
so a prospect with no email is the honest, normal result. This is enforced in
code, not asked of a prompt: Clay and Apollo buy verification waterfalls, an LLM
cannot, and a guessed address bounces onto the sending domain's reputation.

The researcher agent holds exactly `WebSearch` and `WebFetch` under
`tool_mode: "exclusive"` — no write tools, because one would route around that
door.

## Multi-client

Prospects and hunts scope by the existing project primitive, one project per
client. Email sender identity resolves per project at dispatch, per field, with
the workspace as fallback — the client's name in `from_name` and their inbox in
`reply_to`. `from_email` stays pinned to `GROWTH_SENDING_DOMAIN`: mail claiming
a domain you don't hold fails DMARC and burns the domain you *do* hold.

## Verified live, not just tested

- 366 tests in `tests/cloud/growth/`, both import-linter configs clean.
- One real hunt against Pune web studios: 5/5 companies real (every domain
  returns 200), every email observed on a contact page with its URL recorded.
- The filing path run twice: `filed=3`, then `filed=0 skipped_existing=3` —
  dedupe holds.
- Preview verified to write nothing (discovery-sourced prospect count unchanged).

**One bug this caught that tests could not.** 23 researcher tests passed while
every live hunt returned "response could not be read": the event collector was
sweeping in `tool_result` events, which carry the raw WebSearch payload — itself
JSON — so the extractor parsed a page of search hits instead of the answer. The
fake in the tests never emitted tool events, so it could not have caught it.
Fixed, with a regression test.

## Not done, deliberately

- The 06:00 discovery cron has never fired on a schedule here — it needs an arq
  worker. The sweep itself is tested; the trigger is not.
- MSG91's request envelope is unverified against a live account.
- `connectors/msg91.yaml` doesn't exist yet — it needs a hand-inserted row.

Frontend counterpart: qbtrix/paw-enterprise#698. The list endpoint is a
**breaking change** (`{items, next_cursor, total}` instead of a bare array), so
the pair merges together or the list goes blank.


prakashUXtech:
Rebased onto current dev (`fa593da`, 84 commits on the branch now). Notes from the rebase, since a few spots needed judgment:

- `ee/pyproject.toml` — both sides had independently fixed the stale codeagent import-linter contract; kept dev's package-form fix, since the hand-kept name list is what went stale in the first place.
- `ee/pocketpaw_ee/instinct/router.py` — dev added the `_customer_reply` tenancy gate while this branch introduced the `_assert_gated_workspaces` chokepoint. Kept the chokepoint and folded the customer-reply guard into it, so both kinds run at all four dispatch paths (approve / bulk-approve / reject / bulk-reject).
- Rebasing linearized away two adaptations that only lived in the branch's own merge commits — one was a live AttributeError in the WhatsApp dispatch path (`load_draft_for_dispatch` vs the real `get_draft_for_dispatch` seam). Restored byte-for-byte from the pre-rebase integration tip in `96988b3`.
- `tests/ee/test_paw_bar_customer_reply_router.py` seeded its forged `_customer_reply` blob through the generic propose route, which this branch's F2 hardening now refuses with a 422 — that refusal is the fix working. The harness seeds in-process now, matching how `decision_loop` files those proposals in production (`6321a88`).

Verification on the rebased tip: `tests/cloud/growth/` 366 passed; growth + instinct-consumer sweep 515 passed; `lint-imports` on both configs, 37 contracts kept, 0 broken.

PR #1809, opened 2026-07-28
Give the growth agent tools — read, draft, propose, never send

prakashUXtech:
`/growth` now has a chat rail (paw-enterprise#707), and until this lands that
rail is an agent that can discuss your prospects but can't touch one. Every
other verb surface has an MCP server; this is growth's.

`pocketpaw_growth`, nine tools, all of them through `cloud.growth.service`:

| Tool | RBAC |
|---|---|
| `growth_list_prospects` — filters, `q`, `sort`, `cursor` | `growth.read` |
| `growth_get_prospect` — the record plus its drafts | `growth.read` |
| `growth_list_drafts` | `growth.read` |
| `growth_linkedin_queue` | `growth.read` |
| `growth_upsert_prospect` | `growth.write` |
| `growth_create_draft` / `growth_update_draft` | `growth.write` |
| `growth_propose_send` / `growth_propose_send_batch` | `growth.manage` |

Registered through a `pocketpaw.mcp_servers` entry point and added to the
import-linter Growth contract's source modules, so a Beanie import from the
agent surface breaks CI rather than review.

## The gate

The agent's reach stops at a **proposed** draft. No tool sends. No write tool
takes a `status` argument — the only `status` enum on the surface is a read
filter on `growth_list_drafts`, and it moves nothing. Neither
`gate_transition` nor `mark_linkedin_sent` is referenced in the module.

`TestTheGateHolds` asserts that against the tool list and the schemas rather
than against today's tools, so a future tool that opens the door fails the
suite. A dynamic pass drives every tool and asserts the drafts stopped at
`proposed`.

## Three things beyond the brief

**`service.update_draft` + `PATCH /growth/drafts/{id}`.** `growth_update_draft`
had no backing service function. The new one refuses anything past `draft` with
403 `draft.not_editable` — from `proposed` on, the stored body is what the Tray
shows a human and what the worker puts on the wire, so an edit there is a send
bypass wearing an edit's clothes. Exposed over HTTP too, since the UI wants the
same thing.

**`growth_list_drafts`.** `propose_send_batch` is unusable without a way to
collect draft ids across prospects; `get_prospect` reaches one at a time.

**`cursor` on `growth_list_prospects`.** The page already returns
`next_cursor`; without the argument the agent can't reach row 200.

`growth_upsert_prospect` merges rather than replaces. `upsert_by_domain`
overwrites every mutable field, so a naive re-upsert would reset a prospect in
`in_sequence` back to `new` and blank whatever the agent didn't repeat. The
handler reads the row first and carries the rest forward. Prospect status is
never settable by the agent.

## Tests

268 pass in `tests/cloud/growth/` (28 new MCP tests, 6 new draft-edit tests),
43 across the MCP registry and instinct, both import-linter configs clean.
`tests/ee/` minus sites: 1604 passed, 1 skipped.

`tests/ee/sites` is 61 failed / 3 errors and pre-existing — identical on the
base branch, nothing to do with growth.

One red this branch did cause and fixes: `test_mcp_claude_sdk.py`'s
`_strip_builtin_servers` needs every new ambient server listed. `pocketpaw_code`
and `pocketpaw_ship` were never added when they landed, so that file had 6
failing assertions on the base branch. Green now.

## Worth knowing before merge

- **There is no `GROWTH` SurfaceKind.** The tools reach the agent anyway —
  `surface='growth'` falls back to `GENERIC`, whose default profile applies no
  MCP restriction, and this server is ambient. But the rail gets no preamble
  handler and no profile-scoped allowlist. Natural next slice; `GROWTH_TOOL_IDS`
  is exported ready for it.
- The entry point registers on `uv sync --dev --group ee` (editable installs
  cache entry points in dist-info). Deploy needs the same.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -110,6 +110,15 @@ Security review follow-up: documented per-route growth RBAC
 and the fact that a _growth_send blob can only be minted by this route —
 the generic POST /instinct/actions refuses reserved gated parameter keys.
 
+Updated: 2026-07-28 (feat/growth-mcp) — added the Growth — the agent surface
+section: the nine pocketpaw_growth in-process MCP tools the chat agent on the
+/growth rail drives, the table of how that surface is narrower than the HTTP
+one, and why the agent's reach ends at proposed (no send tool, no status
+argument, no route to gate_transition). Also added PATCH /growth/drafts/{id}
+— edit a draft's copy while it is still `draft`; anything past that is
+403 draft.not_editable, because from proposed on the stored body is what the
+Tray shows and what the worker sends.
+
 Updated: 2026-07-28 (feat/growth-api-scale) — the prospect list grew a scale
 surface. BREAKING: GET /growth/prospects now returns
 {items, next_cursor, total} instead of a bare array. Added q search across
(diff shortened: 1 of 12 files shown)


prakashUXtech:
Landed on `integration/growth-v1` by fast-forwarding the stack tip — every commit here is on the integration branch unchanged (GitHub reports no diff between the two). Closing the stacked PR; the review record stays readable here, and `integration/growth-v1` → dev (#1807) is the landing gate.

PR #1812, opened 2026-07-28
One pipeline per client, and a prospect that can be just a domain

prakashUXtech:
An agency isn't one outbound pipeline — it's one per client. `/growth` v1
scoped everything to a workspace, so an agency running outbound for eight
clients got eight pipelines in one undifferentiated list, all sending from the
same identity. This scopes growth by the project primitive that `tasks` and
`cycles` already use, and fixes the import path that made the whole surface
unusable from zero.

Spec: `docs/design/drafts/2026-07-28-growth-discovery-and-projects.md` §3, §5.

## A prospect can be just a domain

`name` and `company` lose `min_length=1` and default to `""`, meaning *not yet
known*. That's the honest shape an import arrives in — someone hands you forty
agency websites and nothing else; research fills the rest in later. `domain`
stays required and still normalises.

The sweep for what breaks on an empty name found three real ones, all fixed:
the LinkedIn markdown export would have written `## — `, the follow-up subject
would have read `Following up —`, and the MCP upsert's `or "unknown"` fallbacks
would have **stamped the literal string "unknown" over a stored blank** on any
enrichment call.

The MCP handler's own refusal to create a nameless prospect is untouched. That
guard is about a different caller: when the agent researches a company and
files it, it should know the name, and an "unknown / unknown" row is worse than
a refusal it can act on. A human pasting a domain list carries no such
obligation. Same field, two callers, two truths.

## `project_id` on Prospect

Mirrors `tasks` exactly — nullable on the domain object, the doc and the DTOs,
validated at entry against the workspace, an optional filter on list, facets
and search, plus a `(workspace, project_id, createdAt)` index. A workspace not
using projects sees no change.

Growth got its own `_ensure_project_in_workspace` rather than importing
`tasks`'. Not a contract problem — both linter configs pass either way — but
that helper is private, and reaching across an entity boundary for a sibling's
private function is the drift the 4-file rule exists to prevent. Both copies
call the same public `projects.service.exists_in_workspace`.

Two edges worth knowing: an upsert only ever *sets* `project_id` and never
clears it, because otherwise every agent enrichment call would silently orphan
a client's prospect — un-assigning is an explicit `PATCH` with `""`. And a row
naming a foreign project is now one indexed error inside a bulk ingest rather
than a 404 that kills the whole batch.

## Sending as the client — and the part that can't work

Per-**field** fallback to the workspace default, resolved in front of the
existing credential read. A project that overrides only the display name keeps
the workspace reply-to instead of losing it, which is the common case: one
agency inbox, many client names.

**`from_email` stays pinned to `GROWTH_SENDING_DOMAIN`, and the spec was wrong
to imply otherwise.** You cannot send as `hello@theirdomain.com` without their
DNS — mail claiming a domain you don't hold fails DMARC, lands in spam, and
burns the reputation of the domain you *do* hold. The honest shape is
`clientname@<sending-domain>` in the from-address, the client's name in
`from_name`, and the client's own inbox in `reply_to`, which is how agencies
actually send and reads correctly to the recipient. Sending from a client's own
domain needs their DNS delegated — an onboarding flow, not a config key.

The token stays the workspace's, so disabling the connector still revokes every
project at once. Email only; WhatsApp is untouched and needs Meta Tech Provider
onboarding separately.

## The follow-up sweep works one client at a time

Threads partition by `(workspace, project)` before being worked, over the same
collapsed thread map the flat loop already walked — one thread, one prospect,
one project — so every thread is still visited exactly once. The open-follow-up
idempotency guard is untouched in place, and a test pins that a second pass over
grouped data proposes nothing. Cheap checks stay in front of the prospect read,
so a not-yet-due backlog costs no more than before.

## Tests

294 pass in `tests/cloud/growth/`, up from 274 on the base — 20 new. Both
import-linter configs clean (1 and 35 contracts). mypy reports the same 10
errors on the same lines as the base commit; none added.

`tests/cloud/tasks/` has 14 failures, confirmed pre-existing by running them on
the base commit — identical 14.

No new routes, so the guard-coverage surface in `test_gate.py` is unchanged by
design: `project_id` rides as a query param on the existing list and facets
routes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -236,7 +236,11 @@ The web dashboard (`frontend/`) is vanilla JS/CSS/HTML served via FastAPI+Jinja2
   invoices and receipts with it. The provider credential itself is **not** an env
   var: it is per-workspace connector state on the workspace's `mailtrap` connector
   row (`MAILTRAP_API_TOKEN`, plus optional `MAILTRAP_FROM_EMAIL` /
-  `MAILTRAP_FROM_NAME`), so disabling the connector revokes sending immediately.
+  `MAILTRAP_FROM_NAME`, `MAILTRAP_REPLY_TO`, and `MAILTRAP_PROJECT_SENDERS` — a
+  `{project_id: {from_email, from_name, reply_to}}` map so an agency sends as the
+  client whose project owns the prospect, resolved per field with the workspace
+  values as the fallback), so disabling the connector revokes sending immediately
+  for every project at once.
   See `ee/pocketpaw_ee/cloud/growth/connector.py` and `docs/api-reference.md`.
 - **Session supervisor config**: `POCKETPAW_SESSION_SUPERVISOR` (default OFF). When
   truthy (`1`/`true`/`yes`/`on`), the cloud chat executor drives every agent turn
--- a/connectors/mailtrap.yaml
+++ b/connectors/mailtrap.yaml
@@ -37,6 +37,21 @@ auth:
     - name: MAILTRAP_FROM_NAME
       description: Display name on the From header (optional).
       required: false
+    - name: MAILTRAP_REPLY_TO
+      description: >-
+        Reply-To header for cold outreach (optional). Unlike the from-address
+        this is NOT constrained to the sending domain — it only tells the
+        recipient where to write back, so it can point at a real inbox.
+      required: false
+    - name: MAILTRAP_PROJECT_SENDERS
+      description: >-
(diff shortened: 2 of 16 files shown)


prakashUXtech:
Landed on `integration/growth-v1` by fast-forwarding the stack tip — every commit here is on the integration branch unchanged (GitHub reports no diff between the two). Closing the stacked PR; the review record stays readable here, and `integration/growth-v1` → dev (#1807) is the landing gate.

PR #1816, opened 2026-07-29
The concierge remembers what the visitor already told it

prakashUXtech:
Stacked on `feat/concierge-knowledge-and-transcripts` (#1797), so review this
diff alone.

## The defect, reproduced before and after

Before, on a live local backend against the Ridgeline HVAC demo site:

    visitor: My name is Priya and I have exactly 12 freelance clients.
    bot:     Nice to meet you, Priya.
    visitor: What is my name, and how many clients did I say I have?
    bot:     I don't have that information, you haven't shared your name...

After, same rig, same two turns:

    bot:     Your name is Priya, and you said you have 12 freelance clients.

This was not a missing feature so much as a structural hole. Every authed
surface persists the user turn as its own Message document, so history reads
back. A concierge visitor is anonymous, has no thread and no Message rows, so
there was nothing to read and every turn was answered cold.

The transcript work on the base branch is what makes the fix small: the stored
run docs already hold both halves of every exchange, so they are the memory
too. Same rows, same query, shaped for the model instead of for the dashboard.

## Isolation is the property that matters

`_concierge_runs_for_visitor` extracts the query the owner transcript already
used, so there is exactly ONE definition of "this visitor's turns" rather than
a second one that can drift. All four predicates carry weight: workspace,
context_type, the site's own pocket, and the anonymous customer handle. Tests
cover a sibling visitor, a sibling site, and another tenant, each proving their
turns never appear.

## Bounds

Three, each closing a different hole: twelve exchanges replayed, 2000
characters per line, 12000 overall. Turns are fitted newest-first so the budget
drops the oldest and the replay stays contiguous rather than gappy, and a
single long reply is clipped rather than allowed to evict every other turn.
Without these, a visitor could grow the prompt indefinitely by pasting, and
every turn would re-send and re-bill the whole conversation.

## Privacy

Gated on the same `concierge_store_transcripts` toggle as the write. Retention
off means no stored text, so no memory. That is the owner's privacy choice
working, not a gap to route around with a parallel store, and the docs now say
so plainly.

Reads happen before the current turn's run doc is written, so the visitor's
message rides in `content` exactly once. A failed history read answers without
memory rather than 500-ing the visitor.

## Tests

79 passing across the concierge chat and admin aggregation suites, including
the four isolation cases, both bounds, the clip-not-evict case, the
failure-soft path, and no-duplicate-current-message. `ruff check`, `ruff format
--check` and both import-linter configs are green.

Verified live end to end, not only in unit tests.

### DIFF
--- a/docs/concepts/concierge-knowledge.mdx
+++ b/docs/concepts/concierge-knowledge.mdx
@@ -95,7 +95,7 @@ an email address, or an order number. Each site therefore has its own switch:
   `GET`/`PATCH /paw-bar/admin/site/{site_id}/settings`.
 - With it **off**, the concierge keeps working and keeps storing its own replies.
   Only the visitor's words are dropped, and transcripts for that site show the
-  assistant side alone.
+  assistant side alone. It also stops remembering across turns — see below.
 - The agent always receives the full message either way. The switch governs storage,
   never the answer.
 - It is read on every message, so turning it off stops collection from the next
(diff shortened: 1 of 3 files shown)


prakashUXtech:

PR #1817, opened 2026-07-30
Put the concierge on the site, and make the bar actually render

prakashUXtech:
DEMO-CRITICAL. Stacked on `feat/concierge-knowledge-and-transcripts`
(#1797), so review this diff alone.

## A published site had no concierge on it

I framed a real deployed demo site in a browser. `document.querySelectorAll('script')`
came back **empty**, `window.__PAWBAR__` undefined, head just charset and title.
A site we generate, with a concierge we auto-provision, shipped without it.

Three causes, all confirmed:

1. The bar was embedded ONLY by a snippet the dashboard showed a human to
   copy-paste.
2. That snippet's script URL was `PAW_BAR_CDN_URL`, whose fallback is
   `https://pp.pocketpaw.dev/widget.js` — a host the code comment itself says is
   not provisioned. A pasted snippet would have 404'd.
3. Nothing in the publish path ever wrote the snippet into the built HTML.

So: the loader is now served by the backend at `GET /paw-bar/widget.js`, and a
live publish injects the snippet into every built HTML page when the site earns
one (concierge on, real embed key, a widget for the pocket, an agent bound). The
script URL derives from the API base the site's own capture endpoint already
uses, never a CDN host, so a locally served site gets a working local URL and
there is no second place to keep in sync. Injection is idempotent on a marker
attribute, skipped entirely for a preview publish, and failure-soft — a site
going live matters more than its bar. A live publish also adds the site's OWN
deployed host to its `allowed_origins`, never a wildcard.

## Then the bar rendered as an empty grey box

Worth reading even if you skip the rest, because it is one character and it
would have been very hard to spot on a shoot day.

The bar mounted, sized itself and docked correctly, and showed nothing. The
frame endpoint returned 200 with valid HTML and every asset resolved. Server
side, nothing looked wrong.

A CSP host-source **with no port matches only the scheme's default port**. So
`frame-ancestors 127.0.0.1` refuses to be framed by `http://127.0.0.1:4174`.
`allowed_origins` is normalized to bare hosts, so no entry ever carried a port,
which means **every site served anywhere other than 80 or 443 was unframeable** —
every local, dev and demo deploy, and any customer site on a non-standard port.

Portless entries now emit `host:*`. That is a consistency fix, not a loosening:
the origin gate this CSP mirrors (`site_keys.origin_allowed`) compares hosts and
ignores the port, so any port on an allowed host was already accepted for chat.
The CSP was the stricter of the two. Explicit ports are honored as written and
the header-injection guard is untouched.

Four assertions pinned the old string, including the owner preview frame's. The
real dashboard origin carries a port, which is why the owner preview kept
working while the public bar never did.

## Verified in a browser, not only in tests

After the fix, on a real published site: the glass bar renders docked
center-bottom with its live dot and input, and a chat from the site's own origin
returns a properly grounded answer — real hours, the after-hours fee, the
service email, all from the site's own pages.

## Tests

636 passing across the sites, frame, widget-loader, concierge chat, settings,
admin aggregation and site-key suites. `ruff check`, `ruff format --check` and
both import-linter configs green.

## One thing to know before shooting

The bar and the API must be same-scheme as the page. A site on HTTPS
`workers.dev` cannot load an HTTP localhost loader or reach an HTTP localhost
API — the browser blocks both as mixed content. Shoot the local publish, or
provision public hosting for the loader and the API.

### DIFF
--- a/ee/pocketpaw_ee/paw_bar/agent_provisioning.py
+++ b/ee/pocketpaw_ee/paw_bar/agent_provisioning.py
@@ -1,6 +1,16 @@
 # ee/pocketpaw_ee/paw_bar/agent_provisioning.py — auto-provision a DEDICATED
 # concierge agent per Paw Site.
 #
+# Updated 2026-07-30 (feat/paw-bar-autoembed): extracted ``site_widget(pocket_id,
+# workspace_id)`` — the "which paw-bar widget belongs to this site's pocket"
+# lookup that ``provision_on_concierge_enable`` used to do inline. The publish
+# path now needs the SAME answer (to decide whether a site has earned an embedded
+# concierge bar and which widget id to put in the snippet), and two copies of a
+# tenancy-scoped lookup is exactly the kind of pair that drifts apart. Behaviour is
+# unchanged: the same workspace-scoped ``list_widgets(limit=1)``, the same
+# empty-pocket_id guard that keeps a blank pocket from widening the query onto a
+# sibling's widget.
+#
 # Created 2026-07-23 (feat/site-dedicated-agent): every site's concierge is
 # answered by an agent that exists FOR that site — never a shared/universal
 # agent. ``ensure_site_agent(site, widget)`` is idempotent: a widget already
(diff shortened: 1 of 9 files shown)


prakashUXtech:

PR #1818, opened 2026-07-30
Paw Bar: email the decision to visitors who leave the page

prakashUXtech:
## What this closes

The decision loop had an on-page half only. A visitor asks something that needs the owner's call, the widget polls for the answer - and the moment they close the tab, the owner's decision goes nowhere. This adds the async half: the visitor can leave an email while the request is pending, and when the owner approves or declines, they get one email with the same reply the poll would have shown.

## Endpoint contract (for the widget team)

`POST /paw-bar/decision-contact`

Body:
```json
{
  "widget_id": "pp_...",
  "signed_key": "<Site.signed_key>",
  "customer_ref": "<anonymous visitor handle>",
  "email": "visitor@example.com"
}
```

Responses:
- `200` -> `{"ok": true, "attached": N}` (N = pending decision rows stamped; 0 is still ok)
- `404` unknown widget
- `429` rate limit (same limiter as chat; contact posts also count toward it)
- `401` bad/unknown/revoked embed key
- `403` disallowed origin, or widget not bound to the key's workspace/pocket
- `400` malformed `customer_ref`
- `422` `invalid_email` (RFC-ish syntax check, 254-char cap)

Gate order mirrors `/paw-bar/chat` via the shared `_front_gate_for_key`. The email is validated after authentication so refusal shapes stay fail-closed-first.

## How delivery works

`deliver_customer_decision` reads the row's prior state before flipping it. If the flip is a real PENDING -> decided transition and the row carries a `contact_email`, it sends exactly one email: subject `Update from <site name>`, body = the customer-facing reply (the operator's edited wording on approve, the decline reason on reject). A re-delivered approval (retry, sweep) finds the transition already spent and does not send again.

## Email transport

Nothing system-owned existed (the Gmail client is user-scoped OAuth), so this adds a deliberately small SMTP sender used only by this path: `POCKETPAW_SMTP_HOST` / `PORT` / `USER` / `PASSWORD` / `FROM`, STARTTLS on by default. Fail-soft by contract - no transport or a send failure logs a warning and returns; the approve/reject response and the on-page poll are never affected.

## PII posture

The address lives only on the `DecisionStatus` row (additive SQLite migration, same pattern as `workspace_id`). It never enters the Instinct proposal, the agent's context, the KB, transcripts, or the event log, and no public read returns it - the decision poll omits it by shape. The invariant is documented at the field definition.

## Tests

25 new in `tests/cloud/test_paw_bar_decision_contact.py`: every gate's refusal shape, attach stamps pending rows only (scoped per workspace and per visitor), the poll never carries the address, a stubbed sender proves one email on approve and on reject with no resend on replay, the real mailer's no-transport and send-error paths, and the legacy-DB column migration. Full paw-bar set: 293 green. Both import-linter configs kept.

Docs: `docs/concepts/concierge-knowledge.mdx` gains a "When the visitor leaves the page" section.

### DIFF
--- a/docs/concepts/concierge-knowledge.mdx
+++ b/docs/concepts/concierge-knowledge.mdx
@@ -134,3 +134,26 @@ decision about whether the site keeps a record of its visitors, not two.
 
 If the record cannot be read for any reason, the concierge answers without memory
 rather than failing the visitor's chat.
+
+## When the visitor leaves the page
+
+Some requests need a human decision, and humans take longer than a browser tab
+stays open. While a request is pending, the visitor can leave an email address
+(`POST /paw-bar/decision-contact`, guarded by the same key, origin and rate
+gates as the chat itself). When the owner approves or declines, the visitor is
+sent one email carrying exactly the reply the on-page poll would have shown —
+no more, no less — with a subject that names the site.
+
+The address is handled as narrowly as possible:
+
+- It is stored **only** on the pending decision rows it was left for. It never
+  enters the agent's context, the knowledge base, transcripts, or the proposal
+  the owner reviews, and no public endpoint ever returns it — the decision poll
+  response does not carry it.
+- One decision, one email. An approval that is re-delivered later (a retry, a
+  sweep) does not send again.
+- Email is best-effort on top of the poll, never instead of it. If no SMTP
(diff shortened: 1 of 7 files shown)


prakashUXtech:

PR #1819, opened 2026-07-30
Paw Bar: reply sources + public articles listing

prakashUXtech:
Stacked on #1818 (`feat/paw-bar-async-email`).

A concierge reply used to arrive with no trail — the visitor couldn't tell which page an answer came from, and there was no way to browse what the bar can quote. This adds both halves of visible grounding, per-reply attribution and a browsable library.

## Sources on each reply

`concierge_chat` now emits at most one `sources` SSE event, after the model stream completes and immediately before `stream_end`:

```
event: sources
data: {"sources": [{"title": "Opening Hours", "url": "https://brewco.pawsites.dev/hours"}]}
```

Max 3 entries, deduped by url, and only entries with both a title and a url. Nothing is emitted when nothing qualifies — no empty event.

Attribution is approximate on purpose (the Crisp-style "Sources" pattern): the server re-runs a cheap KB search of the visitor's message against the same `pocket:<pocket_id>` scope the run reads, keeps hits whose article ids are in `Site.kb_article_ids` (the page sync's own output — owner-uploaded files never surface), and maps the `site-<slug>` ids back to public page URLs. It is not a tool trace. The search runs concurrently with the model turn, so it adds ~0ms to the stream; any error or timeout drops the sources, never the reply.

## Public articles listing

`GET /paw-bar/articles?widget_id=...&signed_key=...` returns `{"articles": [{title, url, snippet}]}` — the site's synced KB pages, cap 20, snippet ≤160 plain-text chars. Same front-gate chain as chat (404 unknown widget → 429 rate limit → 401 bad key → 403 origin/binding), no injection screen since there's no free text. Empty KB is a 200 with an empty list, and a kb hiccup fails soft the same way.

To support this, `_front_gate_for_key` now resolves through `resolve_site_key_with_site` and returns the Site it already loaded (same pattern as the chat path), and `KnowledgeService` grows scope-form read methods (`search_articles_for_scope` / `list_articles_for_scope`) so the router doesn't reach into the private `_kb` wrapper.

## Tests

19 new tests in `tests/cloud/test_paw_bar_reply_sources.py` (event shape, cap + dedupe, synced-only filter, ordering vs `stream_end`, fail-soft on empty/erroring KB, and every articles refusal code). Full paw-bar set: 312 passed. Ruff + both import-linter configs clean.

Docs: `docs/concepts/concierge-knowledge.mdx` gains a "What the visitor sees as sources" section.

### DIFF
--- a/docs/concepts/concierge-knowledge.mdx
+++ b/docs/concepts/concierge-knowledge.mdx
@@ -81,6 +81,22 @@ so it syncs zero documents and reports `no_content`. Grounding one of those mean
 crawling the customer's own origin through the SSRF-hardened import crawler, and is
 not part of this path.
 
+## What the visitor sees as sources
+
+A reply can end with a short "Sources" list — up to three of the site's own pages,
+each a title and a public link. It is sent as a single `sources` event on the chat
+stream, just before the stream ends, and the same pages are browsable through a
+public articles listing (`GET /paw-bar/articles`, behind the same key, origin and
+rate gates as the chat; capped at twenty entries with a short snippet each).
+
+Attribution is approximate by design. The server re-runs a quick knowledge search
+of the visitor's question against the same pocket scope the run reads, and shows
+the synced pages that match — "what the knowledge base would ground this question
+on", not a verified trace of what the model actually quoted. Only pages the site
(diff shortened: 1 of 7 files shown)


prakashUXtech:

PR #1820, opened 2026-07-30
Concierge form cards: structured collection of gated-action details

prakashUXtech:
## What

A booking request came in as a prose blob — the concierge asked the visitor to "send name, phone, address" and the CRM got unusable free text. The widget spec already declares each gated action's typed args, so the concierge preamble now teaches a second `pawbar-card` kind, `form`:

```
{"kind": "form", "verb": "book_visit", "title": "Book a repair visit", "submit_label": "Send request", "fields": [{"name": "name", "label": "Name", "type": "text"}, {"name": "phone", "label": "Phone", "type": "tel"}]}
```

- `verb` must be a declared gated verb; every field `name` must be one of that verb's declared args (the preamble lists them per verb).
- `type` ∈ text | tel | email | number | textarea, derived by the agent from the arg: int/float → number; str → tel/email/textarea by name (`phone`, `email`, `message`/`notes`/`issue`), otherwise text. Bool args stay out of forms. The mapping travels in the prompt text — no server-side enforcement in this slice.
- Emit once, then wait: the widget submits the form and runs the action itself; the agent sees the outcome next turn. A message that already carries every detail skips the form and calls the tool directly, same as today.

## Scope

The block renders only when the widget declares gated actions with args. No-actions, auto-only, and gated-argless widgets keep today's preamble byte-for-byte.

## Tests

`TestFormCardPreamble` pins the contract fence, the per-verb arg list, the type-mapping text, emit-once/skip behavior, and absence for the three no-form cases. Full paw-bar set: 323 passed (`tests/cloud/test_paw_bar_*.py` + `tests/ee/test_paw_bar_customer_reply_router.py`).

Stacked on #1819: base is `feat/paw-bar-reply-sources`.

### DIFF
--- a/ee/pocketpaw_ee/cloud/surface/handlers/concierge.py
+++ b/ee/pocketpaw_ee/cloud/surface/handlers/concierge.py
@@ -1,5 +1,14 @@
 # concierge.py — /paw-bar surface preamble (the public concierge widget).
 #
+# Updated: 2026-07-30 (form cards) — when the widget declares GATED actions
+# that take args, the actions paragraph now teaches a second ```pawbar-card
+# kind: "form". Instead of asking the visitor for name/phone/etc. in prose,
+# the agent emits a structured form card (verb + fields drawn from the
+# action's declared args, with a name-based type mapping spelled out in the
+# prompt), waits for the widget to submit it, and only falls back to calling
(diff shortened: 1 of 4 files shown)


prakashUXtech:

PR #1821, opened 2026-07-30
Let a site concierge read its own agent's knowledge

prakashUXtech:
Paw Bar inbox **D5**. Scope is D5 only — no thumbs/revise-into-knowledge loop, no conversations state.

## The bug

`_kb_scopes_for_context` locked a `CONCIERGE` run to `['pocket:<site pocket>']` and deliberately dropped `agent:`. So an owner who opened their site's concierge in `/agents` and attached knowledge **to the agent** got nothing on the site — only the site→pocket page sync ever reached a visitor. No error, no warning; the concierge just kept saying it didn't know.

## What changed

**1. The scope.** A concierge run now reads `pocket:<site pocket>` **and** `agent:<its own agent id>`. `workspace:` and `user:` stay dropped.

The asymmetry is the whole security argument. Every site concierge is a dedicated agent, one per site (`ensure_site_agent`), so its `agent:` scope holds one site's material that the owner put there for these visitors. `workspace:` is the tenant-wide tier — every pocket, every other agent, every owner upload — and an anonymous caller must never reach it.

The agent id is not the caller's: it rides `RunSpec.agent_id` from `widget.agent_id`, and `_resolve_concierge` already proved that agent belongs to the site's workspace (`_agent_in_workspace`) before the run starts. No wiring was needed — `ctx.target_agent_id` is always populated for a concierge run, or resolution raises `concierge.no_agent`.

**2. The badge.** `GET /agents/{id}/knowledge` now returns `visible_to_site_visitors: bool` beside `items`. True when a paw-bar widget binds that agent to a published site — read from the real binding, not the `concierge-<site_id>` slug, so a hand-bound agent counts too. It labels a surface, it does not guard one: same owner/admin guard, same articles, and an unreadable store yields `false` rather than breaking the tab. Rendering the badge is the frontend's job.

`widget_for_agent(agent_id, workspace_id)` is the new reverse lookup in `agent_provisioning`; workspace-scoped, refuses an unscoped query, bounded scan.

**3. The docs.** `docs/concepts/concierge-knowledge.mdx` said the concierge reads one scope and cannot read `agent:`. It now names both scopes, explains why `agent:` is safe where `workspace:` is not, and carries the blunt rule as a warning: **anything attached to a site concierge is publishable by definition** — private material belongs on an agent no site binds.

## What a visitor can and cannot reach

| | after this change |
|---|---|
| `pocket:<the site's own pocket>` | **yes** (unchanged) |
| `agent:<the site's own concierge agent>` | **yes** (new) |
| `agent:<any other agent>` | no |
| `workspace:<the tenant>` | no |
| `user:<any member>` | no |
| anything in another tenant | no |

## Tests

The two existing assertions that pinned the old behaviour were **updated, not deleted**, and each carries the reasoning in its docstring. New isolation coverage guards the widened blast radius: a sibling agent's scope is never granted (unit and end-to-end through the resolver, with two agents in one workspace), the tenant-wide and member-private tiers are never granted, two tenants' scope sets are disjoint, and a half-resolved context degrades instead of emitting a malformed scope string. Nine more cover the visibility signal, including the wire shape and the failure-soft path.

917 passed, 2 skipped across the concierge, agent-service, chat, sites-KB and paw-bar suites. `ruff check` + `ruff format --check` clean; both import-linter configs pass.

## Note for the sibling branches

Two comments in files I stayed out of are now stale and want a one-line fix from their owner:
- `ee/pocketpaw_ee/paw_bar/router.py:2704` — "(KB locked to pocket:<id>)"
- `ee/pocketpaw_ee/paw_bar/router.py:2486` — "the one scope `_kb_scopes_for_context` grants a CONCIERGE run" (`_concierge_sources` itself is correct: it still searches the pocket scope alone, which is what should be citable)

### DIFF
--- a/docs/concepts/concierge-knowledge.mdx
+++ b/docs/concepts/concierge-knowledge.mdx
@@ -15,18 +15,49 @@ what it is allowed to know, and what it writes down about the people it talks to
 
 ## What a concierge is allowed to know
 
-A concierge run is locked to **one** knowledge scope: `pocket:<pocket_id>`, the
-pocket its site was published from. It cannot read the `agent:` scope or the
-workspace-wide scope. That is deliberate. The concierge answers anonymous callers
-from the public internet, and a public caller must never be able to reach the rest
-of the tenant's knowledge base by asking the right question.
+A concierge run reads exactly **two** knowledge scopes:
 
-The upshot is that a concierge's usefulness is decided entirely by what is in that
-one scope.
+- `pocket:<pocket_id>` — the pocket its site was published from, where the page
+  sync writes.
+- `agent:<agent_id>` — whatever the owner attached to this site's concierge agent
+  directly, in the agent's own Knowledge tab.
+
+It reads neither the workspace-wide scope nor any member's private scope, and that
+line is where the security argument lives. Every site concierge is a **dedicated**
+agent, one per site, created for that site and never shared — so its `agent:` scope
+is one site's own material, put there by the owner for these visitors. The
+workspace scope is the opposite: every pocket, every other agent, every owner
+upload in the tenant. A concierge answers anonymous callers from the public
+internet, so handing it the tenant-wide scope would mean a stranger could reach the
+rest of the business by asking the right question. That scope stays out.
+
+The agent it runs is not the caller's choice either. It comes from the site's own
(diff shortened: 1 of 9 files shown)


prakashUXtech:

PR #1822, opened 2026-07-30
Give the concierge inbox a conversation state row

prakashUXtech:
Slice 1 of the Paw Bar owner inbox. The concierge log becomes a queue.

Today the dashboard can show an owner every conversation their site's concierge has had, and nothing else. There is no way to say "I've dealt with this one", "hide this until tomorrow", or "this one needs me". That is the gap between a transcript viewer and a support inbox, and it is the first of the six slices in `docs/design/drafts/2026-07-30-paw-bar-inbox-unified-thread.md`.

## What's here

A thin state row per conversation, `paw_bar_conversations`, keyed on `(widget_id, customer_ref)` — the same identity the concierge run stream already uses. It holds lifecycle and operator metadata only: state, whether the bot is paused, a snooze deadline, tags, private notes, a contact address, an unread counter. **No messages.** The transcript stays derived from the run docs, so there is one record of what was said and a separate one of how the owner is handling it.

Endpoints, all additive — the existing shapes are frozen contracts the dashboard already consumes:

| | |
|---|---|
| `GET .../site/{id}/conversations` | every existing field unchanged; each item gains `state`, `bot_paused`, `unread_for_owner`, `tags`, `snooze_until`, `contact_email`, `display_name`, `has_pending_action`. Response gains `counts` and an optional `?state=` filter |
| `PATCH .../site/{id}/conversations/{customer_ref}` | `{state?, snooze_until?, tags?, note?, bot_paused?}` → `{ok, conversation}` |
| `GET .../agent/{id}/conversations` | the same question asked of the agent: its sites' conversations unioned, each naming where it came from |

## Three decisions worth reviewing

**Backfill-free, deliberately.** The row is minted on a visitor's next message, or on the owner's first action on an older conversation. Nothing was migrated. A conversation with no row still lists, reading as a plain open conversation — every new field carries a safe default. The one visible consequence: `counts` only counts rows that exist, so on an old site the chips can sum to less than the rows on screen. Inventing totals for conversations we have no state for seemed worse than that.

**A visitor's message re-opens their conversation.** Closed or snoozed, it comes back to `open` the moment they write again. Every product in this class does this, and its absence reads as a lost customer. `needs_human` is the exception — it's already the top of the queue, so re-opening it would be a demotion.

**Snooze expiry is computed on read.** One SQL expression that every filter, count and row detail goes through. No sweeper, no cron: nothing scheduled that can fail quietly at 3am and leave a customer waiting past their snooze.

## Notes for review

- The PATCH gates on `paw_bar.manage`. The design named `paw_bar.write`, which doesn't exist — `manage` is this router's existing mutation action (both ADMIN). Say the word if you'd rather add a third action.
- `has_pending_action` reads the decision rows the Decisions list already reads. No second query into Instinct, whose rows are keyed by the widget owner rather than the tenant.
- A captured contact email is read from the decision row where it was left, never copied onto the conversation row — the PII invariant on that field says it lives in one place, and this owner-authed list is exactly who it was left for.
- The visitor-turn write is failure-soft. A store hiccup logs and the chat streams normally; the queue is at worst one message stale.
- Tenancy is checked at every new read and write: the site (or the agent) resolves workspace-scoped first, then its widget, then the rows. A cross-workspace caller gets a 404 that leaks nothing, including on the agent endpoint.

## Verification

- `tests/cloud/test_paw_bar_conversations.py` — 37 new tests: the additive-ALTER migration path on a legacy DB, upsert create/touch/auto-reopen (and the `needs_human` exception), snooze expiry across the row, filter and counts, the PATCH per field plus note-append and the 422s, the agent union across two widgets of one agent excluding a third agent's, cross-workspace refusal on every new read and write, legacy rows listing with defaults, and the failure-soft upsert driven through a real `POST /paw-bar/chat`.
- Full paw-bar set green: 361 passed (324 before, 37 new).
- `ruff check` and `ruff format --check` clean; both import-linter configs pass.
- `docs/concepts/concierge-knowledge.mdx` gains a "conversation queue" section.

Slice 2 (owner reply, the `bot_paused` gate, the visitor poll) stacks on this branch.

### DIFF
--- a/docs/concepts/concierge-knowledge.mdx
+++ b/docs/concepts/concierge-knowledge.mdx
@@ -151,6 +151,62 @@ decision about whether the site keeps a record of its visitors, not two.
 If the record cannot be read for any reason, the concierge answers without memory
 rather than failing the visitor's chat.
 
+## The conversation queue
+
+Those stored turns make a log. A log tells the owner what happened; it does not tell
+them what still needs them. Each conversation therefore carries a small state row —
+identified, like the runs themselves, by the site's widget and the visitor's handle.
+
+The row holds lifecycle and operator notes only: a state, whether the bot is paused,
+a snooze deadline, tags, private notes, a contact address, and an unread counter. No
+messages. The transcript stays derived from the run documents, so there is exactly
+one record of what was said and a separate one of how the owner is handling it.
+
+Four states, and one rule that matters more than the rest:
+
+| State | Meaning |
+|---|---|
+| `open` | live and in the default view |
+| `needs_human` | escalated — the visitor asked for a person, or the bot could not answer |
+| `snoozed` | hidden until a deadline the owner set |
+| `closed` | done |
+
+**A visitor's message re-opens their conversation.** Closed or snoozed, it comes
+back to `open` the moment they write again — a returning customer must land in the
+queue, not in an archive. The one exception is `needs_human`: that is already the
+top of the queue, so "re-opening" it would be a demotion.
+
+The row is created on a visitor's first message, and on the owner's first action on
+an older conversation. Nothing was backfilled when this shipped, and nothing needs
+to be: a conversation with no row still lists, reading as a plain open conversation.
+
+Snooze expiry is decided when the queue is read, not by a background sweep. A snooze
+that has passed shows up as open wherever the owner looks — the list, the filters,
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1823, opened 2026-07-30
Fix the build-timeout comments for the dynamic sites lane

prakashUXtech:
Found while auditing the dynamic (D1-backed) and svelte Paw Sites lanes on the local rig.

`generator_client.py` told two readers, in two places, that "a legit build is ~45-60s
and a wedged one runs for minutes, so the 120s default cleanly separates them."

That is true for a static site. It is not true for a dynamic one. Measured on the rig
against a 1-table dynamic ripple pocket (one read source, one write action):

    bun run build   4m47s   (.svelte-kit/output/server/chunks/_page.js ~2.6 MB)

A dynamic page is SSR'd (prerender off) so the whole Ripple widget bundle lands in the
server chunk. At the 120s default the gate kills that healthy build and reports it as
`build timed out after 120s - killed the build (likely a wedged workerd prerender)`,
which fails the `provision_site` job on a site that had nothing wrong with it.

This commit records the measurement and the operator note next to the constant. It does
not change the default: the static publish path is request-synchronous and cannot afford
a longer hang, so a deployment that publishes dynamic sites raises
`PAW_SITES_BUILD_TIMEOUT_SEC` (600 leaves headroom on a cold `node_modules`). A
dynamic-aware timeout, given the provision job is already asynchronous, is worth doing
but is a behaviour change and belongs in its own slice.

Comment-only, no behaviour change. `tests/ee/sites` + `tests/cloud/sites` green
(553 passed; the one combined-run failure is the known `test_site_billing.py`
isolation flake, which passes on a clean run). ruff clean, both import-linter
configs pass.

### DIFF
--- a/ee/pocketpaw_ee/sites/generator_client.py
+++ b/ee/pocketpaw_ee/sites/generator_client.py
@@ -8,6 +8,13 @@
 # behind a _runner so the orchestration is unit-testable without Bun/workerd.
 # Created: 2026-05-30 (feat/paw-sites-backend, Task 2.3).
 #
+# Updated 2026-07-30 (fix/sites-d1-svelte-audit): corrected the build-timeout
+# comments. They claimed "a legit build is ~45-60s", which is true for a static site
+# and false for a dynamic one — a measured 1-table dynamic ripple build takes 4m47s,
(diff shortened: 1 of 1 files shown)


prakashUXtech:

PR #1824, opened 2026-07-30
A brand-new site shipped without its concierge

prakashUXtech:
A first publish went live with no concierge bar on it, silently.

`_embed_concierge_bar` only provisioned when a Site doc already existed, but
on a first publish that doc is inserted *after* the embed step. So the widget
and dedicated agent were never created, the snippet check returned empty, and
the page shipped bar-less — with no log line, because an empty snippet returns
early. A re-publish grew a bar normally, which is why the earlier publish-time
provisioning fix looked correct: it was verified by republishing an existing
site.

Now the first pass builds a transient Site to provision against. Provisioning
only reads `workspace` / `owner` / `id` / `name` / `pocket_id` off the object
and never re-reads the database, and the insert further down carries the same
values.

Tests cover the first-publish path and that a second publish adopts the same
widget instead of minting a sibling. 525 sites tests pass.

Found by the dynamic-sites audit: it published a fresh site and found zero
paw-bar markers in the artifact.

### DIFF
--- a/ee/pocketpaw_ee/sites/service.py
+++ b/ee/pocketpaw_ee/sites/service.py
@@ -2021,6 +2021,10 @@ async def _deploy_site_doc(
         signed_key=signed_key,
         project_dir=build.project_dir,
         engine=engine,
+        # A FIRST publish has no Site doc yet — it is inserted further down — so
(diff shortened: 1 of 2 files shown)


prakashUXtech:

PR #1825, opened 2026-07-30
Type to take over: the owner replies and the bot stops talking

prakashUXtech:
Slice 2 of the Paw Bar owner inbox, stacked on #1822 (`feat/paw-bar-conversations`).

Slice 1 made the concierge log a queue. This makes it a place the owner can answer from, and makes the bot get out of the way when they do. A bot that keeps answering over a human is the loudest complaint in every review of every product in this class, so the muting is the part to read closely.

## What lands

**`POST /paw-bar/admin/site/{site_id}/conversations/{customer_ref}/reply`** — `{text}` → `{ok, message, conversation}`. One call stores the owner's line, mutes the bot, stamps `last_owner_at`, clears the unread counter, and reopens a closed or snoozed thread. Gated on `paw_bar.manage`, workspace-scoped at the same two seams as the PATCH. `message` is a `TranscriptMessage` (the shape the thread already renders); `conversation` is the same `ConversationRow` the PATCH echoes.

**The mute in `concierge_chat`** — checked before the run is created. A paused conversation gets no agent turn, nothing metered, no tool surface. The visitor's line is still kept (under the same retention toggle the run path honours), the thread moves to `needs_human`, and the response is exactly two SSE frames:

```
event: human_replying
data: {"message": "Someone from the team is replying — hang tight."}

event: stream_end
data: {"assistant_message_id": null, "cancelled": false}
```

Never an empty stream: the glass app reads clean-but-empty as "No reply.", so a muted bot would look like a broken one.

**`GET /paw-bar/messages/{widget_id}/{customer_ref}?signed_key=&after=`** — the visitor side, behind the same front gate as the articles listing (404 → 429 → 401 → 403) and the same `_request_origin` handling for same-origin GETs. Returns `{"messages": [{role, content, at}], "bot_paused": bool}` and nothing else. Notes, tags, assignee, captured email and queue state stay owner-only, and a visitor's own stored lines are not echoed back. A `…Z` cursor from `Date.toISOString()` is normalized before it is compared, or the poll would skip the message it just delivered.

**Idle auto-resume** — four hours, hard-coded. A mute with no owner activity for that long reads as un-paused everywhere, decided on read like the snooze expiry, so the chat endpoint and the poll can never disagree. Materialized with one conditional UPDATE plus a single system line explaining the hand-back.

**Transcript merge** — `GET .../conversations/{customer_ref}` interleaves both sources by timestamp. `TranscriptMessage.role` widens from `user|assistant` to `user|assistant|owner|system`, additively.

## Why owner turns are not run docs

`metering/sweeper.py::sweep_unbilled_runs` bills every unbilled terminal chat run. An owner reply shaped as one would charge the owner credits for typing their own sentence and appear in run analytics as agent compute that never happened. So they live in a new `paw_bar_owner_messages` table beside `paw_bar_conversations`, same additive-migration and workspace-scoping conventions. Timestamps there are aware UTC so they sort against `ChatRunDoc.createdAt` on one clock rather than drifting by the host's offset.

## Tests

`tests/cloud/test_paw_bar_takeover.py`, 49 tests. The centrepiece asserts a paused visitor turn emits `human_replying`, emits no chunk frames, dispatches nothing to the executor, and creates no `ChatRunDoc` — with the **real** `create_run` wired, and a paired control test proving the same call does create one when the bot is not paused. Otherwise "no run doc" would only be proving the test stubbed it away.

Also covered: the poll's exact key allowlist and every refusal code, `after` filtering including the JS-style cursor, the four-hour resume and its single system message, a fresh pause not resuming, all four roles interleaved by timestamp, an assistant-only thread still rendering, and cross-workspace refusal on every new surface.

Full paw-bar set: 411 passing (362 on the base). `ruff check`, `ruff format --check`, and both import-linter configs clean.

## Also in here

Two comments in `router.py` went stale when D5 shipped (#1821): a concierge run now reads `pocket:<site>` and `agent:<its own id>`, never `workspace:` or `user:`. Wording corrected. `_concierge_sources` still searches only the pocket scope, and that is right — a source has to be a link the visitor can open, which only the page sync's articles have.

The concierge concepts doc gains a "When a human takes over" section.

## Not in here

Handoffs, notifications, in-thread approval cards, and the frontend. `GET /paw-bar/admin/site/{id}/decisions` still returns no `customer_ref`, so a thread cannot yet attribute a pending action to the visitor it belongs to — that is slice 4's to add.

### DIFF
--- a/docs/concepts/concierge-knowledge.mdx
+++ b/docs/concepts/concierge-knowledge.mdx
@@ -207,6 +207,74 @@ Visitors are shown by the email they left, when they left one, and otherwise by 
 short stub of their anonymous handle. The email is still stored only where it was
 captured; the owner's list reads it there rather than keeping a second copy.
 
+## When a human takes over
+
+Typing is taking over. There is no separate button to press first, because the
+failure this is built to prevent is a bot and a person answering the same customer
+at the same time — and a button is a thing you can forget.
+
+`POST /paw-bar/admin/site/{site_id}/conversations/{customer_ref}/reply` sends the
+owner's line, and does everything else that has to happen with it: the reply is
+stored, the bot is muted, the conversation's unread counter clears, and a closed or
+snoozed thread comes back to `open`. An owner answering a customer is engagement;
+leaving that thread filed while a human types into it is how a conversation
+disappears mid-sentence. Like the PATCH, it gates on `paw_bar.manage`.
+
+An owner's reply does **not** go through the approval loop. Instinct gates actions
+the agent proposes, because a machine wants a side effect and a human should decide.
+A human typing a sentence already is the decision.
+
+### The bot stops talking
+
+While a conversation is paused, a visitor's message does not reach the agent at all.
+The check happens before any run is created, so there is no agent turn, nothing
+metered, and no tools in reach. What the visitor gets back is a distinct
+`human_replying` event on the chat stream saying a person is replying, followed by
+the normal end-of-stream event. It is never silence: an empty stream reads as a
+failure, and a muted bot must not look like a broken one.
+
+The visitor's message is still kept, and their conversation moves to `needs_human`.
+Someone typing to a paused bot is by definition waiting on a person, and the owner
+has to be able to read what they said while they were typing. Retention applies here
+exactly as it does elsewhere: with `concierge_store_transcripts` off, that line is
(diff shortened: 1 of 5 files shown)


prakashUXtech:

PR #1826, opened 2026-07-31
Paw Bar backend: the whole concierge line plus the owner inbox

prakashUXtech:
One reviewable branch carrying every open backend PR for the concierge line.
Opening it does not merge anything — it exists so twenty stacked PRs can be read
and approved in one place.

**Folded (each proven contained with `git merge-base --is-ancestor`):**

Ship stack — #1714 concierge seam · #1715 decision-loop tenancy · #1716 frame +
CSP · #1728 soul-write policy · #1731 action registry · #1739 kill switch and
greeting · #1740 owner aggregation · #1775 the wave-1 integration.

Per-site concierge — #1776 dedicated agent · #1797 knowledge sync and
both-sides transcripts · #1816 memory · #1817 autoembed.

Tonight — #1818 email a departed visitor their decision · #1819 reply sources
and public articles · #1820 form cards for gated actions · #1821 the concierge
reads its own agent's knowledge · #1822 the conversation queue · #1825
type-to-takeover · #1823 dynamic-lane doc correction · #1824 a brand-new site
shipped without its concierge.

**Bugs found by driving the live rig, not by the suites:**

- A first publish went live with no bar at all and no log line. The earlier fix
  looked correct because it was verified by *re*-publishing.
- Filing an operator note 500'd for every real owner, and so did the first owner
  reply: the authenticated principal is a `PydanticObjectId` where a `str` was
  expected. That seam has now bitten this layer four times, always with a green
  suite, because every unit test hands in a plain string — so the coercion moved
  onto the models.
- A concierge turn was writing an orphan `Message` keyed by a pocket id into the
  group-room space. Nothing read it; it is gone.

**Verification:** 1048 tests green across the paw-bar, chat and sites suites;
ruff and both import-linter configs clean; and `scripts/inbox-smoke.sh` drives
the full visitor↔owner loop against the running rig — 16 checks, 0 failures,
including the one that matters most: while a human is replying, the assistant
emits `human_replying` and produces no content.

Frontend counterpart: paw-enterprise #714.


prakashUXtech:
Updated — **slice 3 (the escape hatch)** is folded in. `GET .../handoffs` has existed since the dashboard shipped and always returned `[]` because nothing produced a handoff; now a built-in `pawbar_request_human` tool rides every concierge run bound to a widget — not a verb each site has to remember to declare — writes the `_paw_handoffs` object the existing read consumes, flips the conversation to `needs_human`, and notifies the owner. It executes nothing else, so CONCIERGE scope stays zero-authority. Notifications fire on a new conversation, a handoff, and a visitor reply while the bot is paused, all fail-soft. **1094 tests green**; ruff and both import-linter configs clean.

prakashUXtech:
### Code review

Found 4 issues. All are fixed on the branch; links point at the fixed lines so the fix is reviewable next to the finding.

1. The type-to-takeover mute failed OPEN. The `bot_paused` read and the failure-soft bookkeeping write shared one `try/except`, so any store error left `conversation = None` and the gate below read as "not paused" — the bot answered on top of a human who had taken the conversation over. A read failure has to mute, not unmute. Split into two blocks with an explicit `mute_unreadable` arm, fixed in 61ad21f622a4df9f82b84dc6a4d40060748b6277 with two tests that both fail against the original shared block.

https://github.com/pocketpaw/pocketpaw/blob/89e3ccb11cb2d3a189a5453329aecc61d7b62d0c/ee/pocketpaw_ee/paw_bar/router.py#L3742-L3744

2. Gated-action proposals were assigned to the workspace, not the owner. `propose_customer_action` passed `assignee=ws` — the same value it already passed as `workspace_id`. The Tray's assigned filter matches an operator identity and a workspace id is not one, so a form-card submit was hidden from it while an identical ingest event (which has always assigned the owner) showed up. The module's own header states the invariant: "The widget OWNER is a within-tenant human label used as the assignee". Store routing was right on both paths, which is why the existing tenancy tests stayed green over it.

https://github.com/pocketpaw/pocketpaw/blob/89e3ccb11cb2d3a189a5453329aecc61d7b62d0c/ee/pocketpaw_ee/paw_bar/decision_loop.py#L464-L466

3. Four file headers were stale after behavior changes (CLAUDE.md says "When you update/make changes to a file, make sure to rewrite the comment at the top of the file with the changes you made"). `message_service.py` did not mention that a concierge turn now persists no `Message` at all — which anything addressing a message by id needs to know. `sites/service.py` and `models/site.py` were missing both the bounded single-flight guard and the first-publish concierge fix. `paw_bar/models.py` still described the KB scope as a single pocket scope; it has been a pocket + own-agent union since D5. `agent_provisioning.py` said two triggers funnel into `ensure_site_agent`; there are three.

https://github.com/pocketpaw/pocketpaw/blob/89e3ccb11cb2d3a189a5453329aecc61d7b62d0c/ee/pocketpaw_ee/cloud/models/site.py#L6-L16

4. No API docs for the new surface (CLAUDE.md says "New runtime API -> `docs/api-reference.md`" and "Never let docs drift behind code"). The PR adds the public visitor endpoints and the whole owner-inbox admin surface with no reference entry. Added as its own section.

https://github.com/pocketpaw/pocketpaw/blob/89e3ccb11cb2d3a189a5453329aecc61d7b62d0c/docs/api-reference.md#L1-L10

One more finding scored below the reporting bar but is worth a follow-up rather than a fix here: `InstinctStore.reject()` passes no `require_status`, so unlike `approve()` it has no atomic compare-and-swap. Two near-simultaneous rejects on one action can both reach `deliver_customer_decision`, and the decision loop's own prior-state read is not a CAS either, so both could send the decline email. The root gap is upstream in `src/pocketpaw/instinct/store.py`, not in this PR.

PR #1827, opened 2026-07-31
A lost provision job no longer bricks the pocket forever

prakashUXtech:
The dynamic single-flight guard was unbounded: a job that was never consumed (no worker running) or died without a terminal status left the Site in `provisioning`, and every later publish of that pocket became a silent no-op returning 200 while doing nothing. Two demo sites were stuck exactly that way — showing as Draft and never rendering, which is how it was found.

Sites now stamp `provision_started_at`, and a guard older than 30 minutes is treated as a lost attempt: log and re-enqueue. A real provision measures ~5 minutes, so the window is generous; an undateable row reads as stale because the failure modes are asymmetric — a redundant enqueue costs one idempotent job, a stuck guard costs every future publish.

557 sites tests pass (the one failure is the known pre-existing `test_site_billing` isolation flake, which passes alone); ruff and both import-linter configs clean.


prakashUXtech:
Folded into `integration/paw-bar-inbox` (#1826).

PR #1858, opened 2026-08-04
One uncached config read was the whole prompt-assembly cost

DevRohit06:
Stacked on #1857. Merge that first.

Chasing "why is `hello` slow", measured on a live local instance rather than reasoned about.

## `build_behavior_instructions`: 120.9 ms → 0.0 ms

It builds 36,608 chars a turn, so it looked like slow string assembly. Profiling said otherwise — **11.9 million Python function calls for five invocations**, almost all under one line:

```
build_behavior_instructions
  └ composio.service.is_enabled()      ← called with no settings
     └ Settings.load()                 114.9 ms, measured
        └ pydantic_settings re-parses the whole model from .env
```

One uncached config read **was** the entire cost of assembling the system prompt. `get_settings()` is the `lru_cache`d accessor and measures 0.000 ms.

The irony: `is_enabled`'s own docstring warned about it — *"accepts an optional `Settings` so callers that already have one don't pay the `Settings.load()` cost twice"* — while its default path did exactly that. Fixed there, in composio's other read-only helpers, and at the two `Settings.load()` calls on the per-run path in `run_core`.

**Why this is safe.** The invalidation contract already exists — every config-write path calls `get_settings.cache_clear()` (`api/v1/settings`, `budget`, `memory`, `bus/commands`, `dashboard`). An explicit `settings` argument still wins, so nothing that already passed one changed behaviour.

The composio tests had been getting isolation for free from `Settings.load()`'s freshness; they now ask for it explicitly via a conftest. They passed individually and failed together, which is how it surfaced.

## KB snippets: 50.2 ms → 18.5 ms, and no longer unbounded

The per-scope searches ran in a **serial loop**, each one a `kb` subprocess. Two *empty* scopes still cost 50.2 ms, because the floor is process spawn and the loop paid it twice in sequence. Now gathered, order preserved, so the rendered block is byte-identical.

They were also unbounded, which matters more: a workspace scope holding **4,051,312 words took 4.2 seconds per search**, every turn, on a message that was just "hello". Time was flat across queries — `"a"` and a six-word question cost the same — so it's a scan, not a lookup.

That index is being fixed in kb-go by someone else. **The cap stays either way**, because it bounds *any* slow scope (huge corpus, wedged binary, stalled mount), not just that one cause, and it degrades to "no KB block" — exactly what an empty scope already produces.

## Net

| | before | after |
|---|---|---|
| cloud-side prompt assembly, warm | 173.3 ms | **24.1 ms** |

## What's deliberately not in here

The backend half is already fine warm — 17 ms from `run()` to the first byte at LiteLLM. But a **cold process costs 3193 ms**: `build_mcp_tools` 1223 ms, `build_model` 876 ms. A new agent instance in a warm process is ~330 ms. Prewarming those at startup is the obvious next move and is a separate change.

## Tests

Structural invariants, not stopwatches. "Assembling a prompt must not re-parse the config" is exact and names the defect when it fails; an "under N ms" assertion flakes on CI and explains nothing. The one timing assertion that does exist (serial vs concurrent) has a 4x margin.

10 new tests. 5 mutations in `tests/mutations/prompt_latency.json`, all caught — including reverting `is_enabled` to `Settings.load()`, removing the timeout, tuning the timeout down into a real budget, and restoring the serial loop.

`tests/cloud/chat/` + `tests/cloud/composio/`: 204 passed, 0 failed.

### DIFF
--- a/ee/pocketpaw_ee/cloud/chat/agent_service.py
+++ b/ee/pocketpaw_ee/cloud/chat/agent_service.py
@@ -2052,6 +2052,22 @@ def _member_private_user_scope(ctx: ScopeContext) -> str | None:
 # ~400 tokens ≈ 1600 chars (English ≈ 4 chars/token); we cap on chars (a
 # cheap, deterministic proxy — no tokenizer dependency) and truncate with an
 # ellipsis if a rendered block would exceed it.
+# How long one KB scope search may take before the turn gives up on it.
+#
+# 1.5s is a backstop, not a target. A healthy scope answers in ~25 ms (process
+# spawn); the number exists for the unhealthy case, and the unhealthy case is
+# real: a workspace holding 4,051,312 words measured 4.2 SECONDS per search on
+# 2026-08-04, on every turn, because kb-go scans instead of indexing. Search
+# time there was flat across queries — "a" and a six-word question cost the
+# same — which is the signature of a scan. That index is being fixed in kb-go
+# separately; this cap stays regardless, because it bounds any slow scope
+# rather than that one cause.
+#
+# Set high enough that a slow-but-working scope still contributes, low enough
+# that a pathological one cannot own the turn. Exceeding it drops the KB block,
+# which is the same outcome as a scope with no hits.
+_KB_SEARCH_TIMEOUT_SECONDS = 1.5
+
 _BRIEFING_MAX_CHARS = 1600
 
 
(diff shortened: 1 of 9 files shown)


DevRohit06:
Broader verification, since `get_settings()` caching is cross-cutting and the targeted run only covered chat + composio.

Ran every `tests/cloud/` subdirectory that touches settings or env — agents, auth, credits, llm_provisioning, memory, realtime, runs, uploads — on both this branch and its base, and diffed the failure sets rather than the counts:

```
base   26 failed, 1048 passed
perf   25 failed, 1049 passed

regressions (failing here, passing on base):   none
```

The one difference is `test_run_core.py::test_execute_run_cancellation_preserves_original_exception`, which fails at base and passes here. That is **not** a fix — it passes in isolation on both branches, three runs each, so it is order-dependent and something in the suite was tripping it. Flagging it as pre-existing flakiness rather than claiming credit.

`tests/cloud/` in full takes over 10 minutes and I did not get a complete run; it was at 42% with zero failures when it timed out. The subset above is the part that could plausibly be affected by the cache change.

PR #1859, opened 2026-08-04
fix(kb): compile via agent backend, reject uncompiled ingests, guard chat search

prakashUXtech:
## The bug this fixes

The cloud box runs PocketPaw on the Claude Code agent backend with no `ANTHROPIC_API_KEY`. `KnowledgeService.ingest_text_to_scope` shells out to `kb ingest`, whose internal LLM compile needs that key. Without it, the compile failed on every single ingest — and kb quietly stored each document verbatim as an "article". One workspace accumulated a 54-article / 4M-word scope this way, and `_build_kb_snippets_block` runs a kb search on every chat turn, paying about 5 seconds of O(corpus) subprocess time to retrieve 522 chars of junk.

The failure was silent at every layer: kb didn't error, the listener logged success, and chat just got slower and dumber.

## What changed

**Ingest compiles through our own backend when the key is absent.** `ingest_text_to_scope` now branches: with an `ANTHROPIC_API_KEY`, the plain `kb ingest` path is unchanged. Without one, the article is compiled by `PocketPawCompilerBackend` (`ee/pocketpaw_ee/cloud/kb/backend_adapter.py` — the agent-registry adapter built for exactly this, previously unused), then handed to kb pre-compiled via the new `kb ingest --article-json` stdin contract. kb makes no LLM call of its own on that path. An unavailable backend raises a distinct "agent backend not available" error rather than surfacing as an empty response.

**Compile failure raises. Always.** Garbage output, empty title/content, or a "compiled" article that is basically the input echoed back (large doc, content > 60% of what the compiler saw) all raise instead of ingesting. Callers already contain ingest exceptions per their existing patterns (the FileReady listener logs and skips, the site sync counts a skip), so the failure surfaces in logs instead of poisoning the scope.

**The REST ingest routes go through the same funnel.** `POST /kb/ingest/text` and `POST /kb/ingest/url` used to call `_kb("ingest", ...)` directly — no agent-backend compile, no fallback rejection, and a sync subprocess call blocking the async handler. Both now route through `ingest_text_to_scope`, so every ingest path in the codebase shares one hardened door.

**Fallback-marker rejection, defense in depth.** Any ingest result with `compiled_with == "none (fallback)"` — an older binary, or `--allow-fallback` misuse — is rejected with a warning naming the scope and article id. At the REST door this surfaces as the route's 500, never a stored article.

**Chat turns can't be stalled by the KB anymore.** `search_context_for_scope` gets a hard 5s timeout and fails soft: on timeout or subprocess failure it logs a warning (scope + elapsed) and returns `""`, so the KB block is skipped and the turn proceeds. `_kb` now also converts `subprocess.TimeoutExpired` into a clear `RuntimeError`.

**Text-file ingest unified.** `ingest_file`'s text/code path used to hand kb a file path directly, bypassing all of the above; it now reads the file and routes through `ingest_text_to_scope`.

## Deploy coupling

This code targets the kb-go CLI contract shipping in the parallel kb-go PR: `kb ingest` without `--allow-fallback` exits 1 on compile failure (no article), and `kb ingest --article-json --scope S --json` accepts a pre-compiled `{raw_text, article}` payload on stdin. The binary and this change ship to the box together (runbook covers deployment).

**Old-binary detection** (corrected after a live smoke against real binaries): kb-go parses flags by hand and silently ignores unknown ones, so an old binary never errors on `--article-json` — it reads the JSON payload from stdin as raw text, stores the wrapper verbatim via its keyless fallback, and exits 0 with old-style output. The version-proof signal is the `compiled_with` key: the paired binary always emits it on this path, so a result without it means the flag was ignored. The funnel raises with the upgrade hint and names the already-stored article (with the `kb delete` command) so the operator can purge it. The unknown-flag remap remains as belt-and-braces only.

## Tests

`tests/cloud/agents/test_knowledge_ingest_hardening.py` (11 tests) spies at the subprocess boundary — real `_kb`, captured argv + stdin — so the kb CLI contract itself is what's pinned. Only the LLM completion is faked. Covers: the exact `--article-json` argv and payload shape, the key-present path staying byte-identical, compile garbage/empty/echo raising with zero kb calls, the old-binary case modeled as it really behaves (exit 0, old-style JSON, no `compiled_with` → raise naming the article), fallback-marker rejection with the warning, the `ingest_file` text-path reroute, and both search fail-soft cases.

`tests/cloud/kb/test_router_ingest_funnel.py` (5 tests) drives both REST routes through the real funnel to the subprocess boundary: keyless → `--article-json` argv, key present → plain-ingest argv, never a bare direct call; the fallback marker and the old-binary missing-`compiled_with` case both mapping to the route's `kb.ingest_failed` 500 (inherited from the funnel, not re-implemented).

Consumer suites (uploads listeners, kb purge, kb router/service, site kb ingest, paw bar sources, library verbs, cloud chat) are green; the pre-existing failures in `tests/cloud/uploads/test_router*` and `test_agent_visibility_enforcement.py::test_knowledge_list_allowed_for_owner` reproduce identically on a clean checkout of this base. `lint-imports` passes on both configs.

Docs: added a compilation section to `docs/concepts/concierge-knowledge.mdx`, including the 80K compile-input cap and the small-doc exemption from the echo check.

### DIFF
--- a/Dockerfile.enterprise
+++ b/Dockerfile.enterprise
@@ -77,7 +77,7 @@
 # `go install` on the host. Pinned (not @latest) for reproducible builds; bump
 # KB_GO_VERSION to upgrade.
 FROM golang:1.25-bookworm AS kb
-ARG KB_GO_VERSION=v0.1.0
+ARG KB_GO_VERSION=v0.2.0
 # Produces /go/bin/kb-go; we expose it as `kb` (the name the resolver prefers).
 RUN go install github.com/qbtrix/kb-go@${KB_GO_VERSION} && \
     cp /go/bin/kb-go /usr/local/bin/kb
--- a/docs/concepts/concierge-knowledge.mdx
+++ b/docs/concepts/concierge-knowledge.mdx
@@ -87,6 +87,17 @@ to the kb binary. If compilation fails on either path, the ingest **fails loudly
 — the document is never stored verbatim, because uncompiled raw text poisons the
 scope and makes every chat turn pay full-corpus search cost for junk snippets.
 
+Two bounds on the agent-backend path are worth knowing:
+
+- **The compiler sees at most the first 80,000 characters** of a document. The
+  full raw text is still stored alongside the article, but for a very large
+  document the compiled article reflects that first slice only.
+- **The compression sanity check applies to documents over 4,000 characters.**
+  A compiled article for a large document must be meaningfully shorter than the
+  input the compiler saw, or it is rejected as a verbatim echo. Short documents
+  are exempt — an article about a two-paragraph note can legitimately be about
+  as long as the note itself.
+
 ### When it runs
 
 - **On publish.** The site's content just changed, so what the concierge can quote
--- a/docs/concepts/concierge-knowledge.mdx
+++ b/docs/concepts/concierge-knowledge.mdx
@@ -76,6 +76,17 @@ Pages that extract to almost nothing (an empty layout, a redirect stub) are skip
 rather than indexed, and a very large site is truncated rather than allowed to turn
 one publish into an unbounded run of article compilations.
 
+### Every article is LLM-compiled — never stored verbatim
+
+Ingest always compiles the extracted text into a structured article (title, summary,
+compressed wiki-style content) before it is stored. On deployments with an
+`ANTHROPIC_API_KEY`, the kb binary runs that compilation itself. On deployments
+without one (for example the Claude Code agent-backend cloud box), PocketPaw
+compiles the article with its own agent backend and hands the pre-compiled article
+to the kb binary. If compilation fails on either path, the ingest **fails loudly**
+— the document is never stored verbatim, because uncompiled raw text poisons the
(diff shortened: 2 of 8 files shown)


prakashUXtech:
### Code review

Five-lens review (CLAUDE.md compliance, bug scan, git history, prior-PR feedback, code-comment contracts) across this PR and the paired kb-go #31/#32, with confidence scoring. One issue met the report threshold:

1. PR title failed the Conventional Commits gate (bot flagged it as must-fix; workspace CLAUDE.md mandates the format). Fixed — retitled to `fix(kb): compile via agent backend, reject uncompiled ingests, guard chat search`.

Sub-threshold but verified findings were fixed on the branch rather than reported: code files piped over stdin lost kb-go's AST parse (restored via `--lang` + structure-aware keyless prompt, d24a3e4e), a transitional-shim `CloudError` import in a new test (canonicalized), and two docstring over-claims (corrected). Verified live: keyless agent-backend compile, old-binary loud failure, 5s chat-turn fail-soft.

PR #1862, opened 2026-08-04
feat(kb): knowledge REST surface for the living-wiki rebuild

prakashUXtech:
## What this does

The /knowledge frontend is being rebuilt as a living wiki, and the old aggregator endpoint didn't carry enough for it to render anything beyond a title list. This PR extends the read surface and adds reingest, all as thin wrappers — the heavy lifting stays in the hardened `KnowledgeService` funnel that landed earlier today (#1859).

### Extended
- `GET /knowledge/articles` — rows now carry `summary`, `word_count`, `compiled_with`, `version`, `categories`, `concepts`, `compiled_at`. `kb list --json` only emits the first three, so the rest come from the article's own wiki frontmatter (the same on-disk format the orphan scan already reads). `updated_at` falls back to `compiled_at` so newest-first sorting works on real data. Existing keys and the `{articles, total, agent_ids}` shape are unchanged.
- The per-scope kb subprocess now runs off the event loop; it used to block the handler for the whole aggregation.

### New
- `GET /knowledge/articles/{id}?scope=` — full article via `kb show` plus `content`, `backlinks`, `compiled_at`, `source_docs`. Orphan raw docs (ingested, never compiled) return `orphan: true` with the raw text, so the rows the listing surfaces are all clickable. Unknown id or scope → 404.
- `GET /knowledge/stats` — `kb stats` per scope visible to the caller (workspace + its agents), skipping scopes that fail instead of 500ing.
- `POST /knowledge/reingest` — re-runs an article's linked raw doc through `ingest_text_to_scope`. Accepts either a compiled article id (frontmatter `source_docs` names the raw doc) or an orphan raw-doc id directly.
- `POST /knowledge/reingest-upload` — the synchronous sibling of the FileReady listener: resolve the upload (local or S3 via temp file), extract through the configured chain, funnel with the original filename as source, stamp the FL-11b tracking columns. One upload per call; hidden files are refused.
- `GET /knowledge/uploads` — files eligible for ingest with a cheap `has_article` marker: the FL-11b `kb_article_id` column matched against the resolved scope, with a filename-vs-article-sources fallback for uploads indexed before that column existed. No new tracking built.

### Security
Every scope-accepting route binds the client scope through the /kb router's allowlist (`validate_scope_override`), with denials audit-logged at ALERT. Direct file reads guard against path traversal the same way kb-go's `containedID` does, answering 404 rather than acting as an oracle. `hide_from_ai` files are excluded from the uploads listing and refused by reingest-upload.

## Tests

`tests/cloud/test_knowledge_wiki_api.py` (16 tests) fakes kb at the subprocess boundary so the real funnel runs — the reingest argv is pinned to the funnel's ingest shape, same pattern as `tests/cloud/kb/test_router_ingest_funnel.py`. Covers each endpoint's happy path, the scope-guard denial, the 404s, the orphan fallback, and reingest-upload with a faked adapter + extraction chain against a real temp KB home.

```
uv run pytest tests/cloud/test_knowledge_wiki_api.py tests/cloud/test_knowledge_router.py \
  tests/cloud/test_knowledge_aggregator.py tests/cloud/kb tests/cloud/test_site_kb_ingest.py \
  tests/cloud/files -q
158 passed
```

`lint-imports` (core and ee configs) and ruff are clean. The 23 failures in `tests/cloud/uploads/` are pre-existing on origin/dev (401s from the auth env, identical with the branch stashed).

Contract shapes are documented in `docs/api-reference.md` (Knowledge — Living Wiki API). Design doc: `docs/design/drafts/2026-08-04-knowledge-wiki-redesign.md` in the workspace repo.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -1795,7 +1795,9 @@ Response is the row shape above plus `content` (markdown), `backlinks`
 (list of article ids), `source_docs` (raw-doc ids), `scope`, and `orphan`.
 An orphan raw-doc id returns `orphan: true` with the raw text as `content`
 and `compiled_with: null`. Unknown id or an id outside the scope →
-`404 article.not_found`.
+`404 article.not_found`. A kb failure that is NOT a genuine miss (timeout,
+missing binary, transient error) → `500 knowledge.kb_unavailable` — a kb
+outage never reads as "the article vanished".
 
 ### `GET /knowledge/stats`
 
@@ -1833,11 +1835,18 @@ an orphan raw-doc id. Response:
 {
   "scope": "workspace:w1",
   "article_id": "deploy-runbook",
+  "new_article_id": "deploy-runbook-v2",
   "raw_doc_id": "raw-1",
   "source": "notes.txt",
-  "result": { "id": "deploy-runbook", "compiled_with": "llm", "...": "kb ingest output" }
+  "result": { "article": "deploy-runbook-v2", "title": "...", "words": 250, "compiled_with": "llm" }
 }
 ```
+
+`result` is kb-go's ingest receipt passed through verbatim — note the id key
+is `article` (finishIngest's shape), not `id`. `new_article_id` is the
+server-extracted id of the article the recompile produced; when it differs
+from `article_id` (the compile landed under a new slug) the FL-11b tracking
+on any upload row pointing at the old id is re-pointed automatically.
 
 Errors: `404 article.not_found` / `404 raw_doc.not_found`,
 `422 knowledge.empty_raw_doc`, `500 knowledge.reingest_failed`.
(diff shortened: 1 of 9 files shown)


prakashUXtech:
Review round (d25a354d):

- **Pocket isolation.** GET /knowledge/uploads now lists workspace files only — rows with a pocket_id are excluded, matching the files-service precedent. POST /knowledge/reingest-upload refuses pocket-scoped uploads with 403 `knowledge.upload_pocket_scoped`; funneling them into workspace KB would have lifted pocket-private content past the pocket ACL.
- **FL-11b tracking was dead code.** kb-go's ingest receipt keys the article id as `article` (finishIngest), so the `result["id"]` read never matched and set_kb_article never fired — same bug existed in the FileReady listener. Both now go through a shared `extract_ingest_article_id` helper, and the reingest responses carry a top-level `article_id` / `new_article_id` so clients don't touch the receipt keys.
- **Reingest re-points tracking.** When a recompile lands under a new slug, upload rows tracking the old id are moved to the new one (`MongoFileStore.reassign_kb_article`, contained) so a later hide-from-AI purge hits the live copy.
- **Outage vs miss.** GET /knowledge/articles/{id} answers 500 `knowledge.kb_unavailable` on kb timeouts/transient failures; 404 only on kb-go's genuine "Article not found". A kb outage no longer reads as a vanished article.
- **has_article false positive.** The filename fallback only applies to untracked uploads created before the matching article's compiled_at — a fresh re-upload of a same-named file stays pending instead of instantly reading compiled.

Tests: 18 in test_knowledge_wiki_api.py (pins for each of the above), 217 across the touched suites, both import-linter configs green. Docs updated in the same commit. The frontend branch has been notified of the response-shape deltas (top-level ids, the two new error codes).

PR #1874, opened 2026-08-06
Coupling sprint — one OS, not twenty apps (backend)

prakashUXtech:
Integration branch for the coupling sprint. Nine feature branches merged and verified together; the individual PRs (#1864, #1865, #1866, #1867, #1868, #1869, #1870, #1871, #1872, #1873, #1807) stay open for per-slice review, but this is the branch to land.

## What it fixes

An audit found the codebase had grown surface-by-surface, with each wave minting its own copy of shared concepts instead of relating to the existing one. The canonical owner already existed in every case, so these are foreign keys and small event bridges, not new machinery.

Privacy and trust: deleting a file now removes the knowledge extracted from it (it stayed agent-retrievable forever); the Fabric Person that agents orient on refreshes when roles and profiles change instead of staying frozen at invite-accept; decision-chain ids are real columns on Instinct actions and on the audit hash-chain, so the legal record joins the record that explains it; template-level approvals — which authorise whole classes of future writes — now reach /decisions, /activity, the audit log, and an offline owner's bell.

Cross-tenant scoping: the builtin Fabric and Instinct tools passed no workspace at all, so writes landed NULL and reads ran unfiltered — and because scoped reads render as `(workspace_id = ? OR workspace_id IS NULL)`, those NULL rows were actively served to every tenant. All seven builtin tools are scoped now and fail closed on unresolvable identity.

The site-lead funnel: a form submission on a published site rings the workspace, lands on the Leads view, and enters the outbound pipeline as a prospect with provenance back to the submission. It used to be silent.

Coupling: due-dated tasks appear on the calendar; the agent's preamble and the calendar page finally read one event store; operational alerts reach the operator's bell.

## What the integration itself caught

Two defects existed only in the merge and no individual branch could see them.

The atlas artifact went stale — four tests/atlas failures that appear only in the merged tree, found by diffing failure name sets against a matched dev baseline rather than comparing counts. Rebuilt here.

More interesting: five mount-pin tests each call mount_cloud(), which registers all six bridges, but each restored only its own topic. The governance suite passed 23/23 alone and failed four ways after test_integration.py. Investigating it turned up two things I did not expect — the leak predates this sprint (fifteen duplicate copies of pre-existing handlers accumulate too; the new bridges just made it observable, because the governance tests are the first to assert exact notification counts), and one pin was calling `lifecycle.reset_all()`, which cleared the whole lifecycle registry and silently disarmed reset for every later test. Both fixed with one shared fixture. Every pin was then re-verified individually by deleting its production registration and confirming the right pin — and only that pin — fails.

## Verification

All 99 mutations across the sprint's eight plans are caught on the merged tree, which is the claim I most wanted to check: no slice's changes weakened another's gates. OSS suite failure name sets are identical to a matched clean-dev baseline once atlas is rebuilt. Both import-linter configs pass (1 + 36 contracts). The combined cloud lane goes from 49 failures to 45, with four fixed and none new; the remainder are pre-existing groups (tasks_router, uploads folders/router, notifications router-golden) unrelated to this work.

## Known, tracked separately

mount_cloud() in tests replays the developer's real journal database when an earlier test has repointed the decisions store, which is why one test order takes 535s and the other 16s. It predates this branch and wants its own fixture. tests/ee/conftest.py has the same bus leak this fixes for tests/cloud. Rows written by the old unscoped tool path still carry NULL and remain visible to every tenant — stopping new ones does not rewrite history, and cleaning it needs a decision about rows whose owning workspace may be unknowable.


prakashUXtech:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #1883, opened 2026-08-07
Fix Code Mode file tools reporting no browser when one is attached

prakashUXtech:
Every Code Mode file tool failed with `code_delegate.no_client` — "No browser session is attached to this conversation" — while the browser was attached and streaming. Asking the agent to create a file in a local project always failed, and the agent correctly refused to pretend otherwise.

## What was happening

The file tools run inside an in-process MCP server owned by a **pooled** SDK client. That client's task is created during `_prewarm_session`, before the stream loop binds anything, and a task inherits a copy of the context as it stood at creation.

```
attach_agent_identity   bound at TWO sites:  run_core ~1054 (_prewarm_session)
                                             run_core ~1166 (stream loop)
attach_sse_event_sink   bound at ONE  site:  run_core ~1164 (stream loop only)
```

So the pooled task sees identity and never sees the sink. `has_sse_event_sink()` returns false and the channel refuses a browser that is right there.

The log signature is the proof: `code.listDir delegate failed ws=69d0f4d0…` carries a **real workspace id** while the sink reads as absent. One ContextVar from the module visible, its neighbour not, which only happens when the two were bound at different times relative to the task.

This is the same wall ART-2 hit on 2026-06-26, when the per-tenant cwd jail failed closed during warm-up. That was fixed by adding the second identity bind. The sink never got the same treatment.

## Why not just add a third bind

Prewarm has no live stream. Binding the sink there would publish a queue belonging to no turn, and a stale queue is worse than no queue — the caller would park on something nobody drains instead of failing honestly.

The channel needs to be findable by identity rather than by inheritance. It already receives the ids it needs.

## The change

The stream publishes itself under its `session_mongo_id` when it opens and withdraws in the same `finally` that detaches the sink. `delegate_call_to_browser` tries the context first, and falls back to that lookup when the context is empty.

Keyed by session rather than workspace: one workspace can hold two streams open in two windows, and a workspace-keyed lookup would hand a file write to whichever one it found first.

The fast refusal is deliberately preserved. A CLI run, a background job or a test genuinely has no browser and must keep failing at once rather than parking for the full delegate budget to discover it. That behaviour has its own test so a future change here cannot quietly turn every such caller into a timeout.

## Why the existing suite missed it

All twelve tests in `test_codeagent_delegates.py` inject `push`, and `delegate_call_to_browser` only consults the sink when `push is None`. The production path was the one path never exercised. The new test file drives the real path.

## Verification

- New tests: reproduction (fails before, passes after) plus the fast-refusal guard
- `tests/cloud/test_codeagent_delegates.py` + new file: 25 passed
- Reproduced and then confirmed **live** — local backend, desktop client, real `local` runtime project. The file tool failed before the change and writes the file after it.

The wider `tests/cloud` run is slow enough that it was still going at the time of writing; CI covers it.

### DIFF
--- a/ee/pocketpaw_ee/cloud/chat/agent_service.py
+++ b/ee/pocketpaw_ee/cloud/chat/agent_service.py
@@ -9,6 +9,16 @@ handles *what the agent sees*:
 * ``load_history_for_scope`` rehydrates prior chat turns from Mongo so the
   agent carries context across backend restarts and pool evictions.
 
+Changes: 2026-08-07 (fix/code-delegate-pooled-context) — added the session-keyed
+stream registry (``register_stream_sink`` / ``unregister_stream_sink`` /
+``stream_sink_for_session``) alongside the ``_sse_event_sink`` ContextVar. The
+ContextVar answers "is there a sink in MY context", which is right for an
+observability frame and wrong for a caller that WAITS on one: Code Mode's file
+tools run in a pooled SDK client's task created during prewarm, so they see
+identity and never the sink. The registry lets such a caller find the stream by
+session id instead. ``push_sse_event`` is unchanged and still a deliberate
+no-op outside a stream.
+
 Changes: 2026-07-14 (Paw Bar concierge seam, T2) — added ``ScopeKind.CONCIERGE``
 and ``_resolve_concierge``: a PUBLIC, anonymous Paw Bar concierge run resolves
 its ``ScopeContext`` from the server-authoritative spec (Site pocket + widget
--- a/ee/pocketpaw_ee/cloud/chat/agent_service.py
+++ b/ee/pocketpaw_ee/cloud/chat/agent_service.py
@@ -545,7 +545,8 @@ def detach_sse_event_sink(token: Token) -> None:
 # posture this module holds itself to: "the correlation id is unguessable, but
 # tenancy that rests on unguessability is not tenancy". A caller must prove it
 # belongs to the tenant whose stream it is about to push into.
-_stream_sinks_by_session: dict[str, tuple[str | None, asyncio.Queue[tuple[str, dict[str, Any]]]]] = {}
+_StreamEntry = tuple[str | None, asyncio.Queue[tuple[str, dict[str, Any]]]]
+_stream_sinks_by_session: dict[str, _StreamEntry] = {}
 
 
 def register_stream_sink(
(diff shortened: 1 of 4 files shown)


prakashUXtech:
### Code review

Found 4 issues, all fixed on this branch before landing.

1. Teardown could delete a live stream. Nothing serializes runs per session, so two concurrent `_drive_agent_loop` runs can share a scope id (a second tab, or a second send while the first stream is tailing). The later `register_stream_sink` overwrote the earlier entry, and an unconditional pop let whichever run finished first delete the entry belonging to the run still streaming — resurrecting the exact `no_client` failure this PR fixes. `unregister_stream_sink` is now identity-checked.

https://github.com/pocketpaw/pocketpaw/blob/478bdfb4c7f0d89f86a917c07dcc4ec84430e20c/ee/pocketpaw_ee/cloud/chat/agent_service.py#L570-L590

2. The registry lookup did not prove tenancy. The ContextVar path could not get this wrong — the sink was the caller's own stream by construction — but a session-keyed process-global dict can. The module already states the posture: `_Pending`'s docstring says "the correlation id is unguessable, but tenancy that rests on unguessability is not tenancy", and `PendingDelegates.resolve` enforces it on the inbound half. The outbound half now does too.

https://github.com/pocketpaw/pocketpaw/blob/478bdfb4c7f0d89f86a917c07dcc4ec84430e20c/ee/pocketpaw_ee/cloud/chat/agent_service.py#L604-L620

3. Registration sat outside the `try` whose `finally` removes it, so anything raising in that window leaked the entry — `attach_agent_identity` in particular, which rejects an unbound identity by design (added deliberately in d370751b). Unlike a ContextVar token, this entry is a process-level dict and would outlive the task, handing a dead queue to the next turn. Moved inside the `try`.

https://github.com/pocketpaw/pocketpaw/blob/478bdfb4c7f0d89f86a917c07dcc4ec84430e20c/ee/pocketpaw_ee/cloud/chat/runs/run_core.py#L1251-L1253

4. The three modified files' changelog headers were not updated, which the root CLAUDE.md requires ("When you update/make changes to a file, make sure to rewrite the comment at the top of the file with the changes you made") and which `run_core.py`, `agent_service.py` and `delegates.py` each maintain as a dated block.

Cleared during review, recorded so it is not re-raised: the "reintroduces a process-global" concern does not hold. `delegates.py` already owns a process-wide `_registry` singleton (52b1a1e2, #1762), and the past incident this resembles (#1570 → revert #1573 → re-land #1574) was about gating a fail-closed guard on a process-global flag, whose prescribed fix was "use a positive per-run marker" — which is what this is.

Still outstanding: verification stops at the delegate tests (26 passing). Per the bar set on #1858 for a cross-cutting change to this same file, the affected `tests/cloud/` subdirectories should be run on this branch and its base with the failure sets diffed, rather than relying on "CI covers it" given the known pytest addopts gap that keeps `tests/cloud` out of the documented full-suite command.

PR #1934, opened 2026-08-13
React and html sites had a contact form that posted nowhere

DevRohit06:
**Stacked on #1933** — targets `fix/sites-contact-form-schema`, because the prompt renders its field names from the contact schema that PR introduces. Merge #1933 first (with `--rebase`, per the stacked-PR rule), then this retargets to `dev`.

## The bug

The react and raw-html tracks have no server route — no `/api/submit` — and the generator's `rewireForms` fires only on the import path. `paw-sites/src/cli.ts` says so in as many words: *"the ONLY caller of rewireForms in the deployed backend path"*.

The authoring prompt asked for "a native `<form>` with flat named fields" and never said where it posts. A `<form>` with no action posts to itself. The visitor fills it in, presses Send, the page reloads, and the lead is gone. No error, no log, and a site that looks completely fine.

The react skill had already noticed and routed around it, which shows how long this has been true:

> If you have no capture endpoint to target, prefer a `mailto:` form action or a plain `tel:` / `mailto:` CTA over a form that silently drops submissions.

There is an endpoint — the shared, origin-pinned, signed-key-gated `/capture/form` the imported track already posts to.

## Both halves, one commit

**The prompt** (`native_form_contract()`) teaches the whole contract: the action, the three hidden `paw_*` inputs, and the rule that `paw_redirect` must be relative — the endpoint 400s otherwise, so the agent has to know to author the thank-you page it names. Field names render from `contact_form.CONTACT_FIELD_NAMES`, never spelled out: a prompt is one more hand-written field list, and a hand-written field list drifting from the mapping that reads it is exactly the bug #1933 fixes. Writing them again would reintroduce it one layer up.

**The substitution** (`build_generator_input`) resolves `__CAPTURE_API_BASE__` / `__SITE_ID__` / `__CAPTURE_SIGNED_KEY__` — the same tokens `scaffoldProject` already substitutes into the svelte templates. The agent has none of those values while authoring (on a create the site doesn't exist yet; the key is minted at publish), so it writes placeholders.

## Two design calls worth reviewing

**Substitution is in pocketpaw, not paw-sites.** The obvious home is the react/html materializers next to the existing `scaffoldProject` substitution. But `paw-sites` ships as a separate binary resolved via `PAW_SITES_GEN_CMD`, so a prompt live before its substitution would publish a literal `__CAPTURE_API_BASE__` into the action of every react/html site in the gap. Putting it at `build_generator_input` — the single chokepoint both publish paths funnel through — means the prompt and the resolver ship together. It also keeps paw-sites' documented "byte-identical, no transform" invariant for html intact.

**It's gated on a public deploy, and that gate is load-bearing.** The html edit lane records `byteSpan` offsets computed over the bytes the generator was handed. A real signed key is roughly twice its placeholder, so arming a build with substituted source and later splicing an edit into the stored (placeholder) source would land every span after the form out by the difference — edits landing mid-tag, silently. Armed builds keep placeholders, which also means an edit session can't mint live leads.

The pocket keeps placeholders either way: `apply_leaf_edits` has its own subprocess path and never sees a substituted map. That's what keeps `rotate_signed_key` meaningful — the next publish substitutes the new key with nothing to migrate.

## Ripple is deliberately excluded

Its form is composed by code from the same schema (`assemble_landing_spec`), so the skill supplies copy only. Handing that track the native-form contract would tell the agent to write markup it has no way to write — the "prompt may not command something the agent can't do" rule applied to markup rather than a tool. Pinned by a test, and one of the mutations injects the contract into the ripple branch to prove the pin holds.

## Refine

`_REFINE_SHARED_RULES` now says to preserve the form's plumbing. A refine rewrites whole sections, and the action plus three hidden inputs don't look like copy — they're the first thing a rewrite drops as noise. Losing them costs every subsequent lead, with no visible change to the page.

## Also updated

Both bundled skills (`pocketpaw-create-react-site`, the raw-HTML track of `pocketpaw-create-paw-site`) now carry the same contract. They're a second prompt surface reaching the same agent; leaving them saying `mailto:` would have them contradicting the handler.

## Verification

```
1472 passed   tests/ee/sites tests/cloud/surface tests/sites_capture tests/cloud/leads
              tests/test_prompt_names_only_real_tools.py
 178 passed   tests/atlas
all 10 mutations caught   tests/mutations/static_form_capture.json
```

One mutation escaped on the first run — it removed the block's warning sentence but left the endpoint, so the tests correctly didn't care. Retargeted at the `action=` attribute, which is the thing that actually matters.

## Not covered

Whether the deployed backend's `paw-sites-gen` binary matters here: it doesn't for this change (substitution is entirely upstream of the generator), but the react/html **publish smoke** has not been run against a real deploy. Worth one live publish on each track before this is trusted end-to-end.

### DIFF
--- a/ee/pocketpaw_ee/cloud/surface/handlers/sites.py
+++ b/ee/pocketpaw_ee/cloud/surface/handlers/sites.py
@@ -579,7 +579,8 @@ def _create_preamble(meta: SurfaceMeta) -> str:
             "this track — do not draft a rippleSpec or call the pocket "
             "specialist. If the skill is unavailable, author the source map "
             "yourself and call `mcp__pocketpaw_sites_manager__create_svelte_site` "
-            "(then STOP at the draft — publish only on explicit request)."
+            "(then STOP at the draft — publish only on explicit request).\n"
+            + native_form_contract()
         )
     elif engine == "react":
         engine_note = (
--- a/ee/pocketpaw_ee/cloud/surface/handlers/sites.py
+++ b/ee/pocketpaw_ee/cloud/surface/handlers/sites.py
@@ -234,6 +234,7 @@ import functools
 import logging
 from typing import Any
 
+from pocketpaw.sites_capture import contact_form
 from pocketpaw_ee.cloud.surface.domain import SurfaceMeta, SurfacePreamble
 from pocketpaw_ee.cloud.surface.handlers._helpers import content_key, meta_key
 from pocketpaw_ee.sites.react_paths import (
@@ -617,7 +618,8 @@ def _create_preamble(meta: SurfaceMeta) -> str:
             "rippleSpec or call the pocket specialist, and there is no "
             "`/api/submit` route (no server runtime), so a lead form is a native "
             "`<form>` with flat named fields.\n"
-            "CHANGES GO THROUGH THE EDIT TOOL. Once the site exists, ANY further "
+            + native_form_contract()
+            + "CHANGES GO THROUGH THE EDIT TOOL. Once the site exists, ANY further "
             "change the user asks for in this conversation — 'shorten the hero "
             "headline', 'darker pricing cards', 'add a testimonials section' — is "
             "`mcp__pocketpaw_sites_manager__edit_react_component` with the SAME "
@@ -638,8 +640,7 @@ def _create_preamble(meta: SurfaceMeta) -> str:
             'and stamps the source pocket `type="site"` + `pattern="landing"`. '
             "THEME the page with the chosen design system's tokens + your asset "
             "URLs. The lead-capture form must be FLAT native "
-            '`input`/`textarea`/`button{type:"submit"}` widgets with real field '
-            "names (name, email, phone, message) — NEVER the `form` or "
+            '`input`/`textarea`/`button{type:"submit"}` widgets — NEVER the `form` or '
             "`newsletter` widget, which nests an invalid `<form>` on a static "
             "site and captures zero leads. If the skill is unavailable, fall back "
             "to `mcp__pocketpaw_pocket_specialist__create` (build the "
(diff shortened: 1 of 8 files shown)


DevRohit06:
Extended this PR to cover the svelte track, per the captain: **svelte joins the direct-capture path and `/api/submit` is removed entirely** (509a5b4).

## Why the endpoint had to go rather than be extended

It was never a working endpoint on all four engines. It worked on ripple and dynamic svelte, never existed on react or html — and on **static svelte it was deleted by the scaffold**. `adapter-static` refuses to build a route it cannot prerender and a POST handler is exactly that, so `materializeSvelte` removed `src/routes/api` while `pocketpaw-create-svelte-site` went on teaching `action="/api/submit"`.

Those two shipped together. Every static svelte site's form 404'd, and the owner read the empty Leads view as nobody filling it in. That's the third track losing leads silently, found while wiring the first two.

## What's in this commit

- the svelte create branch gets `native_form_contract()` — it carried no form guidance at all, because the skill owned it and the skill pointed at the deleted route
- the skill is rewritten: the form, the fixed field names, and the `thank-you` route `paw_redirect` needs (the endpoint that returned a thank-you page is gone)
- **a rewrite carries the already-published fleet across.** The skill taught the old action for as long as the route existed, so it's in the stored source of every svelte pocket authored before now. `build_generator_input` repoints those forms and injects the credentials the endpoint gates on. Same reasoning as the alias table in #1933: nobody republishes a site that looks fine, and a form that 404s looks exactly like a site nobody is filling in.

The rewrite emits placeholders and lets the existing substitution resolve them, so one place knows the site id and signed key. Idempotent, leaves new-contract source alone, and matches only a form whose action **is** the legacy route — a search box on the same page isn't hijacked into minting leads.

## Paired PR

The generator half is qbtrix/paw-sites#46 — template, thank-you route, route deletion. **Either can deploy first**: the template substitution is self-contained in `scaffoldProject`, and the forms this prompt produces don't depend on the route existing or not.

## Verification

```
1656 passed   sites, surface, leads, atlas, prompt gates
 463 passed   paw-sites (vitest), tsc clean
all 14 mutations caught   tests/mutations/static_form_capture.json
```

Still not covered, and unchanged from the original description: **no live publish on any track.** Nobody has submitted a form on a deployed site and watched the lead land. That's the thing I'd want done on ripple, static svelte, react and html before trusting this end to end.

PR #1937, opened 2026-08-13
feat(sites): the free tier had nothing making it free

DevRohit06:
Removing the attribution badge is what the paid per-site tier sells, and there was no badge. Not built, not designed. Every `badge` in the tree is the gallery's capability chip, which is a different thing. So the ladder had a paid step that handed the customer what they already had.

Step 1 of the pricing build order in `docs/design/drafts/2026-08-13-paw-sites-pricing-spec.md` (workspace repo), which reviews the captain's pricing spec — finding #3 is this gap.

## How it works

Stamps onto the built pages between the build and the deploy, the same seam `_embed_concierge_bar` uses, so the artifact that lands is already right and there is no second deploy.

It **inverts that function's failure posture**, which is the point. The concierge bar swallows everything because a site going live matters more than its bar. This swallows nothing: an unreadable or unwritable page raises `BadgeInjectionError` and the publish aborts. A badge that fails open is not enforcement, because the exploit is simply to make injection fail.

One case is deliberately survivable: an empty output root. It resolves through the same `resolve_static_output_rel` the deploy reads its upload source from, so no tree means nothing ships from that root either. Raising there would protect no site and would turn every stubbed-build publish test into a failure. The reasoning is in the module and pinned by two tests, because it is the one place a future reader would reasonably "restore" a raise.

Server-rendered rather than a script, and every property that could hide it is inline and `!important`. The real attack is a rule in the customer's own stylesheet, which we build from, not our artifact, which they never touch.

## The gate

New `badge_removal` flag on `SitePlanTier`. `basic` keeps its badge; `pro` and `business` drop it. A first publish reaches the stamper before its `Site` doc exists, so an absent tier resolves to badged — the opposite default would ship every new site clean.

When `basic`/`pro`/`business` are rekeyed to the new plan ids, that mapping is the only place the badge's plan gate lives.

## Verification

29 tests. 1260 passing across `tests/ee/sites/` and `tests/cloud/sites/`, no regressions. `ruff check` and `ruff format` clean.

Eight mutations in `tests/mutations/free_badge.json`, all caught. The first sweep had one escape worth recording: `rfind`→`find` walked through `test_the_last_closing_body_wins`, because asserting the badge sits before the *last* `</body>` is true even when it lands inside a `<code>` sample. The assertion now pins it after the first.

## Not in this PR

Republish-on-plan-transition. A downgrade that flips `badge_required` has to rebuild a deployed static artifact, and that needs the plan catalog that step 3 introduces. Filed as a follow-up rather than half-built here.

### DIFF
--- a/ee/pocketpaw_ee/cloud/billing/site_plans.py
+++ b/ee/pocketpaw_ee/cloud/billing/site_plans.py
@@ -13,6 +13,9 @@
 #     records the intended tier WITHOUT a live charge.
 #   * ``cloudflare_features`` — the set of Cloudflare features a higher tier
 #     resells (BC-10 provisions these on publish). The base tier resells none.
+#   * ``badge_removal`` — whether a site on this tier may ship WITHOUT the
+#     free-tier attribution badge. This is the per-site paid tier's headline
+#     feature, so it lives on the tier rather than on the workspace plan.
 #
 # Read-only by design (mirrors ``billing.plans``): no DB, no writes, no emit.
 # ``list_site_plans`` / ``get_site_plan`` return frozen ``SitePlanTier`` value
@@ -20,6 +23,16 @@
 # never drift.
 #
 # Created 2026-06-24 (integration/billing-credits, BC-9): new module.
+# Updated 2026-08-13 (feat/sites-free-badge): added ``badge_removal`` — the gate
+#   ``sites.badge`` reads to decide whether a publish must stamp the attribution
+#   badge. The base tier does NOT carry it (that is what free means); the paid
+#   tiers do. Sourced from its own constant rather than folded into
+#   ``cloudflare_features``, which is specifically about RESOLD Cloudflare
+#   capability and would be the wrong home for a billing-policy flag.
+#   NOTE for the pricing-spec migration (step 3 of the build order in
+#   docs/design/drafts/2026-08-13-paw-sites-pricing-spec.md): when basic/pro/
+#   business are rekeyed to free/site/staff, this mapping moves with them and is
+#   the ONLY place the badge's plan gate is expressed.
 
 from __future__ import annotations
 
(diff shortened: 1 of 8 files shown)


DevRohit06:
## Step 2 pushed — and it closes a hole the first commit left open (`0bc5abd6`)

Wiring the badge through a real per-site entitlement resolver surfaced this: **cancellation sets `subscription_status` and leaves `plan_tier` on the paid key.** Nothing resets it. So the gate in the first commit, which read the tier alone, gave a cancelled site a permanent badge removal.

The same hole is open wider on `dev` right now. No Dodo product is configured, so `_dodo_product_for` returns `None` for every tier and a paid publish "degrades gracefully" to recording the intended tier **without taking money** — landing on `plan_tier="pro"`, `subscription_status="none"`. Under a tier-only gate that site ships unbadged, free, forever.

### What changed

`SiteEntitlements` + `resolve_site_entitlements` in `entitlements/`, the per-site scope beside the workspace one. Sites are the only thing billed per-object, so "may this site drop its badge" was never answerable from the workspace plan.

Every paid capability now requires the tier to grant it **and** the subscription to be active. `none`, `pending` and `cancelled` all resolve to no paid capability. Failing closed on `pending` is safe because a pending site is not deployed yet — the charge-first flow deploys on activation.

The resolver is pure and takes the site's billing fields rather than reading them: `entitlements` may not import `models.site` (EE cloud rule 2), and passing them in means every branch tests without a database.

Carries `badge_required`, `custom_domain`, `concierge_enabled`. `conv_allowance`, `conv_rate_usd` and `white_label` are deliberately absent — the first two wait on which meter owns a concierge run, the third on an org entity that doesn't exist. A field that always returns 0/False reads as implemented.

### Verification

80 tests across the two files. 1474 passing across sites/entitlements/billing/credits. 11 mutations, all caught.

Two of those mutations were rewritten mid-sweep for escaping, both worth recording:
- the first-publish anchor **rotted** when its line moved into the resolver
- its replacement then **escaped**, because defaulting `plan_tier` alone changes nothing once the subscription gate exists. It only bites when both defaults flip together, which is what it now mutates. Defense-in-depth working, but a mutation that changes no behavior tests nothing.

### Pre-existing, not from this branch

`tests/cloud/billing/test_plans.py` has 2 failures here (`test_dodo_product_id_is_none_until_bc7`, `test_get_billing_plans_returns_the_catalog`). They reproduce with these changes stashed — real Dodo product config is reaching that test's env, so it asserts `is None` against a live `pdt_...` id. Untouched by this PR, but worth someone's attention since it means the workspace-plan catalog is picking up production config in tests.

PR #1958, opened 2026-08-17
Teach the atlas self-model source-truth and the verify loop

prakashUXtech:
## What

Two subsystems shipped after atlas was seeded and never reached it: Fabric source-truth (FST-1..8) and the self-verifying loop (SVL-1..5 + J-1). Ask atlas "where did this number come from" or "how does the OS verify a task got done" and it returned nothing — the capability existed but was undiscoverable to agents. This is stack base 1 of 4 (AST-1).

## Change

**Two new primitives** in `authored/primitives.json`, both with primer gists:
- `primitive:source-truth` — statements, the trust ladder (`human-pin > human > connector > mirror > agent > inferred`), one resolver → winner + `is_disputed`, read-time freshness, un-rankable conflicts → Instinct stewardship (PIN / IGNORE). Disambiguator: not a knowledge base, not a fact-checker — it ranks sources, it doesn't judge truth.
- `primitive:verify-loop` — verify-on-completion `OutcomeVerdict` (`SOLVED / PARTIAL / NOT_SOLVED / UNKNOWN`), bounded requeue, convergence guard, LLM judge in shadow. Disambiguator: not a test runner, not the Instinct gate — it verifies outcomes; humans still approve actions.

**Two capability cards** in `authored/capabilities.json` (`role:member`): `fabric.provenance_read` and `fabric.conflict_steward`.

They're modelled as two primitives, not one umbrella — the code has two independent subsystems and "company brain" has no code referent.

## One naming call worth flagging

`capability:fabric.conflict_steward` is named "Arbitrate a source conflict", not "Arbitrate a disputed Fabric fact". A name carrying the pinned intent tokens out-scores the governing primitive at name-weight, so the primitive would lose "is this fact disputed" to its own capability card. Fixed by authored-name discipline; the scorer is untouched. Arguably the more accurate name anyway — it's a source conflict, matching the disambiguator.

## Verified

- `pytest tests/atlas` → 187 passed (baseline 178).
- Eval strict hits 35/36 → 43/44, all 8 new cases rank #1, pinned in both directions (the new primitives win their intents; `primitive:fabric` and `primitive:instinct` still win theirs).
- Primer 1987/2000 chars with both gists; the primer test fails loudly on overflow.
- `atlas build --check` up to date; both import-linter configs green; ruff clean.

Stack: this → #AST-2 (trust aggregate) → #AST-3 (flag-aware) → #AST-4 (docs + rule). Merge with `--rebase` (stacked dependents).

### DIFF
--- a/src/pocketpaw/atlas/authored/capabilities.json
+++ b/src/pocketpaw/atlas/authored/capabilities.json
@@ -1,4 +1,5 @@
 {
+  "_comment": "capabilities.json — hand-authored capability cards, each role-gated by a role:<tier> marker in requires (WA-3). Updated 2026-08-17 (feat/ast-1-atlas-primitives, AST-1): added the two member-level fabric cards capability:fabric.provenance_read and capability:fabric.conflict_steward. Their names/summaries deliberately avoid the tokens 'fact'+'disputed' together (and only provenance_read says 'Fabric') so the two-content-token eval intents ('is this fact disputed', 'what is fabric') keep the primitives at #1 — a name hit outscores the primitive's keyword ceiling.",
   "schema": "paw.atlas/v1",
   "entries": [
     {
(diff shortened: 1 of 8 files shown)


prakashUXtech:

PR #1959, opened 2026-08-17
atlas_describe fabric:<Type> carries a live source-truth aggregate

prakashUXtech:
## What

The atlas Fabric introspector describes entity *types* from the ontology registry — it knows `Customer` has a property `arr` but nothing about whether `arr` is disputed, stale, or inferred. Source-truth attaches to object *instances*, in a different database. This adds one bounded, type-level read that bridges them. Stacked on #1958 (AST-2).

## Change

`atlas_describe fabric:<Type>` now carries an additive `source_truth` roll-up: per property, how many tracked objects, how many disputed / stale / aging, and the winner writer-class mix, plus `mode` / `tracked` / `sampled` / `object_count` and a pointer to `fabric_query include_provenance=true` for the per-object answer. It points at the shipped instance-level read instead of duplicating it.

Mechanics:
- A new optional, **async**, duck-typed `RegistryFabricIntrospector.entity_type_source_truth` — documented in the same "optional, not a Protocol member" block as `list_entity_properties` (adding it to the Protocol would break `isinstance` for two-method fakes).
- `build_workspace_fabric_introspector` binds the OSS `get_fabric_store(workspace_id=)` in an inner try, so a store failure degrades alone.
- Only the new `describe_fabric_id_async` consumes it (`getattr` + `callable` + `await`, from the already-async handler). The sync `describe_fabric_id` is unchanged. Search never calls it — search stays properties-only.
- `FabricStore.list_statement_keys` gained additive `type_id` / `limit` kwargs: an indexed `fabric_objects(type_id) ⋈ fabric_statements(object_id)` join. Existing callers are byte-identical.

The winner cache stores only the value, so `is_disputed` / freshness / winner writer-class are computed by running the pure resolver per tracked key — cost bounded by the tracked-key set, never a full-fabric scan. Cap 500 keys → `sampled=true`. Untracked type = one indexed lookup, zero statement reads.

## Verified

- `pytest tests/atlas` → 202 passed (baseline 187 + 15 new); the #1645 N+1 regression test still green.
- Zero-read proof (spy over a real store): untracked → `{list_statement_keys: 1, get_statements: 0}`.
- Tracked demo: two competing statements on `arr` → `properties.arr.disputed == 1`, `winner_writer_mix == {connector: 1}`.
- Sample cap: 510 tracked keys → walked 500 in 0.28s, `sampled` true.
- `EXPLAIN QUERY PLAN` pinned: `SEARCH ... USING INDEX idx_objects_type` / `USING COVERING INDEX idx_statements_object_property`, no SCAN.
- Store-binding failure → registry payload intact, `source_truth` absent.
- `atlas build --check` unchanged; both import-linter configs green; ruff clean.

### DIFF
--- a/src/pocketpaw/agents/sdk_mcp_atlas.py
+++ b/src/pocketpaw/agents/sdk_mcp_atlas.py
@@ -49,6 +49,15 @@
 # role-gated id like an unknown id, and the known-ids error listing excludes
 # them. Belt-and-braces: admin capabilities can never leak from ANY atlas server
 # wiring, provider or not (matches DefaultEntitlementProvider's role hiding).
+#
+# Updated: 2026-08-17 (feat/ast-2-atlas-trust-aggregate, AST-2) — the describe
+# handler awaits ``describe_fabric_id_async`` instead of the sync
+# ``describe_fabric_id``: same registry payload for ``fabric:<type>`` ids, plus
+# the additive ``source_truth`` field (live per-property disputed / stale /
+# aging counts + winner writer mix from the OSS FabricStore) when the
+# introspector exposes the optional ``entity_type_source_truth``. Handlers are
+# already async, so the aiosqlite read runs in the handler's own loop — no sync
+# bridge. Search is untouched (properties-only, FINDING B).
 
 from __future__ import annotations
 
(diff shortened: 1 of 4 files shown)


prakashUXtech:

PR #1960, opened 2026-08-17
Mark the flagged atlas primitives with their live rollout mode

prakashUXtech:
## What

Both new primitives ship behind rollout flags that default off (`fabric_source_truth_mode`; `deep_work_verify_mode` / `cloud_plan_verify_mode`). atlas has to stay discoverable *and* honest: always describe them (the agent must know the OS has them), but say whether they're live for this OS config. Hiding them when off would recreate a known misdirection bug; ignoring the flag would claim capability that isn't there. Stacked on #1959 (AST-3).

## Change

The overlay now stamps flag-driven `available` plus an additive tri-state `OverlaidEntry.mode` (`off | shadow | enforce`) on exactly `primitive:source-truth` (from `fabric_source_truth_mode`) and `primitive:verify-loop` (the higher of the two verify modes). Flags are read from live settings once per overlay pass in the provider-independent `_overlay_one`, so the EE role-aware provider has parity by construction (pinned by test, not duplicated).

- Sort logic is untouched — an off primitive demotes through the existing `available is False` re-sort but is never hidden.
- Describe and search cards carry `mode`; an off primitive's describe adds an `enable_hint` (the env flag to set, plus `docs/atlas.md`) instead of the connector `connect_hint`.
- `is_granted` is unaffected by the flags. `available` / `mode` are discovery hints only, never an enforcement gate — stated in a comment and asserted by test.
- `TestFactualClaimGuard` gained a liveness guard: a describe cannot assert live source-truth or verification when the mode is off.

## Verified

- `pytest tests/atlas tests/ee/agent/test_atlas_role_provider.py` → 246 passed (25 new overlay-flag tests, 4 new EE parity tests).
- Live env demo, real handlers: default → `available False, mode off, enable_hint set`; `POCKETPAW_FABRIC_SOURCE_TRUTH_MODE=shadow` → `available True, mode shadow, no hint`; `POCKETPAW_DEEP_WORK_VERIFY_MODE=enforce` → verify-loop `available True, mode enforce`.
- `git diff overlay.py | grep 'demoted|sort'` → no changed lines.
- `atlas build --check` unchanged; both import-linter configs green; ruff clean.

### DIFF
--- a/docs/atlas.md
+++ b/docs/atlas.md
@@ -67,6 +67,9 @@ search dropped a 3N sqlite N+1. SECURITY (separate PRs): the role-aware
 provider now re-resolves per turn (a warm shared client had leaked owner
 cards to a member), and the admin path gained an OWNER guard on owner-grants
 + an atomic approve.
+Updated: 2026-08-17 (feat/ast-3-atlas-flag-aware, AST-3) — flag-aware
+`available` + tri-state `mode` for primitive:source-truth / primitive:verify-loop
+in the overlay section; `mode?` on search cards.
 -->
 
 # Atlas — the OS self-model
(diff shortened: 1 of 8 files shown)


prakashUXtech:

PR #1973, opened 2026-08-19
The owner's inbox showed one row for a visitor who asked four questions

DevRohit06:
Conversation identity shipped on the visitor's side a while back and stopped there. A visitor can now hold several conversations, each with its own id, its own run `session_key` and its own history replay. None of the owner-facing code moved with it — every seam still keys on `customer_ref`, the person — so the two halves of the product disagree about what a conversation is.

## What that costs

| Seam | Behaviour today |
|---|---|
| `GET …/conversations` | Dedupes the run scan by `run.user_id`. Four separate questions render as **one** row showing the latest sentence. |
| `POST …/conversations/{ref}/reply` | Resolves to the visitor's **active** row. Answering yesterday's question writes into whatever they're saying today. |
| `PATCH …/conversations/{ref}` | Same. Taking one thread over mutes the assistant in the thread they're typing in. |
| `GET /paw-bar/messages/{widget}/{ref}` | Serves every owner line the visitor has ever been sent, into whichever thread is on screen. |

The missing three conversations aren't collapsed behind a disclosure — they're absent, with nothing to click. And `paw_bar_owner_messages` had no conversation column, so a reply couldn't be addressed to a thread even in principle.

## The fix

Carry the id the rest of the system already had. Owner lines store the conversation they were said in; the inbox groups runs by the conversation encoded in `session_key` and joins state by id; the reply, the patch and the visitor's poll all take a `conversation_id`.

Deployed owner lines are **backfilled** on migration to the visitor's active conversation — there was exactly one when they were written, which is the whole premise — rather than left empty behind a reader fallback. A nullable "means the old thing" is how the pair-keyed bug survived this long.

Compatible in both directions: a client that sends no `conversation_id` gets the old resolve-to-active behaviour, so cached dashboard and widget bundles keep working, and a legacy run whose `session_key` carries the visitor handle still groups per visitor and still lists.

One thing worth calling out — **a client-supplied id is an address, not a credential.** It's checked against the `(widget, visitor)` in the path before anything is written, and a mismatch is a 404. The silent fallback is right on the *visitor* path, where a refusal would confirm an id was real; it's wrong here, where the caller is already authenticated and workspace-scoped.

## Verification

- 7 new tests in `tests/cloud/test_paw_bar_owner_conversation_identity.py` — **run red first**, the headline one failing `assert 1 == 2` on the inbox
- 8 mutations in `tests/mutations/paw_bar_owner_conversation_identity.json`, **each observed to break its test**
- One escaped on the first run, and it was a hole in my test rather than in the code: it asserted the reply was *absent* from the wrong thread but never that it *reached* the right one, so storing it with no conversation at all passed. Fixed, re-run, all 8 caught.
- `--validate` repo-wide: 581 anchors across 48 plans resolve. My edit rotted one anchor in the sibling `paw_bar_conversation_identity` plan; it's repointed and that plan's 11 mutations were re-run and all still catch.
- Full `-k paw_bar` suite: 572 passed. The 12 failures are pre-existing — verified by stashing this diff and getting the identical list on a clean tree.
- `ruff check .` clean.

Docs updated in `docs/api-reference.md` for all four routes.

## Not in this PR

The `/chat` sites section that reads this — the owner surface for browsing sessions and stepping in — is next, and needs this merged first since it's the thing that makes more than one row exist.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -2333,7 +2333,7 @@ the split is the security model:
 | `POST /paw-bar/action` | Run a verb the widget spec declares. `auto` verbs touch only the visitor's own cart or a checkout link; `gated` verbs execute nothing and raise an Instinct proposal for a human. |
 | `GET /paw-bar/cart` | The visitor's own cart. |
 | `POST /paw-bar/decision-contact` | Leave an email so a decision reaches the visitor after they close the page. The address is stored on the decision row only — never in agent context, the KB, or transcripts. |
-| `GET /paw-bar/messages/{widget_id}/{customer_ref}` | Poll for owner and system messages once a human has joined. Returns `role`, `content`, `at` and `bot_paused` — never notes, tags, assignee or contact address. |
+| `GET /paw-bar/messages/{widget_id}/{customer_ref}` | Poll for owner and system messages once a human has joined. Returns `role`, `content`, `at` and `bot_paused` — never notes, tags, assignee or contact address. Pass `conversation_id` to scope the read (and `bot_paused`) to the thread on screen; omitting it answers for the visitor's whole history, which is what a cached widget bundle does. |
 | `GET /paw-bar/articles` | The site's own synced pages, for a self-serve reading list. |
 
 ### Admin — the owner surface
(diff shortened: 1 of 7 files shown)


DevRohit06:
Pushed one more fix after re-reading my own grouping code.

**A conversation that spans the migration was rendering twice.** A visitor mid-thread when this deploys has runs of *both* spellings on the same conversation — older ones carrying their handle in the `session_key`'s conversation slot, newer ones carrying the real id. Grouped by those raw tokens they're two rows, and both resolve to the same conversation, so the pair share a `conversation_id`. The owner sees the thread twice, and any client keying its list on the id has duplicate keys.

That's the deploy-day shape, not an edge case. Grouping is now by the conversation a run **resolves to**, falling back to the visitor only when there's no conversation to name.

Reproduced first (`assert 2 == 1`), then fixed. Now 9 tests and 11 mutations, all caught — including one that escaped because my mutation set a variable a later line overwrote, so it wasn't testing anything.

584 anchors across 48 plans still resolve; 574 passing in the `-k paw_bar` suite with the same 12 pre-existing failures.

PR #1993, opened 2026-08-21
The domain paywall took the concierge off every site

DevRohit06:
**Live regression from #1988, reported hours after it merged.** Setting `POCKETPAW_SITES_BILLING_ENFORCED=1` to cap custom domains removed the visitor concierge from every site.

**Mitigation while this is in review:** `POCKETPAW_SITES_BILLING_ENFORCED=0` and restart. That brings the concierge back and takes the domain cap down with it. This PR is what lets you have both.

## What happened

#1988 routed five seams through `sites_enforced()`. Two of them are the domain caps. Two of them are the concierge — the visitor-facing `concierge_available` and the publish-time embed decision. One switch, two unrelated products, and only one of them was what anyone meant to turn on.

## Why it could not have gone any other way

This is the part that makes it more than a mis-scoped flag. `concierge_entitled` needs `tier.sells_concierge` **and** an active subscription:

- `sells_concierge` is `key != BASE_SITE_PLAN_KEY`. Every site in production is on the floor, so no site qualifies.
- Even one that did would need `subscription_status == "active"`, which is only ever written by `activate_site` off a Dodo webhook, which needs a product id from `POCKETPAW_DODO_SITE_PRODUCTS` — configured in no deployment file anywhere.

So the gate refuses every site, and no customer can buy their way out of it. A paywall that can only refuse and never convert is an outage wearing a paywall's clothes. There was never a version of that flag flip that ended differently.

## The fix

The concierge gets its own switch: `sites_concierge_enforced`, default off, read by **neither** `billing_enforced` nor `sites_billing_enforced`. `concierge_enforced()` sits beside `sites_enforced()` in `cloud/billing/enforcement.py` and the docstring says plainly why it is not an OR.

The gate itself is untouched and still correct. It wants turning on the day a tier can actually sell a concierge, and not one deploy earlier.

**`billing_enforced` is unhooked from it too**, and that half is deliberate rather than incidental. It had the identical trap armed for whoever eventually turns credits enforcement on. Fixing one and leaving the other would have meant knowingly shipping this same report a second time.

## Tests first

Four assertions written against the merged code, all four failing:

```
FAILED test_the_domain_flag_does_not_switch_off_the_concierge
FAILED test_the_global_billing_flag_does_not_switch_off_the_concierge
FAILED test_both_paywall_flags_together_still_leave_the_concierge_alone
FAILED test_its_own_flag_still_enforces_the_plan
```

Ten now pass. The set covers the reported case, the same trap on the other flag, the gate still working when asked for, a paying site still being served, the owner's own `concierge_enabled` switch winning under every flag combination, and the real `Settings()` default rather than a stub — that last one because every other test monkeypatches `get_settings` and so cannot see the declared default, which is the whole safety property.

Three existing stubs updated to set the new flag: `test_paw_bar_concierge_entitlement`, `test_concierge_autoembed`, and #1988's own `test_sites_billing_flag`. That third one had asserted the coupling — `test_the_sites_flag_alone_makes_the_concierge_gate_bite` — and now asserts the opposite, filed with the workspace-cap negatives where it always belonged.

## Mutations

`concierge_entitlement.json` goes from 17 to 19, all caught. The two new ones are the exact mistakes in play: **re-attaching the concierge to `sites_enforced()`**, and **flipping the new flag's default to `True`**. Three anchors also needed repointing, since the seams moved.

Worth noting the second one escaped on the first run and the escape was informative: the mutation was pointed at a test that monkeypatches `get_settings`, so it could not observe a field default at all. It needed the `Settings()` test to become killable.

## Docs

New row in `configuration-reference.mdx` telling operators to leave it off until a tier can sell a concierge, plus a clause on the `sites_billing_enforced` row saying it no longer touches the concierge and why.


DevRohit06:
Closing — the premise was wrong.

The concierge disappearing from a free site is the free plan working, not a regression. I read "a setting I added changed something the captain didn't ask about" as a bug and escalated it to an outage; it was neither. Free doesn't include a concierge, enforcement went on, the concierge went away.

Separating the switches would still be defensible on operability grounds, but it buys flexibility nobody needs today and carries a real footgun: merged without also setting POCKETPAW_SITES_CONCIERGE_ENFORCED=1, it hands the concierge back to every free site. Not worth it to keep behaviour that is already correct.

The investigation did turn up something real, filed separately: per-site plans cannot be purchased at all. dodo_site_products is read via getattr on Settings and is not a field on Settings, so POCKETPAW_DODO_SITE_PRODUCTS does nothing and every tier's dodo_product_id is None no matter what. That is the upgrade path behind the domain cap, and it is a dead end.

PR #2027, opened 2026-09-01
An employee can ask an admin for a paid site plan

DevRohit06:
Stacked on #2024 (`feat/sites-plan-purchase-authz`), which owns `purchase_authorized` and the `sites.buy_plan` rule. Merge that one first.

## Why

#2024 stopped a member charging the company card for a site plan. That was the right refusal, and it was also a dead end: the employee who built the site got a 403 and no route forward except finding an admin out-of-band and describing what they wanted.

This gives the refusal an answer. `POST /sites/plan-requests` files an Instinct Action; an admin approves it in The Tray and the executor performs the publish that was refused. **Filing the request publishes nothing and charges nothing.**

## The inversion

The design mirrors the `_admin_action` gate closely, with one deliberate difference that is the whole point of the module.

Every other gated proposal kind re-checks the **proposer's** role at execute time — right there, because the proposer is an admin whose rights might have been revoked since. Here the proposer is a member who *never had the right*; that is the premise. So the check runs against the **approver**, read from `Action.approved_by`, which the instinct router writes from the authenticated identity (`str(user.id)`) and never from a request body field.

Get it backwards and the feature refuses every request it exists to serve. Leave it out and any pending request buys the moment anyone touches it. Both failure modes have a test.

This is safe because `instinct.approve` and `sites.buy_plan` are both ADMIN — an approver has necessarily cleared the purchase bar already. The check runs anyway rather than being inferred: those two rules live in different files and can drift, and "the approve endpoint already checked something similar" is not a control. A test fails if the two ever separate, so raising `sites.buy_plan` to OWNER later is a caught change rather than a silent one.

## Price drift: reported, not enforced

The blob records the price the requester saw. The purchase prices from the **live catalog**, because the catalog is the truth about money and a week-old snapshot is not — and the outcome names the difference, so an admin who approved "$7/month" can see they were charged $9.

Refusing on any drift would turn an ordinary price change into a pile of dead Tray cards. The identity hash already pins what must not move: workspace, site, tier. Price is deliberately excluded from it.

## Guards

Ordered so nothing that charges runs before all of them: blob present → schema → tenancy → approver identity → identity hash → idempotency (under a per-action lock, re-read fresh) → RBAC re-check → the write. Tenancy joins `_assert_gated_workspaces`, so all four dispatch paths (approve / bulk-approve / reject / bulk-reject) inherit it, and `_site_plan_request` joins `RESERVED_GATED_PARAM_KEYS` so an approve-time edit can neither mint the blob on an innocuous card nor move its tenancy.

## The mutation run

All 16 caught (`tests/mutations/site_plan_request_gate.json`). One escaped first, and it was the most security-relevant one: my test asserted the approver id came from `approved_by`, but the fixture only carried that field — so it passed just as happily against a version preferring a client-controlled `approver`. The fix was an Action carrying **both** with different values, which is the only input that separates the two readings.

Two other things the run caught: a `site_scoped_tier` → `get_site_plan` swap (org-scoped flats become requestable, producing cards that can only fail on approval), and pricing from the stale blob instead of the catalog.

## Tests

363 pass across every suite this touches — `tests/cloud/sites/`, `tests/instinct/`, plus the instinct approval-security, bulk-action, dispatch and fail-closed-tenancy suites. 21 new. Ruff clean, import-linter contract kept.

The UI half (the refusal offering "ask an admin") is a separate paw-enterprise PR.

### DIFF
--- a/ee/pocketpaw_ee/cloud/site_plan_requests/__init__.py
+++ b/ee/pocketpaw_ee/cloud/site_plan_requests/__init__.py
@@ -1,0 +1,47 @@
+# ee/cloud/site_plan_requests — an employee asks for a paid site plan; an admin
+# approves it, and the approval is what buys.
+#
+# Created: 2026-09-01 (feat/sites-plan-purchase-request).
+#
+# The gap this fills. ``sites.buy_plan`` (ADMIN) stopped a member charging the
+# company card for a site plan, which was the right refusal and a dead end: the
+# employee who builds the site gets a 403 and no way forward except finding an
+# admin and describing what they wanted. This turns that refusal into a request.
+#
+# The blob is the 14th gated proposal kind, alongside ``_admin_action`` and
+# ``_artifact_change``. Its shape mirrors ``_admin_action`` closely because the
+# problem is the same one — gate a privileged write behind a human — with ONE
+# inversion that is the entire point of this module:
+#
+#   ``_admin_action`` re-checks the PROPOSER's role at execute time. Here the
+#   proposer is, by construction, NOT authorized: a member asking for a plan is
+#   exactly who this exists for. So the execute-time RBAC check runs against the
+#   APPROVER (``Action.approved_by``, the authenticated identity the router
+#   records — never a free-text field). Re-checking the proposer would refuse
+#   every request this feature creates; omitting the check entirely would let any
+#   pending request buy the moment anyone approved it.
+#
+# That inversion is safe because ``instinct.approve`` is itself ADMIN — the same
+# minimum as ``sites.buy_plan`` — so an approver has already cleared the bar the
+# purchase needs. The executor re-checks anyway rather than inferring it: the two
+# rules live in different files and can drift, and "the approve endpoint already
+# checked something similar" is not a control.
+#
+# See ``propose.py`` for the blob shape and ``executor.py`` for what approving
+# actually does.
+
+from pocketpaw_ee.cloud.site_plan_requests.propose import (
+    SITE_PLAN_REQUEST_KIND,
+    SITE_PLAN_REQUEST_PARAM_KEY,
+    SITE_PLAN_REQUEST_SCHEMA,
+    compute_request_hash,
+    propose_site_plan_request,
+)
+
+__all__ = [
+    "SITE_PLAN_REQUEST_KIND",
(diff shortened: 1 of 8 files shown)


DevRohit06:
Worth flagging before review: the two green checks on this PR are not the test suite.

`ci.yml` triggers on PRs targeting `dev`, `main`, or `feat/*-integration`. This one targets `feat/sites-plan-purchase-authz` (the stacked base), so pytest, Lint, the mutation-anchor check and the OSS-EE boundary contract never ran here — what passed is the quality gate and a secret scan, and the scan itself logged "No Python files changed".

So the test evidence for this branch is local only:

- 363 pass across `tests/cloud/sites/`, `tests/instinct/`, and the instinct approval-security / bulk-action / dispatch / fail-closed-tenancy suites
- all 16 mutations in `tests/mutations/site_plan_request_gate.json` caught
- `ruff check` clean, import-linter contract kept

#2024 underneath it does target `dev` and got the real matrix. Once it merges, rebasing this onto `dev` will pick up full CI — worth doing before merging rather than after.

DevRohit06:
Followed up on the failures I hit running `tests/cloud/` locally — they are **pre-existing, not caused by this branch**, and the comparison is exact.

I ran every `tests/cloud/` file that imports a module this PR touches (22 files, found by grepping for `instinct.router` / `sites.router` / `site_plan_requests`) against a clean worktree of the base branch:

```
this branch : 14 failed, 458 passed, 1 xfailed
base branch : 14 failed  (same 22 files minus my new test file)
diff of the sorted FAILED lists: empty
```

Same 14 test names on both sides, byte for byte. The bulk are in `test_ee_correction.py` (approve-endpoint + corrections) and `test_decision_traces_wiring.py`.

**Why nobody had noticed:** `pyproject.toml` sets `addopts = "--ignore=tests/cloud --ignore=tests/e2e"`, and both CI test jobs invoke `uv run pytest tests/` — so that ignore applies to them too. `tests/cloud` has never run in CI on any branch. The EE job's 10,634 passing tests exclude the whole directory. That is a much larger gap than this PR and predates it; I have not touched it here, but it is worth its own issue.

Test evidence for this branch stands as: 363 passing across `tests/cloud/sites/`, `tests/instinct/` and the instinct approval-security / bulk-action / dispatch / fail-closed suites, all 16 mutations caught, and the no-regression comparison above.

PR #2073, opened 2026-09-04
fix(sites): import from URL failed on most real sites

DevRohit06:
Importing a site from a URL fails on most real sites. It fails late, in the deploy, and the report the user ends up looking at says nothing useful.

## What happens

The link scanner queues every `<link href>` as an asset. Plenty of those address pages rather than assets, `rel="canonical"` above all, which every static-site generator emits on every page.

The page loop already handles the reverse case: a URL queued as a page that answers with a non-HTML content type gets re-routed and stored as an asset. The asset loop had no mirror of that check. So a page fetched through the asset queue was stored at its extension-less path, and `/about` landed at the file `about`.

The same page reached through `<a href="/about/">` lands at `about/index.html`. A file map holding the file `about` and the directory `about/` cannot exist on a filesystem. The generator's `materializeHtml` mkdirs over the file and dies:

```
EEXIST: file already exists, mkdir '.../about'
```

`crawl_site_from_url` catches that and writes `status:"failed"` on the report, so the whole import ends with nothing.

## Evidence

Crawling pocketpaw.xyz with the real crawler produced 112 files, 48 of them a path that is both a file and a directory. Handing that map to `materializeHtml` reproduces the EEXIST. Handing it the map this branch produces writes all 64 files with no error.

| | before | after |
|---|---|---|
| Files in the map | 112 | 64 |
| File and directory collisions | 48 | 0 |
| Assets fetched | 63 | 15 |
| Bytes fetched | 23.7 MB | 16.2 MB |

The two file sets are byte-identical where they overlap. The 48 that disappear are exactly the extension-less duplicates, and every one of them has its proper twin in the after map, so nothing was lost to buy the 32% drop in bytes.

## The fix

The asset loop gains the page loop's content-type guard in reverse. An asset that answers `text/html` claims the page path, so both spellings of a URL converge and first-fetch-wins deduplicates them. The body gets the page loop's exact treatment, same-origin rewrite included, because storing raw bytes would leave that one page linking back to the site it was imported from.

The scanner also stops queueing navigational rel values, so those fetches never happen at all. That is a denylist rather than an allowlist: an unfamiliar rel is still fetched, and the content-type guard catches it if it turns out to be a page, so a site can hang a page off any rel it likes and still import. A `<link>` with no rel, and `rel="alternate stylesheet"`, both stay assets.

The asset ceiling now counts fetch slots rather than stored assets. Gating on `assets_fetched` alone would have let a re-routed fetch escape the cap and drain the whole 2000-entry queue, several hundred fetches where 200 was the ceiling.

No SSRF, scope, robots, redirect, or byte-budget behavior changed.

## Testing

62 tests pass across the crawler and import suites, four of them new. Each new test was checked against a mutated copy of the module to confirm it actually gates: dropping the rel filter, routing the guard back to the asset path, and gating the loop on the old counter are all caught.

Worth knowing: the first test written for this, the canonical one, no longer reaches the content-type guard, because the denylist now filters canonical before it is queued. It exercises the filter instead. A second test uses `rel="me"` to reach the guard, and that one does fail under the mutation.

One report number moves. `cross_origin_refs` drops from 85 to 83 on the same site, because two external connection hints are no longer counted as content references.

The wider sites suite has three failures on this branch, in the concierge auto-embed and custom-domain billing tests. Neither file imports the crawler or the import service. They are pre-existing on dev.

## Still broken, not fixed here

When an import does fail, the frontend gives no sign. The backend writes both the failed status and a readable error, and both reach the browser, but the report panel only branches on the queued state. A failed import renders as an ordinary report with zero pages and zero assets. Nothing polls for the background crawl either. That is a paw-enterprise change and wants its own PR.

### DIFF
--- a/ee/pocketpaw_ee/sites/url_crawler.py
+++ b/ee/pocketpaw_ee/sites/url_crawler.py
@@ -34,6 +34,19 @@
 # Cross-origin refs are left as-is and counted for the import report.
 # Edited 2026-07-23 (SSRF review): redirect scope guard (off-host 30x rejected
 # for non-seed fetches) + seed apex->www re-seed; opposite-scheme origin rewrite.
+# Edited 2026-09-04 (page/dir collision): the asset loop gained the page loop's
+# content-type guard, the other way round. A <link href> that answers text/html
+# (rel="canonical" above all) now claims the PAGE path, so /about stops landing at
+# the file "about" beside the directory "about/" the same page claims through
+# <a href="/about/">. That FileMap cannot exist on a filesystem: the generator
+# mkdirs over the file, EEXIST, and the whole URL import reports "failed". On one
+# real site 48 of 112 harvested files collided this way. Such a page counts
+# against MAX_CRAWL_PAGES, and against the asset loop's fetch ceiling (which is
+# now counted in slots, not stored files, so re-routing a fetch out of
+# assets_fetched cannot widen it). _LinkScan also stops queueing navigational
+# <link rel> values outright (canonical, prefetch, alternate & co) — the
+# content-type guard stays as the backstop for the ones it can't know about.
+# Scope, SSRF, robots and byte-budget behaviour are unchanged.
 
 """Same-site crawler with SSRF-hardened fetching for Paw Sites URL imports."""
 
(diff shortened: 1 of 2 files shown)


DevRohit06:
Two more bugs landed on this branch, both found by importing a real user's site (rohitk06.in) rather than by reading code. Both ended in the same failure the user hit:

```
SmokeGateFailed: html smoke: internal link/asset does not resolve: '/rss.xml'
```

## The feed regression was mine

The first commit put `alternate` in the navigational-rel denylist. That was wrong. It is how every RSS and Atom feed on the web is declared, and the feed is a genuine asset the page references. With it denied the feed was never fetched, the reference dangled, and the gate refused the publish.

`alternate` is dual-purpose, and hreflang alternates are exactly the case the content-type guard already handles, so the denylist entry was never needed. It is gone; the other eight rels stand.

## The gate was too strict for an import

Older, and the reason the user's import failed before this branch existed. The gate refuses to publish unless every internal reference in index.html resolves. Correct for a generated site. Wrong for an imported copy, where dangling refs are guaranteed: the crawl stops at 50 pages and 200 assets, robots can block paths, and the original can link to its own dead pages. That site does exactly that a dozen times, all logged as 404 skips during the crawl.

Unresolved refs are now warnings for an import and stay fatal everywhere else. Containment and every structural check stay fatal in both modes, so a ref escaping the site root is never downgraded. The refs reach the import report rather than only the log.

## Measured on the site that surfaced it

| | before this branch | now |
|---|---|---|
| Files | 93 | 69 |
| Feed harvested | yes | yes |
| File and directory collisions | 24 | 0 |

The generator writes all 69 files with no error. Before this branch the same map dies with EEXIST.

93 tests pass across the four affected suites. All 13 planned mutations were caught, none escaped.

## Three things worth knowing

The gate was never whole-site. It only ever parsed index.html, so dangling refs on the other imported pages were never checked at all. That makes the always-strict contract harder to defend than it looked, and it caps what the report can say: the warnings cover the home page only.

Leniency follows the pocket's pattern rather than the import event, so a user who later edits and re-publishes an imported site still gets the lenient gate. That is deliberate, since the source is still a copy, but it is a decision rather than a side effect.

One existing test asserted that an `alternate` link is never fetched, which is the rule being retracted here. It now covers a `prev` link instead and points at the new feed test.

PR #2098, opened 2026-09-06
Agentic browser tools, scoped to the /browser surface

prakashUXtech:
Adds the agentic browser behind the `/browser` surface. The user types what they want, a headless Chromium in their own container does it, and the answer comes back in chat. No live browser view, no URL bar — the browser is the tool, not the interface.

This is the first of seven slices. It ships the tool server and the boundary around it. The preamble, the route, `extract`, persistent profiles and cookie import come next.

## What's here

A new in-process MCP server, `pocketpaw_browser`, with seven verbs: navigate, snapshot, click, type, scroll, screenshot, close. One browser per workspace. It fails closed when there's no workspace in context.

## What was already broken

`src/pocketpaw/browser/` has been dead since Playwright removed `page.accessibility`. Every snapshot call raised `AttributeError`. The tests stayed green because the mocks had an attribute the real `Page` doesn't, which is worth remembering next time a mock and a real object disagree.

Snapshots now come from a single `page.evaluate` that walks the visible DOM and stamps `data-paw-ref="N"` on interactive elements. `RefMap`'s contract is unchanged; selectors are just `[data-paw-ref="N"]` now.

## The three guards

**Scoping.** The browser drives a real Chromium as the tenant, so it belongs on `/browser` and nowhere else — least of all `/chat`, which is otherwise unrestricted and carries the send-capable connector tools. An allow-list can't express that, because the surfaces needing the block are exactly the ones declaring no allow-list. So the tool ids go into `deny_mcp_tool_ids` for every non-browser surface, applied at the single `resolve_profile` chokepoint. That covers the default profile, unmapped kinds, and any row added later by someone not thinking about the browser. Deny is subtracted before the SDK launches, so the tools are unreachable rather than discouraged.

**SSRF.** Checked at the request level via `context.route`, reusing a new `safe_fetch.assert_public_url` so there's one IP-rule implementation in the codebase. This catches redirects, subresources and JS fetches, not just the entry point. Service Workers are blocked at the context, since `context.route` never sees their requests.

**Credentials.** `type` refuses password, card and OTP fields, read from the element's live attributes at type time rather than from the stale snapshot. The driver reports the field, the tool decides. This lives in code, not in a prompt, because it's the boundary the whole surface rests on.

Every action writes one audit row, refusals included, through the existing `record_tool_call`.

## Two bugs found in review

The SSRF guard was a self-inflicted denial of service. Aborting a top-level navigation makes Chromium commit its own error page, and that commit interrupts the next `goto` — one blocked address bricked the session for everything after it. Settling with `wait_for_load_state` doesn't help, because the error page has already reached `domcontentloaded` while the commit is still in flight. `navigate` now refuses before calling `goto`, so the common case never reaches the browser, and parks on `about:blank` when the route guard aborts a redirect hop.

The test meant to cover this started from a blank page and passed the whole time. An agent is never on `about:blank` when it hits a blocked address — it's mid-task on a real page, and that's the case that broke. The added test starts from a loaded page and fails without the fix.

The second was a fail-open in the scoping guard. It read its tool ids from a loader that wraps ten sibling imports in one `try/except`, so any unrelated module failing to import emptied the set and silently turned the deny into a no-op — while the provider, which imports on its own path, still registered the server. The browser became reachable from `/chat` behind nothing but a warning log. The ids now load independently, so the only failure that empties them is one that also stops the server existing.

## Tests

311 passed. The 2 failures in `tests/cloud/test_surface_profile.py` are pre-existing on `dev` from a skill rename; this branch touches `skill_names` zero times. Both import-linter configs clean.

Live smoke against a public page: snapshot with refs, metadata IP and loopback both blocked, session usable afterward.

## Worth knowing

`--no-sandbox` only applies inside a container and can't be exercised on macOS, so that launch path wants one run in the built image.

WebSockets are outside the interceptor. That needs `route_web_socket`. "Every request is checked" means every HTTP request.

`BrowserSessionManager.cleanup_idle` exists but nothing calls it, so one Chromium per workspace lives until `close`, uncapped. That predates this branch, but it wants a sweeper before this runs anywhere shared.


prakashUXtech:
Landed via #2100, which was collapsed and squash-merged to dev with the whole /browser backend arc (this PR's commits included). Closing as superseded, not abandoned.

PR #2099, opened 2026-09-06
A preamble for the /browser surface

prakashUXtech:
Gives `/browser` its own preamble. Without one the surface falls back to the generic prompt and the agent builds a dashboard instead of going to look at the web — the same drift the sites and studio preambles were written to fix.

Stacked on pocketpaw#2098, which ships the tools and the scoping. Targets that branch, so merge 2098 first.

## What the preamble establishes

**The browser is invisible.** The user cannot see the tab, so the agent never says "the tab", "the window", or "have a look". It reports what it found, not what it clicked.

**Refs go stale.** Snapshots carry `[ref=N]` markers, and those belong to the latest snapshot only. Anything that changes the page means take a fresh one before using a ref again.

**Credentials, never.** The agent never asks for a password and never types one. The tool already refuses password, card and OTP fields in code, so the preamble says that plainly — otherwise the agent burns a turn discovering it. When a page needs a login the workspace does not hold, it says so and stops looking for another way in.

**Page text is data, not instructions.** Anything read off a page is untrusted. Text addressing the agent directly gets ignored and surfaced to the user instead. This is the load-bearing anti-injection rule, and it is the reason the browser is confined to a surface with no send-capable tools on it.

**Blocked addresses and CAPTCHAs end the attempt.** A blocked URL is refused on purpose, so no retry and no hunting for a way around. A CAPTCHA gets reported, never solved or evaded.

**Prose by default, a pocket when the shape earns it** — a list, a table, a comparison. Existing widget types only, and no invented button verbs, because a button wired to a verb the dispatcher does not know renders fine and does nothing, which reads as a broken product.

The profile stays `ripple_mode="trim"` rather than Studio's `"off"`, because here a widget is often the deliverable rather than a distraction from it.

## One fix on top

The first draft listed `image` among the emittable widget types, for "a captured page worth showing". The screenshot tool hands the agent base64 bytes as an MCP image block, not a URL, so the agent would have had to invent a src and ship a pocket rendering an empty box. That is precisely the failure the no-invented-verbs rule warns about two sentences later in the same paragraph.

Removed, with a test pinning the absence, so the promise can only come back once the asset-URL contract exists. That is a later slice.

## Tests

318 pass. The 2 failures in `tests/cloud/test_surface_profile.py` are pre-existing on `dev` from a skill rename and are untouched here.

Seven tests cover this slice: the browser kind resolves to this handler rather than the generic one, the surface marker renders, and the credential, untrusted-content and blocked-address rules are all present. One pins the profile's allow set by equality while every other surface denies those ids, because the scoping had a fail-open in review and it is worth holding down from the handler side too.

Both import-linter configs clean.

## Known staleness

The preamble tells the agent that importing saved logins from settings is "coming but not live yet". True today. The slice that adds cookie import has to come back and update this line, or the agent will keep telling users a shipped feature does not exist.


prakashUXtech:
Landed via #2100, which was collapsed and squash-merged to dev with the whole /browser backend arc (this PR's commits included). Closing as superseded, not abandoned.

PR #2118, opened 2026-09-07
Terrarium launch: the clock, bring-your-own-key, books as files, births

prakashUXtech:
Everything the launch needs that the first runtime PR did not have. Stacked on #2108.

## What is here

**The clock.** A universe ticks on its own physics and drops to its dormant cadence when nobody is watching. This did not work when it was written: `mount_cloud` registered it with `@app.on_event("startup")`, which FastAPI silently drops because the app is built with a custom `lifespan=`. Every unit test passed and no world ever ticked. It now starts from `CloudLifecycleHook`, the seam the host really calls, where Paw Sites already works around the same trap.

**Bring your own key.** A universe pays with its workspace's key, resolved through `cloud.byok` — the product's one BYOK store, imported here from the Other-hand PR (#2015) rather than reinvented. There is no per-universe key and no second copy. The HTTP transport is key-agnostic; the key never lands in a wire dict, and a test asserts it appears in no response body.

**Books are files.** A book, law or map a citizen writes lands in the workspace's /files surface through the same `write_text_file` path the planner uses, so a viewer opens it in the normal inline viewer. Best-effort: a storage failure degrades to inline-only and the tick still lands.

**Births are visible.** A citizen asking for a child files a real Instinct Action, but only the tray knew, so nobody approved it and no world could reach generation 2. `GET /universes/{id}/gates` lists a universe's pending decisions. Read only — approving stays `POST /instinct/actions/{id}/approve`, because the Instinct gate is the single chain authority.

## Two bugs a live run found that the tests could not

**The rung contradicted the ladder printed beside it.** Five citizens showed "town" directly above the ladder's own line saying town means 10² souls. The old rule returned "town" on two unlocked tech nodes regardless of population. Population now gates the rung at exactly the thresholds the observatory prints, with tech as a floor.

**A child was its parent shifted, not a child.** Every trait got a flat `+0.08`, so a child was strictly more open, more conscientious, more extraverted, more agreeable and more neurotic than its parent, all five at once, every birth. Lineages never diverged in shape and six generations of it pins everyone at 1.0 — the evolution claim would have been empty while the numbers looked busy. Each trait now gets its own signed delta across the full range, deterministic and seeded from the parent DID and child name because the engine takes no RNG and a Journal replay has to reproduce the world.

Measured on the real approve path, a child of a 0.5 parent: `{O +0.011, C +0.023, E −0.022, A −0.070, N −0.020}`.

## Verification

`tests/ee/terrarium/`: 125 passed, 1 skipped. The skip is the end-to-end fork test, which states its reason — `Soul.fork()` has not reached the published soul-protocol floor. `lint-imports` 38 kept, 0 broken. `ruff` check and format clean.

Verified live against a running stack rather than by assertion: a world with a tick due every five seconds sat at zero through three sweep intervals before the fix and ticked on its own after it; unwatched it fell to dormant, reading the feed woke it, and the pool climbed as citizens acted. A spawn was proposed, approved through the real Instinct path, and the child arrived at generation 2 with five different signed deltas.

### DIFF
--- a/docs/c4/model.json
+++ b/docs/c4/model.json
@@ -30,7 +30,8 @@
               {"id": "pockets", "name": "Pockets + rippleSpec", "description": "Pocket workspace containers (agents+data+tools+KB+rules). The rippleSpec merge endpoint; data sources (read GETs) and write actions gated through Instinct. ee/cloud/pockets.", "technology": "Python", "kb_article": ""},
               {"id": "sites", "name": "Paw Sites", "description": "Publish pockets as edge sites. Two engines x two modes: {ripple,svelte} x {static,dynamic}; dynamic = per-tenant Cloudflare D1. ee/pocketpaw_ee/sites + the paw-sites generator.", "technology": "Python, Bun, SvelteKit", "kb_article": ""},
               {"id": "foresight", "name": "Foresight", "description": "Scenario simulation / rehearsal: soul-seeded personas run per-tick cognition loops with calibration and backtest gating. ee/pocketpaw_ee/foresight.", "technology": "Python", "kb_article": ""},
-              {"id": "enterprise-ops", "name": "Enterprise Ops", "description": "Fleet install, Calendar (external sync + conflict detection), Guards (RBAC/ABAC), Audit (placeholder), and the Paw Bar customer decision loop. ee/pocketpaw_ee/{fleet,calendar,guards,audit,paw_bar}.", "technology": "Python", "kb_article": ""}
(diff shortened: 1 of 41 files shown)


prakashUXtech:
Collapsed into #2109. Rebasing onto today's dev meant resolving the same conflicts twice through a two-branch stack, and the launch commits now live directly on `integration/terrarium` — all thirteen in one legible chain. Nothing is lost; #2109 carries everything this PR described plus the rebase fixes.

PR #2121, opened 2026-09-08
fix(guards): per-member action grants never applied, and the fix that looks obvious is worse

DevRohit06:
Audit finding H-4.

## What was wrong

`_has_action_override` was sync and reached an async service like this:

```python
overrides = asyncio.get_event_loop().run_until_complete(
    get_member_action_overrides(workspace_id, user_id)
)
```

Every caller is an async request handler, so a loop is already running and that raises `RuntimeError: This event loop is already running`. A bare `except Exception` swallowed it, `overrides` became `[]`, and the empty list went into a dict with no expiry — the comment said "Lives for process lifetime".

So every per-member grant an admin made in the UI silently did nothing, for the life of the process. It fails **closed**: users are denied actions they were explicitly granted. That is a permissions bug, not an escalation — it reads to a customer as "I gave them permission and nothing happened".

## Why the obvious fix is worse than the bug

This is the part worth reviewing carefully. `check_workspace_action` had both checks in one `try`:

```python
try:
    role = resolve_workspace_role(user, workspace_id)   # may raise not_member
    check_action(action, role)                          # may raise denied
except Forbidden:
    if _has_action_override(...):
        return role                                     # role may be UNBOUND
```

Two different failures arrived at one handler. If `resolve_workspace_role` raised, `role` was never bound — and that path was unreachable **only** because the override lookup always returned False.

Make the lookup work without restructuring, and it either raises `UnboundLocalError`, or — if someone gives `role` a default to silence that — it hands a **non-member of the workspace** access on the strength of an override row.

Membership is now resolved in its own `try`, before the override path exists. There is a test that authenticates a user belonging to no workspace, carrying a grant row for the action, and asserts both that it is denied `workspace.not_member` and that the override lookup never ran.

## Also fixed

A failed read is no longer cached — a transient database error used to pin a member to "no overrides" for the process lifetime. The cache has a TTL, and both `action_permissions` writers invalidate it, so an admin's change applies immediately rather than after the TTL.

## The conversion, and the test churn

`check_workspace_action` is now a coroutine function. 13 call sites await it, and two sync helpers (`_require_fleet_install`, `_may_buy_site_plan`) became async along with their single callers.

That also made 22 stubs across 14 test files wrong: a sync stub returning `None` makes the production `await` raise. Those now use `AsyncMock` / `async def`.

The `_deny` stubs are deliberately **untouched** — they raise while `f(x)` is being evaluated, before the `await`, so they still behave correctly.

**Worth a reviewer's eye on the test diff, not just the production diff.** I described this churn as mechanical and it wasn't: my first pass converted lambdas only and silently missed a named `_record` stub, which surfaced as a genuinely failing gate rather than a lint error.

## Tests

`tests/cloud/test_member_action_overrides.py` with `tests/mutations/member_action_overrides.json`. Six mutations, all caught, including both dropped-`await` cases.

That last gate walks the AST of both packages. A dropped `await` on an async guard builds a coroutine, never runs it, raises nothing, and the route allows — Python only warns on stderr at GC time, which is exactly how the original bug in this module stayed invisible.

## What is and is not verified

**Verified like-for-like:** `tests/cloud/test_audit_router.py` + `tests/ee/agent/test_workspace_admin_mcp` give 8 failed / 50 passed on this branch and 8 failed / 50 passed on a clean `dev` checkout. Identical, so those 8 are pre-existing.

**Still running as I open this:** the full `tests/cloud` + `tests/ee` sweep. `tests/cloud` carries substantial pre-existing failures on `dev` and is excluded from CI by `addopts` in `pyproject.toml`, so the totals need a clean-`dev` baseline to mean anything. I will post both numbers here when it lands rather than leave a claim unbacked.

Flagging that exclusion again because it applies to this PR's own gate: `tests/cloud/test_member_action_overrides.py` will not run in CI as things stand.

### DIFF
--- a/ee/pocketpaw_ee/agent/mcp_servers/fabric.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/fabric.py
@@ -400,7 +400,7 @@ async def _gate_admin(tool: str, workspace_id: str, user_id: str | None) -> dict
         return _error_response(f"{tool} could not resolve the calling user for the RBAC check.")
 
     try:
-        check_workspace_action(user, workspace_id, "fabric.admin")
+        await check_workspace_action(user, workspace_id, "fabric.admin")
     except Forbidden as exc:
         logger.info(
             "%s denied: user=%s workspace=%s code=%s", tool, user_id, workspace_id, exc.code
--- a/ee/pocketpaw_ee/agent/mcp_servers/growth.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/growth.py
@@ -185,7 +185,7 @@ async def _gate(tool: str, action: str) -> tuple[str, str, Any] | dict[str, Any]
         return _error_response("could not resolve the calling user for the permission check.")
 
     try:
-        check_workspace_action(user, workspace_id, action)
+        await check_workspace_action(user, workspace_id, action)
     except Forbidden as exc:
         logger.info(
             "%s denied: user=%s workspace=%s code=%s", tool, user_id, workspace_id, exc.code
--- a/ee/pocketpaw_ee/agent/mcp_servers/workspace_admin.py
+++ b/ee/pocketpaw_ee/agent/mcp_servers/workspace_admin.py
@@ -379,7 +379,7 @@ async def _gate_read(tool: str, action: str, deny_message: str) -> tuple[str, st
         return _error_response("could not resolve the calling user for the RBAC check.")
 
     try:
-        check_workspace_action(user, workspace_id, action)
+        await check_workspace_action(user, workspace_id, action)
     except Forbidden as exc:
         logger.info(
             "%s denied: user=%s workspace=%s code=%s",
@@ -500,7 +500,7 @@ async def _members_list_handler(args: dict) -> dict:  # noqa: ARG001 — no args
     # as a structured deny envelope (never raised out of the tool). The gate
     # already audits the denial via guards/audit.log_denial.
     try:
-        check_workspace_action(user, workspace_id, _READ_ACTION)
+        await check_workspace_action(user, workspace_id, _READ_ACTION)
     except Forbidden as exc:
         logger.info(
             "members_list denied: user=%s workspace=%s code=%s",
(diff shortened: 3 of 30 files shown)


DevRohit06:
CI is green after a rebase and one fix that is worth calling out, because it was not in the guard change itself.

## The three red checks had one cause, and it was collateral

Making `_may_buy_site_plan` async rotted an anchor in `tests/mutations/site_plan_purchase_authz.json` — a plan belonging to a different feature. Its anchor named `check_workspace_action(user, workspace_id, "sites.buy_plan")`, which now reads `await check_workspace_action(...)`.

That is worse than a stale string. **A plan with a bad anchor applies no mutations at all** — it aborts before the first one, so it reports as covered while proving nothing. Seven mutations guarding the site-plan purchase gate, including "the purchase gate is deleted — any member charges the company card", were silently disarmed by my change.

`Test (OSS-only)` and `Test (Python 3.12)` failed for the same reason: `tests/test_mutate_validate.py` runs the validator as a test and lives in the OSS scope.

Fixed by repointing the anchor at the code as it is now, keeping the harm the label describes, then re-running the whole plan. All 7 caught again. My own plan still catches all 6.

## A stale base, caught by a number that did not match

My local validator reported "1323 anchors across 103 plans"; CI reported 1336 across 105. That mismatch was the tell — `dev` had moved 3 commits underneath me. I rebased, and the local count now matches CI exactly at **1336 anchors across 105 plans**.

## What is verified, and what I did not finish

Verified like-for-like: `tests/cloud/test_audit_router.py` + `tests/ee/agent/test_workspace_admin_mcp` give **8 failed / 50 passed** on this branch and **8 failed / 50 passed** on a clean `dev` checkout. Identical, so those 8 are pre-existing.

Not finished: the full `tests/cloud` + `tests/ee` sweep. I stopped it deliberately — a mutation sweep rewrites source files, so it cannot run alongside a test run, and fixing CI came first. I am not claiming a full-suite comparison I did not complete.

What that leaves resting on: the per-file parity above, the six mutations, and the AST gate that fails if any call site drops its `await`. The dropped `await` is the one silent failure mode of this conversion (the guard would build a coroutine, never run, and the route would allow), and that is the gate covering it.

Happy to run the full sweep plus a clean-`dev` baseline before merge if you want the number — it is about 70 minutes of wall clock for the pair.

PR #2125, opened 2026-09-08
A file with no typed message is still a message

DevRohit06:
Attach a file, type nothing, send. The agent asks what you wanted — with the answer already in its prompt.

## What was happening

The composer puts a zero-width space in the message body when you attach files and type nothing. That is right for the two ends it was written for: the row has to be non-empty or the backend drops the message, and every renderer strips the character so no placeholder shows in the transcript.

The model is the end nobody wrote it for. It received a user turn that said nothing, plus the file's extracted text inlined into the knowledge channel — where it reads as reference material rather than as the request. The only reasonable move left is to ask what the user wanted. Typing a single character alongside the attachment made it work, which is the tell that the attachment was never the problem.

This bites hardest on the paste path, because the composer converts any bulky paste into a `pasted-*.txt` attachment. So "I pasted my brief into the chat" and "I uploaded a file" are the same code path, and pasting a full site brief produced `I don't have a brief yet — what's the site for?`.

## The fix

`_drive_agent_loop` resolves that one case before anything downstream reads the string — the KB query, the session titler, and the prompt handed to `pool.run` all consume it. The substitute names `<uploaded-files>` so the instruction points at the block the text actually landed in.

A turn with visible text is returned byte-identically, sentinel included. With no attachments nothing is invented.

## Second cut on the same trip

The per-file inline cap was 8000 chars, around 1200 words. A real landing-page brief clears that easily, so the agent was reporting the paste as truncated and reconstructing the rest from what survived — the user had truncated nothing, we had. Per-file goes to 40000 and the total to 100000. Still capped, just set by what people paste rather than by what a document can weigh; the total cap is what keeps five large files from becoming the whole prompt.

## Tests

Two new files, both red before the change:

- `tests/cloud/runs/test_run_core_attachment_only_turn.py` — the attachment-only turn reaches the model as a request; the substitute reaches both `pool.run` and the KB query; a typed message passes through byte-for-byte; an empty turn with no attachments invents nothing.
- `tests/cloud/chat/test_attachment_inline_budget.py` — a 30k-char brief inlines whole; something genuinely enormous is still cut and still says so; the total cap still bounds a batch.

`tests/cloud/runs`, `tests/cloud/chat` and `tests/ee/agent`: 1003 passed. Six failures in `test_run_core.py` and `test_run_core_entity_profile.py` are pre-existing — verified failing identically on a clean `origin/dev` worktree, and they concern surface/entity profile resolution, not this path.

### DIFF
--- a/ee/pocketpaw_ee/cloud/chat/agent_service.py
+++ b/ee/pocketpaw_ee/cloud/chat/agent_service.py
@@ -9,6 +9,16 @@ handles *what the agent sees*:
 * ``load_history_for_scope`` rehydrates prior chat turns from Mongo so the
   agent carries context across backend restarts and pool evictions.
 
+Changes: 2026-09-08 (fix/attachment-only-turns) — added
+``resolve_user_content`` / ``visible_message_text``: the model's copy of a turn
+that carried attachments and no typed text. The composer sends a zero-width
+space as the body so the row persists and renders blank, and that sentinel was
+being handed to the model as the user's whole message — a pasted brief reached
+the prompt but read as reference material, so the agent asked for a brief it
+already had. The inline-attachment caps moved with it (per-file 8000 -> 40000,
+total 30000 -> 100000): a bulky paste IS an attachment on this path, and the
+old per-file bound cut a normal landing-page brief in half.
+
 Changes: 2026-08-07 (fix/code-delegate-pooled-context) — added the session-keyed
 stream registry (``register_stream_sink`` / ``unregister_stream_sink`` /
 ``stream_sink_for_session``) alongside the ``_sse_event_sink`` ContextVar. The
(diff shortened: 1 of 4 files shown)


DevRohit06:
Pushed a second commit. The first pass fixed half of it, and the half it left is the half that mattered.

The agent's newest reply — `Looks like your message came through empty` — is the model reporting exactly what it saw at the user-turn position. The file's text was in the prompt, but inside the knowledge wrapper, whose header reads:

> `## Your Knowledge Base` — Use the following information from your knowledge base to answer questions.

That is the wrong sentence to print above something the user wrote thirty seconds ago. Telling the model "the files are the message" while the framing one layer up says "this is reference data you may consult" is a fight worth not having.

So on an attachment-only turn the `<uploaded-files>` block now rides the user turn itself. `build_knowledge_context` takes `skip_attachments`, so it is **moved, not copied**: extraction runs once and the text appears in the prompt exactly once. The KB query and the session titler keep the short substitute — a 40k-char brief handed to a keyword search is not a search.

A turn with typed text keeps the routing it has always had, block included. This is a fix for the empty turn, not a rewrite of every upload, and there is a test pinning that.

PR #2189, opened 2026-09-15
Let one surface grant one external MCP server

DevRohit06:
## The thing I got wrong first

Chasing "can Refero's MCP reach the Paw Sites agent", I concluded a surface could only be granted an external MCP server via `ALWAYS_ALLOWED_MCP_SERVERS` — i.e. by granting it to *every* surface — and that a new field plus threading through `pool.run` → `claude_sdk.run` was needed to fix it.

That was wrong, and it was wrong because I reasoned about the allow-list shapes instead of running them.

`_collect_mcp_tool_ids` allow-lists each enabled external server wholesale with a **bare `mcp__<server>` token** (no tool segment), because an external server's tool names are not known until the client connects. The surface allow-list is compared against `allowed_tools` by exact string — and the token on both sides is the same string. Naming `mcp__refero` in `sites_allow` admits the server, on that surface and nowhere else. Measured against the real `resolve_profile(SurfaceKind.SITES, SurfaceMeta())`:

| case | survives |
|---|---|
| `mcp__refero`, not named by the surface | `False` |
| `mcp__refero`, named in the surface's allow set | **`True`** |
| `mcp__pocketpaw_foresight__save_scenario` (control) | `False` |

So the mechanism already existed on the SDK backend. Nobody had used it. `SurfaceSpec.mcp_provider` stays dead and does not need reviving.

## What was actually broken

**The two backends disagree about what a bare server token means, and the one that gets it wrong is the one that matters.**

`pydantic_ai._gate_mcp_toolsets` filters by comparing each `tool_def.name` against `_expand_tool_ids(permitted)`. That maps `mcp__refero` → `refero`, which matches no tool the server exposes, so every tool is dropped. Per the method's own docstring, its MCP toolsets **are** "the user's EXTERNAL configured servers" — the grant meant nothing in precisely the place external servers live.

`_bare_mcp_server` now reads a bare id as a server-level grant, and tools are kept by their `<server>_` prefix — the shape `PrefixedToolset(MCPToolset(client), cfg.name)` gives them.

### A second divergence, found by the test for the first

`_gate_mcp_toolsets`' early return fired on an **exclusive** turn with no allow set and no deny set, returning every toolset untouched. The documented CX-1 rule is the opposite: an exclusive turn with no allow set is an empty permitted set, so it strips *all* MCP tools — which is what the SDK backend does. An exclusive agent therefore leaked every external MCP tool on this backend. `exclusive_mcp_tools` now reaches the gate.

I found this only because I wrote `test_an_exclusive_turn_with_no_grant_drops_the_server` expecting it to pass trivially.

## The opt-in

`POCKETPAW_SITES_MCP_SERVERS`, comma-separated, **empty by default** — an install that has not opted in resolves an identical allow set (verified: 21 ids before, 21 after).

Two deliberate choices:

- **A comma-separated string, not `list[str]`.** pydantic-settings JSON-decodes complex types *before* any validator runs, so `POCKETPAW_SITES_MCP_SERVERS=refero` would crash boot on a list field.
- **Names are not validated against `load_mcp_config()`.** An unmatched grant is already inert — the token never appears in `allowed_tools`, so nothing passes. Validating would couple surface resolution to the MCP config file and hand an optional integration a way to break it. `_external_sites_mcp_grants` also swallows its own exceptions, because it is called inside `_load_mcp_tool_ids`' big try/except, where an escape would degrade `sites_allow` to `None` and turn the sites MCP filter off entirely — a much worse outcome than no grant.

## Tests

11 new, with the controls that make them worth having:

- an **unnamed** external server is still refused (proves the grant is scoped, not a widened gate)
- `reference_search` is in the fixture because `"reference_search".startswith("refero")` is `True` — it pins that the prefix match needs its separator
- deny still beats a server grant
- a fully-qualified id still matches on its own
- the default allow set contains no bare server token at all

## Verification

- `tests/cloud/surface/ tests/test_pydantic_ai_backend.py tests/test_inprocess_mcp_bridge.py` → **435 passed, 1 failed**
- `scripts/mutate.py --validate` → **OK: 1754 anchors across 148 plans each resolve exactly once**
- `ruff check` / `ruff format --check` clean

**The one failure is pre-existing and not mine.** `test_local_machine_tools_are_never_offered` fails identically on clean `dev` with the same command (confirmed by stashing this branch and re-running: 424 passed, same 1 failed). It is order-dependent — `remember` drops out of the bridged tool surface when the whole `tests/cloud/surface/` directory runs first, and it passes in isolation. Worth its own look; I have not touched it here.

## Observed, not fixed

`get_settings` is `@lru_cache`d with `force_reload` **as part of the cache key**, so `get_settings(force_reload=True)` reloads on its first call and every later call with that argument is a cache hit returning the stale object. Force-reload works exactly once per process. The tests here call `get_settings.cache_clear()` instead. Fixing it properly would change when settings reload process-wide, which does not belong in this PR.

## Not in scope

No bundled skill mentions Refero yet. Until a deploy sets the setting, telling `sites-design-sources` to "research with Refero first" would name tools the agent does not have.


DevRohit06:
Full-suite result, as promised.

`uv run pytest tests/ --ignore=tests/e2e --ignore=tests/test_frontend_syntax.py -q -p no:randomly` → **11556 passed, 69 failed, 53 skipped** (28m30s, Windows).

**None of the 69 are in a file this branch touches** — no failure in `test_pydantic_ai_backend.py`, `tests/cloud/surface/`, or anything config-related. They land in four unrelated subsystems and are platform, not logic:

| file | count | cause |
|---|---|---|
| `test_herdr_runtime.py` | 26 | assertion + cp1252 decode |
| `test_codex_cli_backend.py` | 18 | `asyncio.start_unix_server` |
| `test_code_mode_runner.py` | 10 | `asyncio.start_unix_server` |
| `test_code_mode_bridge.py` | 8 | `asyncio.start_unix_server` |
| 4 others | 7 | assorted |

`asyncio.start_unix_server` does not exist on Windows, and `UnicodeDecodeError: 'charmap'` is the cp1252 default encoding — both are this machine, not this change. CI runs Linux and should be clean.

Worth noting: `test_local_machine_tools_are_never_offered`, the one failure I reported above from the narrower run, **passed** in the full ordered suite. That confirms it is order-dependent rather than broken, and reinforces that it is not related to this branch.

PR #2192, opened 2026-09-15
Vendor Refero's research methodology as sites-design-research

DevRohit06:
> **Stacked on #2190**, which is stacked on #2189. Base is `feat/refero-design-research`, so this diff shows only the skill. Merge the stack bottom-up with `--rebase`.

## What this is

The methodology half of the [Refero Skill](https://github.com/referodesign/refero_skill) (MIT, commit `a9b54a3e62a6`), vendored as an on-demand `/sites` design skill and **renamed**.

It is the research step that feeds `pocketpaw-design-taste`: this skill finds and locks a direction with evidence, that one builds the page from it. #2190 gave us the data; this is the discipline for using it.

## Why renamed

Upstream is `refero-design`. The capability is "research the design before you make it", not "call Refero" — and a skill named after a vendor either lies or needs renaming across every preamble the day the subscription changes. `sites-design-research` says what it does.

## Most of upstream is deliberately not here

Upstream is ~4,200 lines across a SKILL.md and ten reference files. About 200 are additive here. The rest would have actively hurt, and the reasoning is written up in `VENDORED.md` rather than left implicit:

**The craft references** — typography (737 lines), color (563), motion (504), craft-details (525), anti-ai-slop (292), copywriting (222), icons (166). We already carry all of it, tuned to this surface, across `pocketpaw-design-taste`, `sites-craft`, `sites-theme-system` and `sites-restraint` — and `sites-craft` is **embedded whole** in the create preamble. Bundling upstream's craft half would put a second, differently-opinionated design system in the same prompt as ours on every create turn, with no way for the agent to adjudicate a disagreement.

**The primacy framing — the load-bearing removal.** Upstream's `description` declares itself the *"Primary/default skill"* for essentially all UI work and instructs the agent to *"prefer over broad generic product design, frontend design, UI polish, CSS framework, landing page, or craft-only skills; those may only supplement implementation details after Refero research and synthesis."* Read against our catalogue that names `pocketpaw-design-taste`, `sites-craft`, `sites-design-sources` and `impeccable`, and tells the agent to demote all four. Installed unmodified it would have quietly become the design authority on `/sites` while knowing nothing about the surface's constraints.

The rewritten description claims the research step only, and the skill body opens with an ownership table handing authority back:

| Question | Owner |
|---|---|
| What should this look like, and on what evidence? | **this skill** |
| How is the page built, scoped, kept from looking AI-made? | `pocketpaw-design-taste` |
| What does the page SAY, and in what order? | `sites-conversion-structure` |
| How do I hand-write this ground / shader / type choice? | `sites-design-sources` |

**The flows layer**, because #2190 bundles styles and screens only — a Paw Site is a marketing page, not a multi-step journey. Leaving it in would name a tool the agent does not have, which is already on this codebase's record as a real failure mode. Verified absent: no `search_flows` string survives in the skill.

**`visual-workflow.md` and the screenshot QA pass**, which need a browser and a filesystem that `_SITES_BUILTIN_DENY` strips on every `/sites` mode.

**`mcp-tools.md`**, which documents Refero's raw tool names. Ours are namespaced differently per backend, so the routing table was rewritten to carry both spellings (`search_styles` / `get_style` / `search_screens` on the MCP surface, `refero_design_styles` / `refero_design_screens` as BaseTools).

## What was worth taking

The discipline we genuinely lacked: search several angles varying the **axis** rather than re-wording one query; styles first, screens for structure; **don't copy one reference and don't average several** (blending yields the safe centroid, which is exactly the generic output research was meant to escape); **preserve roles** (a CTA-only colour stays CTA-only or is omitted); and the reference lock + decision ledger, whose test is the point — *if a major choice has no source, it is not a design decision yet.*

## Added beyond upstream

- The ownership table above, and the reminder that **a reference never justifies a section the brief did not ask for** — MODULE 0 scope discipline is ours, not upstream's.
- **The unconfigured path.** Refero needs a paid plan, so an empty result is the *normal* outcome here, not an error. The skill says so explicitly, including "never cite a reference you did not receive". Upstream assumes the MCP is connected and treats research as mandatory — which on an unconfigured deploy is an instruction to fabricate.

## Wiring

Registered scope `both` in `_SITES_DESIGN_SKILLS`, so it derives into `create_design_skill_names()` and the svelte/react create `skill_names` allowlist automatically — that derivation is what `test_sites_create_skill_names_cover_the_advertised_skills` proves, and naming it in only one place would have made it advertised-but-unloadable. Not embedded: its trigger is a real branch ("the brief leaves the look open"), not every turn.

Confirmed rendering in the create `<design-skills>` block, and present in svelte-create `skill_names`.

## Verification

- `tests/cloud/surface/` + both refero suites: **277 passed**
- `ruff check` clean
- `scripts/mutate.py --validate` → **1754 anchors across 148 plans, each resolving exactly once**

That last one is not a formality here. My first attempt at the `handlers/sites.py` edit rewrote the file LF, flipping it from CRLF — a 12-line addition that reported as **2136 insertions / 2124 deletions** and silently broke 6 mutation anchors across 2 plans. A plan with one bad anchor applies *no* mutations, so it reads as covered while proving nothing. Redone in binary with CRLF preserved; the diff is now the 12 lines it should be.

## Licence

MIT, attribution retained in `VENDORED.md` alongside what was ported, what was not, and what to re-check on a refresh — specifically whether an upstream refresh has pasted the primacy framing back, since stripping it is what this vendoring is for.

## Not verified

No site has been authored under it, and the tools it routes to have never returned live data here — this repo has no Refero token, so every run so far has exercised the unconfigured path.


DevRohit06:
CI's `atlas build --check` was right — fixed in `b6e645a9`.

Adding a bundled skill means the generated atlas artifact (`src/pocketpaw/atlas/data/atlas.json`) goes one entry short until it is rebuilt. Ran `pocketpaw atlas build` from the repo root; the diff is **14 lines, 0 deletions** — one entry slotted alphabetically between `sites-craft` and `sites-design-sources`, its `summary` and `narrative` derived from the skill's own frontmatter description. No absolute paths in the output and the file stays LF, so a Linux runner regenerates the same bytes. `--check` now reports up to date.

**One thing worth flagging:** my first push was rejected because this branch had been **rebased** on the remote while I was working — all three commits carry new SHAs and `b513c099` (#2171, image attachments) is now in the history. Rather than force-push over that, I reset onto the remote tip and regenerated the artifact there. Building it fresh beats replaying a copy made before the rebase, since the file is generated rather than authored.

Re-verified on the rebased tip: **277 passed**, and `mutate.py --validate` now reports **1763 anchors across 149 plans** (up from 1754/148 — #2171 brought its own plan in) all resolving exactly once.

I also checked the three sibling branches — #2189, #2190 and #2193 all report the atlas artifact up to date, so this was the only one affected.

PR #2193, opened 2026-09-15
Harvest Refero's copywriting craft into sites-conversion-structure

DevRohit06:
First harvest from the Refero craft references, taken as the alternative to porting all ~3,600 lines and deleting ours.

## Why this file first

A gap analysis across all seven upstream reference files put `copywriting.md` clearly first:

- **~60% net-new to us** — the highest ratio of the seven
- **The only one with zero capability assumptions.** The other six assume npm (`virtua`, `nuqs`, GSAP, Rive, Lottie), local WOFF2 files, `next/image`, or looking at a rendered page. `/sites` has no package manager, no filesystem, and denies `Read`/`Glob`/`Bash`.
- **Zero conflicts.** The other six carry fourteen contradictory concrete values between them, two of which are straight reversals of rules we hold (they recommend Inter by name; they list dark-mode-by-default as an AI tell).

## Where it lands and why

`sites-conversion-structure` already owns *what the page says* — it has the headline formulas and CTA rules. It is also **invoked on demand rather than embedded**, so 110 lines of craft costs nothing on a turn that doesn't need them. That matters given the active prompt-diet work on this surface.

New `## 6. Copy craft`:

- **Clarity > Respect > Character**, in that order — with the diagnosis that most bad brand voice starts at character and never gets back to clarity
- **Dead vs living words**, with the removal test: if a word can come out without changing the meaning, it was dead
- **The six-part check** for copy that feels empty — context, job, solution, outcome, conditions, next step
- **Info-style** — facts beat adjectives, verbs beat nouns, and *higher stakes means calmer tone*
- **Scenes, not claims**, with the who/where/what-happens/what-changes template (and a note not to invent a customer to build one)
- **Rhythm**, and **the sticky-line test** — one quotable line per page, not every line straining to be it
- **Headings describe, they do not aspire**, with the operator litmus: if someone scans only headings, labels and numbers, do they understand the page?
- **The three-rung error ladder** — "Something went wrong" → "Couldn't save. Check your connection." → "You're offline. Reconnect to save changes."
- **Four banned-word categories.** The AI-opener set had *no* equivalent here: "In today's fast-paced world…", "Look no further", "Supercharge your…", "Take your X to the next level".
- **The cut-30% test**

## Not ported

The button verb+object rule and the marketing-hero question list (already in section 5); "Proof beats hype" (covered by section 4's be-specific rule plus design-taste MODULE 4's organic metrics); empty-state copy, which is product-UI guidance with little bearing on a marketing page.

## One deliberate duplication

`pocketpaw-design-taste` MODULE 4 keeps its seven-word filler list even though section 6 now carries a ~60-term catalogue that contains it.

MODULE 4 is **embedded** in the create preamble unconditionally. Section 6 is a skill the agent has to choose to invoke. Deleting the short list would trade a guarantee for a probability — which is the exact failure `_design_taste_system()` was written to prevent ("the sites build agent was not reliably INVOKING the design-taste skill… so sites came out generic"). Section 6 names the relationship explicitly so neither reads as contradicting the other.

## Mechanics

Sections 6-8 renumbered to 7-9, and the one internal cross-reference ("build section by section per section 6") updated with them. Checked first that no mutation plan anchors on this file's headings — the one plan that mentions this skill anchors in `handlers/sites.py`, not here.

## Verification

- `tests/cloud/surface/` → **251 passed**
- `scripts/mutate.py --validate` → **1754 anchors across 148 plans, each resolving exactly once**
- File stays pure LF (110 insertions, not a whole-file flip)

MIT attribution and the full ported/not-ported/not-deduplicated reasoning recorded in `VENDORED.md`, alongside the skill's existing first source.

## Next

Six files left, ~400 net-new lines between them, and they need the fourteen value conflicts adjudicated before anything lands. Ranked by yield: `craft-details` (~120 — the autocomplete table, `:focus-visible`, the touch block, ARIA error wiring), `typography` (~100), `color` (~90 — the `color-scheme`/`theme-color` plumbing is genuinely missing here), `motion` (~75 — five named easing curves where we name one), `anti-ai-slop` (~40 — the six litmus tests), `icons` (~30).


DevRohit06:
Second harvest pushed to this branch: **craft-details → `sites-craft`** (`a534bbcd`).

## Placement was the whole decision

`_craft_system("floor")` slices this file from the `## 5. The floor` **heading to EOF**, so anything added at or below it rides **every refine turn** — and the 2026-09-08 diet exists precisely because refine was paying 2,868 tokens to shorten one headline.

So the split followed the floor's own stated test: *can a copy edit break this?*

| destination | reaches | what landed |
|---|---|---|
| `## 5. The floor` | every refine turn | `:focus-visible` rather than `:focus`, with the reason (`:focus` fires on mouse click too, which is why rings get removed) and the two-ring `box-shadow` replacement |
| **`## The page shell`** (new, before the floor) | create only | viewport-meta zoom rule, `[id] { scroll-margin-top }`, `touch-action: manipulation`, `preconnect` |
| **`## Forms and input`** (new, before the floor) | create only | `autocomplete` table, `type`/`inputmode`, `spellcheck="false"`, never-block-paste, label-plus-control as one hit target, `aria-invalid`/`aria-describedby`/`role="alert"` + focus the first failing field |
| `## 4. Surface` | create only | `fetchpriority="high"` on the LCP image with no `loading="lazy"`, lazy + `decoding="async"` below the fold, explicit dimensions |

**I got this wrong on the first pass and measured it.** All four page-shell rules initially went into the floor, which grew it **+31%** on every refine turn. They are written once at create and a copy edit cannot break them, so charging every refine for them was the diet running backwards. Moved them out; the floor now grows **+12%**, and that 12% is the one rule a refine can actually break.

Measured: refine floor 2,312 → 2,601 chars. Verified the slice still resolves, does not degrade to the full file, and that neither new section leaks into it.

## What I did not touch

`_CRAFT_FLOOR_HEADING = "## 5. The floor"` is a literal in `sites.py` **and** itself a mutation anchor. Renumbering the floor would have made `body.find()` return -1 and silently degraded refine to the whole 2,868-token file — a regression with no error anywhere. The heading is untouched; both new sections are inserted above it and are unnumbered, which matches the file's existing style (`## Before you ship a section`).

## Conflicts resolved in our favour

Our values were measured against this surface, so all three upstream contradictions were dropped rather than merged:

- `scale(0.98)` on press — we specify `0.96` and say above `0.98` is invisible
- 32×32px desktop icon hit area — ours is 44×44 everywhere
- "Title Case for buttons" — contradicts `sites-interface-review`'s verb-first `Save changes` and our one-capitalization-policy rule

## Not ported

Everything npm- or framework-shaped (`virtua`, `nuqs`, `next/image`, uncontrolled-input patterns); `content-visibility` virtualization and `Intl.*` formatting, both marginal on a static marketing page; and `<link rel="preload">` for a local WOFF2, which cannot exist here — `sites-design-sources` states flatly that a font file cannot be brought into the project.

## Verification

251 passed · 1754 mutation anchors resolve · pure LF · MIT attribution and the full placement reasoning in `sites-craft/VENDORED.md`.

DevRohit06:
Third and final pass pushed (`1a97763f`). **All seven upstream reference files are now harvested.**

| upstream file | landed in | net-new |
|---|---|---|
| copywriting | `sites-conversion-structure` §6 (new) | ~110 |
| craft-details | `sites-craft` floor + 2 new create-only sections + §4 | ~120 |
| motion | `sites-craft` `## Motion` (new) | ~65 |
| colour | `sites-craft` §2 | ~40 |
| typography | `sites-craft` §1 | ~40 |
| anti-ai-slop | `sites-ship-fixes` §8B (new) | ~30 |
| icons | `sites-craft` §4 | ~15 |

542 insertions, 7 deletions, across 3 skills.

## The number that needs a decision

`sites-craft` is embedded **whole** on create, so:

- **create preamble: +46%** — 4,080 → 5,956 approx tokens
- **refine floor: +0%** — every line went above `## 5. The floor`

The refine side was the thing the 2026-09-08 diet was protecting, and it is untouched. Create is where design quality *is* the job, so growth there is more defensible — but 46% is real and it is your call, not mine.

If you want it smaller, the honest candidate is `## Forms and input` (~50 lines): it is conditional even on create, and a brochure page with no form pays for it today. Moving it to a named on-demand skill would take create to roughly +25%. I left it embedded because a form that silently ships without `autocomplete` or `aria-invalid` is exactly the failure an invocation-gated skill produces four turns in five.

## Conflicts, all resolved our way

None of these were merged — our values were measured against this surface:

- **Inter / Geist / Plus Jakarta** recommended by name upstream; `pocketpaw-design-taste` 2.F bans them by name as the loudest AI tell
- **light-mode-by-default**, which would make three of our six aesthetic families unreachable
- `rgba(0,0,0,0.45)` tertiary text — `sites-restraint` documents that rung as failing 4.5:1 and requiring a lift to ~56%
- 1.2 default ratio, 1.7 body leading, 11px minimum size, `0.06-0.10em` caps tracking — all outside our stated ranges
- enter=ease-out / exit=ease-in — contradicts `sites-interface-review`'s "exits are softer than enters, ease-out both ways"
- `scale(0.98)` press, 32px desktop hit area, Title Case buttons (from the earlier pass)

## Also not ported

Everything npm- or framework-shaped (Framer Motion, GSAP, Rive, Lottie, `virtua`, `nuqs`, `next/image`); `<link rel="preload">` for a local WOFF2, which cannot exist here; `content-visibility` and `Intl.*`, marginal on a static marketing page; and upstream's screenshot/browser QA passes, which need tools `_SITES_BUILTIN_DENY` strips.

## Verification

251 passed · 1754 mutation anchors resolve · floor slice confirmed still resolving with nothing leaking into it · all files pure LF · MIT attribution and full ported / not-ported / conflict reasoning in each `VENDORED.md`.

